PR #31, opened 2015-05-08
tomviz: shorten the RC command line length

mathstuf:
Clear out the include paths so they don't overflow the command line
length limit.

---
Better would be to use ParaView's branded application logic which already does this, but I'll need to look into that.


cryos:
I am guessing this actually overflowed something when compiling tomviz for Windows

PR #51, opened 2015-05-27
Non modal python dialog

cryos:


mathturtle:
+1
I read over it and it makes sense.  I tested it and it works.

cryos:
Thanks!

PR #55, opened 2015-05-29
cmake: add version detection code

mathstuf:
---
@cryos

### DIFF
--- a/cmake/Git.cmake
+++ b/cmake/Git.cmake
@@ -1,0 +1,31 @@
+#=============================================================================
+# Copyright 2011 Kitware, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#=============================================================================
+
+# Check for a hint left by the 'GitInfo' script.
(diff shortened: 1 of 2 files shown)


cryos:
I don't see anything that uses this, did you want to merge this before anything uses it?

cmake/TomVizDetermineVersion.cmake
```diff
@@ -0,0 +1,86 @@
...
+#  All rights reserved.
+#  See Copyright.txt or http://www.paraview.org/HTML/Copyright.html for details.
+#
+#     This software is distributed WITHOUT ANY WARRANTY; without even
+#     the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+#     PURPOSE.  See the above copyright notice for more information.
+#
+#=========================================================================
+
+# Used to determine the version for TomViz source using "git describe", if git
```
cryos: Just tomviz, all lower case...

cmake/Git.cmake
```diff
@@ -0,0 +1,31 @@
+#=============================================================================
+# Copyright 2011 Kitware, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
```
cryos: Is this tiny file really Apache 2? I thought it would be BSD like most of the CMake code in our projects. Guessing maybe it came from ITK or something?

mathturtle:
It is used by his new modifications to the superbuild I think.  https://github.com/OpenChemistry/tomviz-superbuild/pull/26

cmake/Git.cmake
mathstuf: It probably came from ITK. Brad imported it back in 2011 but doesn't mention where it came from.

mathstuf:
It is used by the superbuild change.

PR #64, opened 2015-06-02
Fix #42 open dialog knows only common tomography file types

### LINKED ISSUE #42: File Input: White list relevant file types only
Reduce the list of file types for input in order to reduce potential bugs and make the user experience easier.

.jpg, .png, .tiff, .raw, .txt are all useful

mathturtle:
@cryos How is this?

### DIFF
--- a/tomviz/LoadDataReaction.cxx
+++ b/tomviz/LoadDataReaction.cxx
@@ -32,6 +32,8 @@
 #include "vtkSMSourceProxy.h"
 #include "vtkTrivialProducer.h"
 
+#include <QFileDialog>
+
 namespace tomviz
 {
 //-----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


cryos:
This looks a lot better, I love the native file dialog being back too! There are one or two extensions that need adding. For example TIFF has .tiff and .tif (from the good old days when Windows only allowed three letter file extensions).

tomviz/LoadDataReaction.cxx
```diff
@@ -48,27 +50,38 @@ LoadDataReaction::~LoadDataReaction()
...
-  QList<pqPipelineSource*> readers = pqLoadDataReaction::loadData();
+QList<DataSource*> LoadDataReaction::loadData()
+{
+  vtkNew<vtkSMParaViewPipelineController> controller;
 
-  // Since we want to ditch the reader pipeline and just keep the raw data
-  // around. We do this magic.
-  foreach (pqPipelineSource* reader, readers)
+  QStringList filters;
+  filters << "Common file types (*.jpg *.png *.tiff *.raw *.txt)"
```
cryos: I think we should add *.dat, *.jpeg, *.tif and *.bin up here (and they should be added below too).

mathturtle:
I've updated the branch to add the new file types you mentioned and rebased so that it will merge cleanly.

cryos:
Why is it still displaying WIP? This was your original commit message?

mathturtle:
Yes, that was the original commit message.  I took it out when I amended
the commit, but Github didn't update the PR.

On Wed, Jun 3, 2015 at 11:03 AM, Marcus D. Hanwell <notifications@github.com

tomviz/LoadDataReaction.cxx
```diff
@@ -48,27 +50,40 @@ LoadDataReaction::~LoadDataReaction()
...
 
-  // Since we want to ditch the reader pipeline and just keep the raw data
-  // around. We do this magic.
-  foreach (pqPipelineSource* reader, readers)
+  QStringList filters;
+  filters << "Common file types (*.jpg *.jpeg *.png *.tiff *.tif *.raw *.dat *.bin *.txt)"
+          << "JPeg Image files (*.jpg *.jpeg)"
+          << "PNG Image files (*.png)"
+          << "TIFF Image files (*.tiff *.tif)"
+          << "Raw data files (*.raw)"
```
cryos: The data and binary should all be lumped with Raw data files I think, as that is the reader we have to read them. We probably also need to do a little work to coerce that reader into being used.

cryos:
Ah, that is weird (didn't want to merge something marked as WIP).

Issue #59, opened 2015-05-31
Manual alignment: allow user to pick direction


ElliotPadgett:
In the current version the user always aligns image n+1 to image n.  The user should be able to choose between "align to next" and "align to previous" options.

Issue #70, opened 2015-06-03
Manual alignment: allow user to pick static reference image for alignment


ElliotPadgett:
User should be able to pick a static reference image to align to (e.g. align all images to feature in image 15) in addition to using the previous or next images as reference as described in issue 59.

PR #71, opened 2015-06-03
Manual align widget: added features to choose reference image

ElliotPadgett:
Added features to allow user to choose if they want to align to the previous image, next image, or a static reference image.  This resolves issues #59 and #70.

Also adjusted default frame rate and starting image, and added some comments.


cryos:
Removed merges, added full author information, commit to make changes to adhere to coding style, looks ready to merge now.

PR #74, opened 2015-06-05
Add a resample action that downsamples a dataset

mathturtle:
@cryos This is definitely still preliminary (horrible memory efficiency among other things), but it works.  I'm defaulting to the 'linear' sampling type on vtkImageReslice.  I didn't see bilinear as an option, so we may have to add that if we want it.  The other options are 'cubic' and 'nearest neighbor'.  I'm not sure how to save the history of how the downsampled dataset got created though.


cryos:
I think this looks good as a first pass, I am going to merge it and we can open bugs for enhancements once it is in.

Issue #38, opened 2015-05-12
Allow moving of objects in the 3D view, interactively as well as using interaction.


utkarshayachit:

utkarshayachit:
Note all modules should also move when the data is moved.

mathturtle:
@Hovden @cryos This is another one where I'm not entirely sure how this should work.  From what I recall of our discussions we wanted to move datasets relative to each other and then lock down the center of rotation for the scene to the origin (#37).  Is this still correct?

cryos:
This is fairly independent of the other issue, we need to be able to interactively move datasets relative to each other. For example, if I have two volumes loaded I want to display them both in the same view, then translate and rotate the active volume in the scene (the other volumes would remain static).

mathturtle:
Progress on this: I have a branch that lets you grab *some* representations of datasets and drag the datasets around.  I'm not getting mouse motion events yet though.  It updates when you release the button.  But there are issues with picking that make this very difficult to use.

* Using ParaView's picking neither the orthogonal slice nor the volume rendering can be selected.  And our non-ortho slice is not even a proxy, so ParaView's picking can't pick it at all.  I'm using this right now.
* Using VTK's picking, we still can't pick volumes.  The orthogonal slice can be picked... but it is much harder to take the vtkProp and get the datasource that needs to be moved.

I also found a bug in the orthogonal slice actor: it ignores position set on the actor so even if you move all the others this won't move.  Tracking this down now.

mathturtle:
Waiting on this ParaView bug too for translating slice representations: http://www.paraview.org/Bug/view.php?id=15891

mathturtle:
ParaView bug filed for slice representation refusing to translate:
http://www.paraview.org/Bug/view.php?id=15891

cryos:
Thanks for the updates on issues, we will need to figure out how to support volume picking.

cryos:
We should get a better icon for this.

mathturtle:
We have a better icon now.  Can we close this now or are there other things to fix?  I know the "don't require the datasource to be selected" is still outstanding.

cryos:
I would like to keep this open to track that outstanding tweak, ideally this should be finished before the release is tagged.

mathturtle:
The last PR has been merged and you can move objects whenever the mode is enabled, regardless of what is selected.  Can we close this or are we waiting for something else?

PR #75, opened 2015-06-08
Add a custom behavoir to keep center of rotation at origin - fixes #37

### LINKED ISSUE #37: Fix center of rotation to (0,0,0)

mathturtle:
@Hovden Is this what you had in mind?  #38 probably won't make it into 0.5.  Do you still want this in?

@cryos

### DIFF
--- a/tomviz/Behaviors.cxx
+++ b/tomviz/Behaviors.cxx
@@ -25,6 +25,7 @@
 #include "pqStandardViewFrameActionsImplementation.h"
 #include "pqViewStreamingBehavior.h"
 #include "ProgressBehavior.h"
+#include "FixCenterOfRotationBehavior.h"
 //#include "ScaleActorBehavior.h"
 #include "vtkSMSettings.h"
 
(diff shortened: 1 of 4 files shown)


Hovden:
We want to be able to move the center of rotation and translate each data object (with its Viz Modules) individually. There are several ways to do this:

1) The center of rotation can be translated and data objects can be translated. This is the most generic scenario (and probably what we want). Here objects and/or the center of rotation can be moved. Moving the center of rotation will also have the added effect of translating the camera. If this camera translation is not wanted, then the user can just leave the center of rotation alone or reset it by clicking the existing "reset center".

2) Hold the center of rotation constant. Also allow the user to translate each data objects relative to the this center of rotation. In this scenario, the rotation center is always in the center of the viewing panel unless the user translates the camera (this can be done by going into 2D viewing mode). This will also require removal of the "reset center" and "pick center" functionality.

Fixing the center of rotation alone only adds limitation to the visualization. I don't think we want this.

This discussion was a little complex and needs input from @cryos and @utkarshayachit. There may be aspects I'm overlooking.

PR #77, opened 2015-06-10
Change user menu text to "Downsample & Clone" - Hovdenbranch

Hovden:
User menu text has been changed to read "Downsample & Clone"

The algorithm does not allow all resampling but only permits downsampling. It also clones the data instead of adding it to the pipeline. The text now matches the functionality.


cryos:
Looks good to me.

PR #83, opened 2015-06-19
Add cropping  using box widget

### LINKED ISSUE #36: Adjustable cropping box interface
Pick the subvolume to be opened, only load the subvolume into memory. Provide graphical interface to move the corners etc.

cjh1:
@cryos This is ready for testing/review. I wasn't sure what todo with Shawn's WIP commits so I left them as is.  Let me know if you run into any issues. Fixes #36 

@cryos Ok fixed up

### DIFF
--- a/tomviz/CMakeLists.txt
+++ b/tomviz/CMakeLists.txt
@@ -36,6 +36,8 @@ set(SOURCES
   CentralWidget.h
   CloneDataReaction.cxx
   CloneDataReaction.h
+  CropDialog.cxx
+  CropDialog.h
   CropReaction.cxx
   CropReaction.h
   CropWidget.cxx
@@ -136,6 +138,7 @@ endforeach()
 qt4_wrap_ui(UI_SOURCES
   AboutDialog.ui
   CentralWidget.ui
+  CropDialog.ui
   DataPropertiesPanel.ui
   EditPythonOperatorDialog.ui
   MainWindow.ui
--- a/tomviz/CMakeLists.txt
+++ b/tomviz/CMakeLists.txt
@@ -36,6 +36,10 @@ set(SOURCES
   CentralWidget.h
   CloneDataReaction.cxx
   CloneDataReaction.h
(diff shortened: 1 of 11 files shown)


cjh1:
@cryos I have squashed to the WIP commits into one.

cryos:
Really minor comments inline, just about to test it out, the interaction you showed me looked great. Thanks for working on this.

cryos:
I think this is ready to merge if you can take a look at the minor issues, naming isn't a big deal to be honest for the CropWidget class. We need to do a little heavier work to properly add this operation to the pipeline, but that should be done in a new feature branch.

Hovden:
Very nice interactivity with the crop box. It's nice to see the bounds changing as the box is adjusted and vice versa. The user can leverage the visualization and still fine tune the crop by entering numbers.

Couldn't break it, or cause a crash on my OSX compile. When the crop box is graphically translated (clicking the center ball) completely outside the bounds, it behaves un-intuitively. This is a very minor issue caused from user error.

This crop tool will let us intelligently chop down our big datasets (2k^3) into something more manageable.  It's one of the first things we do.

cjh1:
We you say behaves un-intuitively, you mean that you can select a region outside the extent? Currently the widget doesn't have away of restricting the selection to a certain extent, but I think this could be added if necessary.

Hovden:
When the crop box is shifted such that it partially overlaps (only half the extents are outside the data) it behaves as expected. It will keep only the region where the crop box and data box overlap. I think this is a nice feature. For example, I can very quickly crop a region by snagging the center box and moving it.

When the crop box is completely outside the data volume (two extents are outside the data) it will crop the data down to the edges. I would have expected it to crop nothing, since there is no overlap of the two boxes. That being said, it doesn't make sense to set the crop box outside of the volume - it would be a user error.

I don't think any restrictions are necessary.

cjh1:
I have raise a PR that should resolve this issue.

PR #85, opened 2015-06-23
Use data extent if crop volume doesn't intersect

cjh1:
@Hovden Try this out it should resolve the issue you where seeing.


Hovden:
Perfect. Works great.

PR #96, opened 2015-06-25
Revert "tomviz: shorten the RC command line length"

cryos:
This reverts commit ece6d30fa1dc7751f63ee2571e181ab73b120b48. The
executable was no longer showiong an icon, reverting to verify this
commit introduced the issue.


cryos:
@mathstuf reverting as this seems to have caused issues, will revisit if path length becomes an issue.

PR #106, opened 2015-06-29
Automatic axis finding python transform

ElliotPadgett:
Includes python transform code for automatically finding in-plane rotation of tilt series relative to axis.  Merging will probably require some cherry-picking of the relevant parts.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,8 +8,8 @@ include(BuildLocation)
 include(CompilerFlags)
 
 set (tomviz_version_major 0)
-set (tomviz_version_minor 4)
-set (tomviz_version_patch 1)
+set (tomviz_version_minor 5)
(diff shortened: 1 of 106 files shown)


Hovden:
This code uses:
import matplotlib.pyplot as plt

However, matplot lib use is commented out. Since matplotlib is currently not included in tomviz, we should remove it.

ElliotPadgett:
Sorry, that's some testing code that I forgot to clean out. Thanks for spotting it.  I will fix it.

ElliotPadgett:
I removed the matplotlib code and includes.

cryos:
Pull request #145 cherry picks the two new commits, fixes the author name in the git commit, and supersedes this branch that contains divergent history of a previous branch that we fixed up before merging.

PR #108, opened 2015-06-29
Made align widget default view match data size

### LINKED ISSUE #107: Manual alignment: fit default field of view to data size
Now the widget starts at a fixed view width regardless of the data size.  For larger data sets, this makes the default widget state zoomed in significantly.

The widget should make its field of view match the size of the data.

ElliotPadgett:
The align widget now starts out viewing the whole data set.  This fixes #107 .


cryos:
LGTM, will merge once we have the 0.5.1 release out as it requires SciPy.

PR #109, opened 2015-06-29
Rotate Data With Python

Hovden:


cryos:
LGTM, will merge once we have the 0.5.1 release out as it requires SciPy.

PR #112, opened 2015-07-07
Data file paths are now saved as relative paths - fixes #72

### LINKED ISSUE #72: State files should use relative paths for data
State files are storing absolute file paths for data.  If a state file (dir1/state) and data (dir1/data) are packaged in the same directory (dir1), then the entire directory is copied (to dir2) and dir1 is renamed or deleted, opening dir2/state fails and results in an error saying "Initialize: could not open file dir1/data".  Relative file paths should be used instead, so that Tomviz tries to open dir2/data.

This is a key issue for scientific reproducibility, as it prevents state files from being shared across computers.

mathturtle:
Paths are relative to the directory that the state file is saved in.  Filenames are found by looking for this:
```
<Property name="FileNames" ...> <!-- It checks for both FileNames and FileName -->
    <Element name="value" value="/path/to/file" ...></Element>
    ...
</Property>
```

@cryos @utkarshayachit This covers all the cases I could find.  There may be others that I am missing though where the file name property is called something else, but I couldn't find any in the state files I created.

### DIFF
--- a/tomviz/SaveLoadStateReaction.cxx
+++ b/tomviz/SaveLoadStateReaction.cxx
@@ -24,6 +24,7 @@
 #include "vtkSMProxyManager.h"
 
 #include <QtDebug>
+#include <QDir>
 
 namespace tomviz
 {
@@ -93,7 +94,8 @@ bool SaveLoadStateReaction::loadState(const QString& filename)
     return false;
     }
 
-  if (ModuleManager::instance().deserialize(document.child("tomvizState")))
+  if (ModuleManager::instance().deserialize(document.child("tomvizState"),
+                                            QFileInfo(filename).dir()))
     {
     RecentFilesMenu::pushStateFile(filename);
     return true;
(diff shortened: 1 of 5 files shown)


tomviz/SaveLoadStateReaction.cxx
```diff
@@ -113,7 +115,9 @@ bool SaveLoadStateReaction::saveState(const QString& filename)
         .arg(vtkSMProxyManager::GetVersionMinor())
         .arg(vtkSMProxyManager::GetVersionPatch()).toLatin1().data());
 
-  return (ModuleManager::instance().serialize(root) &&
+  QFileInfo info(filename);
+
+  return (ModuleManager::instance().serialize(root,info.dir()) &&
```
cryos: Missing space after ,

tomviz/Utilities.cxx
```diff
@@ -30,12 +30,66 @@
...
+  // This pugi::xml_tree_walker converts filenames in the xml to and
+  // from relative paths.
+  class XMLFileNameConverter : public pugi::xml_tree_walker
+  {
+  public:
+    XMLFileNameConverter(const QDir& dir, bool rel)
+      : rootDir(dir), toRelative(rel) {}
+    virtual bool for_each(pugi::xml_node& node)
+    {
+      if (strcmp(node.name(),"Property") != 0)
```
cryos: Missing space after ,

cryos:
Can't test right now, but looking through the code this looks reasonable to me. Not sure it Utkarsh has any thoughts, happy to merge this and we can test the state files more extensively.

cryos:
Pointed out minor whitespace issues inline.

mathturtle:
Fixed minor whitespace issues.

cryos:
Great, thanks. Looks good to me.

PR #116, opened 2015-07-09
Warning cleanup

mathturtle:
@cryos I got sick of scrolling through pages of compiler warnings to find errors.  Now that we ignore warnings from the VTK/ParaView headers, I can clean up the remaining warnings.  This branch cleans up all warnings given by gcc 4.8.4 on my machine except the extra ';' warnings from the VTK code style used in vtkNonOrthoImagePlaneWidget.


cryos:
Looks good to me, I think the VTK style is wrong and so we should not conform to that pieces of style - it is easy to fix, and unlikely to be fixed in VTK to let's be more correct in tomviz.

PR #118, opened 2015-07-10
Fixed the data extents on the data shift GUI. Now allows full shift r…

Hovden:
…ange.

Previous limit was 0 to 99. Now it is 0 to end of data range.


cryos:
Looks good to me.

PR #120, opened 2015-07-14
Loads views into the correct layouts -- fixes #21

### LINKED ISSUE #21: Viz panels and tabs not loading properly from save State File
When saving and loading the tomviz state, the panels and tabs are not restored properly. For example, if one has a tab with three split viz panels and a second tab containing a single viz panel the state will restore with four viz panels in a single tab. 

Also, the size / spacing of the panel splits are not saved.

mathturtle:
@cryos @utkarshayachit This works... I know the code is a bit of a mess though.  tomviz was doing the minimum required to "load" views and layouts.  This change forces the views and layouts through the ParaView load state pipeline so that they are loaded correctly.

### DIFF
--- a/tomviz/ModuleManager.cxx
+++ b/tomviz/ModuleManager.cxx
@@ -21,11 +21,16 @@
 #include "Utilities.h"
 #include "ActiveObjects.h"
 
+#include "pqApplicationCore.h"
+#include "pqActiveObjects.h"
 #include "pqDeleteReaction.h"
 #include "vtkNew.h"
+#include "vtkPVXMLParser.h"
+#include "vtkPVXMLElement.h"
 #include "vtkSmartPointer.h"
 #include "vtkSMProxyIterator.h"
 #include "vtkSMProxyLocator.h"
+#include "vtkSMProxyManager.h"
 #include "vtkSMSessionProxyManager.h"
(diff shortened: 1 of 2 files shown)


tomviz/ModuleManager.cxx
```diff
@@ -317,58 +327,73 @@ bool ModuleManager::deserialize(const pugi::xml_node& ns, const QDir& stateDir)
...
+  if (!parser->Parse(stream.str().c_str()))
+    {
+      return false;
+    }
+  pqActiveObjects* activeObjects = &pqActiveObjects::instance();
+  pqServer *server = activeObjects->activeServer();
+
+  pqApplicationCore::instance()->loadState(parser->GetRootElement(),
+                                           server);
+  return true;
```
utkarshayachit: You're missing a `disconnect` here. I'd also cleanup the `Internals->node` and `Internals->dir` here since they are no longer applicable once this method returns.

utkarshayachit:
Besides the suggestions above, this looks good.

cryos:
I don't see the suggestions - were they addressed? I don't know this code well, and will defer to @utkarshayachit on the best solution.

utkarshayachit:
+2

PR #121, opened 2015-07-15
Adds context menus and keyboard shortcuts to delete modules -- fixes #14

### LINKED ISSUE #14: Delete data and transforms is non-obvious
“Delete data and modules” right now is hidden in the “data transforms” menu. This should be somewhere more obvious. It was one of the first things I looked for in the interface since I loaded my data wrong several times and wanted to clear it without restarting. I expect this will be a common experience.
--Elliot

Proposed solution: Users should be able to delete the data from the pipeline. This would include pressing the delete key and possibly right clicking and selecting 'delete'.

mathturtle:
@cryos @Hovden

### DIFF
--- a/tomviz/CentralWidget.cxx
+++ b/tomviz/CentralWidget.cxx
@@ -56,7 +56,6 @@
 Q_DECLARE_METATYPE(vtkSmartPointer<vtkImageData>)
 Q_DECLARE_METATYPE(vtkSmartPointer<vtkTable>)
 
-
 namespace tomviz
 {
 
@@ -275,8 +274,8 @@ CentralWidget::CentralWidget(QWidget* parentObject, Qt::WindowFlags wflags)
   this->Worker->start();
   this->HistogramGen->moveToThread(this->Worker);
   this->connect(this->HistogramGen,
-     SIGNAL(histogramDone(vtkSmartPointer<vtkImageData>,vtkSmartPointer<vtkTable>)),
-     SLOT(histogramReady(vtkSmartPointer<vtkImageData>,vtkSmartPointer<vtkTable>)));
+     SIGNAL(histogramDone(vtkSmartPointer<vtkImageData>, vtkSmartPointer<vtkTable>)),
+     SLOT(histogramReady(vtkSmartPointer<vtkImageData>, vtkSmartPointer<vtkTable>)));
 }
 
 //-----------------------------------------------------------------------------
@@ -375,7 +374,7 @@ void CentralWidget::histogramReady(vtkSmartPointer<vtkImageData> input,
 
   // If we no longer have an active datasource, ignore showing the histogram
   // since the data has been deleted
-  if (! this->ADataSource )
+  if (!this->ADataSource)
     return;
 
   vtkTrivialProducer *t = vtkTrivialProducer::SafeDownCast(
--- a/tomviz/CentralWidget.cxx
+++ b/tomviz/CentralWidget.cxx
@@ -53,6 +53,10 @@
 # include "dax/ModuleStreamingContour.h"
 #endif
 
+Q_DECLARE_METATYPE(vtkSmartPointer<vtkImageData>)
+Q_DECLARE_METATYPE(vtkSmartPointer<vtkTable>)
+
+
 namespace tomviz
 {
 
@@ -132,26 +136,32 @@ void PopulateHistogram(vtkImageData *input, vtkTable *output)
   output->AddColumn(populations.GetPointer());
 }
 
-// Quick background thread for the histogram calculation.
-class HistogramWorker : public QThread
+class HistogramMaker : public QObject
 {
   Q_OBJECT
 
   void run();
 
 public:
-  HistogramWorker(QObject *p = 0) : QThread(p) {}
+  HistogramMaker(QObject *p = 0) : QObject(p) {}
 
-  vtkSmartPointer<vtkImageData> input;
-  vtkSmartPointer<vtkTable> output;
+public slots:
+  void makeHistogram(vtkSmartPointer<vtkImageData> input,
+                     vtkSmartPointer<vtkTable> output);
+
+signals:
+  void histogramDone(vtkSmartPointer<vtkImageData> image,
+                     vtkSmartPointer<vtkTable> output);
 };
 
-void HistogramWorker::run()
+void HistogramMaker::makeHistogram(vtkSmartPointer<vtkImageData> input,
+                                   vtkSmartPointer<vtkTable> output)
 {
   if (input && output)
     {
     PopulateHistogram(input.Get(), output.Get());
     }
+  emit histogramDone(input, output);
 }
 
 class CentralWidget::CWInternals
@@ -228,9 +238,13 @@ bool vtkChartHistogram::MouseDoubleClickEvent(const vtkContextMouseEvent &m)
 CentralWidget::CentralWidget(QWidget* parentObject, Qt::WindowFlags wflags)
   : Superclass(parentObject, wflags),
     Internals(new CentralWidget::CWInternals()),
-    Worker(NULL)
+    HistogramGen(new HistogramMaker),
+    Worker(new QThread(this))
 {
   this->Internals->Ui.setupUi(this);
+
+  qRegisterMetaType<vtkSmartPointer<vtkImageData> >();
+  qRegisterMetaType<vtkSmartPointer<vtkTable> >();
 
   QList<int> sizes;
   sizes << 200 << 200;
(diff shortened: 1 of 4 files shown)


Hovden:
Visualizations persist even after deleting the parent data from the context menu. Context menus behave properly when deleting individual visualization modules.

To repeat:
1) Open data, after which a data icon and two viz modules appear
2) Right click on the data icon in the pipeline and click delete
3) Visualizations persist but pipeline is empty. Software will lockup when quitting.

cryos:
Thanks for testing Rob, we should fix this bug before merging.

mathturtle:
Thanks for testing Rob.  I've pushed a new version that fixes the bug you found.

Hovden:
Works on my OSX.

Both delete keys or context menus are responsive.
Things can be deleted in any order.
Deleting the dataset deletes all of the viz modules and data transforms.

mathturtle:
Now I've found a bug.  If you load data and immediately delete it before the histogram is loaded, then the data will appear to delete correctly.  However if you wait a few seconds, the histogram of the data will appear anyway.  Is this worth fixing on this branch @cryos?  Or accept the current branch and open a new issue?

utkarshayachit:
Let's fix it. You may just need to check before painting the histogram if the datasource for which we're showing the histogram is still alive. You could also look into aborting the histogram generation on delete.

mathturtle:
I redid the background thread creating the histogram to properly respect Qt thread object ownership while I was at it.  This fixes the bug I found.

Hovden:
@mathturtle I've tried testing on my OSX build and I'm getting an error when using the slicer. If this is something I'm doing wrong, then merge after successful testing. Cornell people can use and test the nightly builds while time still remains.

ERROR: In /Users/hovden/Research/Software/TomViz_working/ParaView/VTK/Rendering/OpenGL2/vtkOpenGLPolyDataMapper.cxx, line 2524
vtkOpenGLPolyDataMapper (0x7fdfb21ae800): failed after BuildBufferObjects 1 OpenGL errors detected
  0 : (1281) Invalid value

mathturtle:
Can you describe more specifically what you are doing to get the error?

Hovden:
When I load the non-ortho slicer on the error appears. I can interact with the slicer, but new errors appear.

Aside from the error, everything seems to work correctly.

screen shot 2015-07-22 at 6 35 49 pm

tomviz/CentralWidget.cxx
```diff
@@ -53,6 +53,10 @@
 # include "dax/ModuleStreamingContour.h"
 #endif
 
+Q_DECLARE_METATYPE(vtkSmartPointer<vtkImageData>)
+Q_DECLARE_METATYPE(vtkSmartPointer<vtkTable>)
+
+
```
cryos: Just leave one newline

tomviz/CentralWidget.cxx
```diff
@@ -257,14 +271,26 @@ CentralWidget::CentralWidget(QWidget* parentObject, Qt::WindowFlags wflags)
 
   this->EventLink->Connect(chart, vtkCommand::CursorChangedEvent, this,
                            SLOT(histogramClicked(vtkObject*)));
+
+  this->Worker->start();
+  this->HistogramGen->moveToThread(this->Worker);
+  this->connect(this->HistogramGen,
+     SIGNAL(histogramDone(vtkSmartPointer<vtkImageData>,vtkSmartPointer<vtkTable>)),
```
cryos: Missing space after comma here and line below

tomviz/CentralWidget.cxx
```diff
@@ -257,14 +271,26 @@ CentralWidget::CentralWidget(QWidget* parentObject, Qt::WindowFlags wflags)
...
 
 //-----------------------------------------------------------------------------
 CentralWidget::~CentralWidget()
 {
-  if (this->Worker && this->Worker->isRunning())
+  // disconnect all signals/slots
+  QObject::disconnect(this->HistogramGen, NULL, NULL, NULL);
+  QObject::connect(this->HistogramGen, SIGNAL(destroyed()),
+                   this->Worker, SLOT(quit()));
+  QMetaObject::invokeMethod(this->HistogramGen, "deleteLater");
```
cryos: Why aren't you just calling deleteLater on this class? Not sure why the need to use invokeMethod here.

tomviz/CentralWidget.cxx
```diff
@@ -53,6 +53,10 @@
 # include "dax/ModuleStreamingContour.h"
 #endif
 
+Q_DECLARE_METATYPE(vtkSmartPointer<vtkImageData>)
```
cryos: What is the motivation to pass vtkSmartPointer around rather than the pointer? I can see why we would do this with a C++ shared pointer object, but not sure I see the motivation here due to the internal reference counting in vtkObject.

tomviz/CentralWidget.cxx
```diff
@@ -325,23 +351,15 @@ void CentralWidget::setDataSource(DataSource* source)
...
-  else if (this->Worker->isRunning())
-    {
-    // FIXME: Queue, abort, something.
-    qDebug() << "Worker already running, skipping this one.";
-    return;
-    }
-  this->Worker->input = image;
-  this->Worker->output = table.Get();
-  this->Worker->start();
+  QMetaObject::invokeMethod(this->HistogramGen,
```
cryos: Again here I don't see why you don't just call the C++ method rather than using invokeMethod

tomviz/CentralWidget.cxx
```diff
@@ -349,15 +367,27 @@ void CentralWidget::refreshHistogram()
...
+void CentralWidget::histogramReady(vtkSmartPointer<vtkImageData> input,
+                                   vtkSmartPointer<vtkTable> output)
 {
-  if (!this->Worker || !this->Worker->input || !this->Worker->output)
+  if (!input || !output)
+    return;
+
+  // If we no longer have an active datasource, ignore showing the histogram
+  // since the data has been deleted
+  if (! this->ADataSource )
```
cryos: Don't need spaces either side of this->ADataSource.

cryos:
I added some comments and questions inline. I think it is mainly style questions along with the use of invokeMethod in a couple of places. Overall the patch looks good to me, but I would like to know a little more about why a few things were done the way they were. Thanks,

tomviz/CentralWidget.cxx
mathturtle: So the answer to several of your questions is here.  The worker thread is stared and then the HistogramGen object is moved to the other thread.  By Qt's threading rules the main (GUI) thread should not mess with its data anymore and slots triggered by Qt signals (or invokeMethod) will be run on the thread that owns the object.

tomviz/CentralWidget.cxx
mathturtle: Here, I want the method to be called on the background worker thread rather than the main event loop thread.  invokeMethod schedules the method to be called on the event loop of the thread that owns the object.  I don't want to call the method directly on the GUI thread since histogram creation is expensive.

mathturtle:
Here is an article that explains the general idea of how I redid the threading code:
https://mayaposch.wordpress.com/2011/11/01/how-to-really-truly-use-qthreads-the-full-explanation/

tomviz/CentralWidget.cxx
mathturtle: I'm not 100% sure you have to call deleteLater() on the thread that owns the object, but I'm playing it safe.

tomviz/CentralWidget.cxx
mathturtle: This was the issue with crashing when the application was shut down too quickly.  The background thread needs to own a reference to the vtk object, otherwise the VTK object can be deleted while the background object is still reading data from it to create the histogram.  I figured it was safer to pass the smart pointer to avoid potential race conditions.

mathturtle:
@Hovden Apparently the OpenGL errors you are seeing is an OpenGL2 error in ParaView/VTK.  I found the same thing on a lot of the ParaView dashboards.  I'm emailing the OpenGL2 maintainer about it.

mathturtle:
I added two new commits that I'll squash before merging: one to fix the whitespace issues pointed out and one to add comments explaining the new background thread mechanics I used.

@cryos How's this?

Hovden:
Merge for more testing in builds?

mathturtle:
Not yet.  I'll need to rebase to squash those fixup commits at least.  But I would like this to get merged soon.  I can't finish the context menus until this goes in.  Let me talk to @utkarshayachit, since @cryos is travelling today and unlikely to be watching email.

mathturtle:
Or we could leave the fixup commits...

mathturtle:
Ken just gave me a VTK patch to fix these OpenGL2 errors, more on that later as it moves through the review process to get into VTK/ParaView.

mathturtle:
Keeping it on this thread is silly... I'm going to open a new issue to track the error prints mentioned above and the fix for them.

PR #122, opened 2015-07-16
Tilt angles

mathturtle:
This is still definitely work in progress right now.  So far this branch does the following:

* Adds a type field to DataSource (it can be either VOLUME or TILT_SERIES)
  * All tilt series sources have a vtkFieldData array called 'tilt_angles'
* Adds a table to the data properties panel to view and edit tilt angles
* Adds a menu item to toggle the current data source between the tilt series and volume data set types

### DIFF
--- a/tomviz/DataSource.h
+++ b/tomviz/DataSource.h
@@ -37,9 +37,13 @@ class DataSource : public QObject
   typedef QObject Superclass;
 
 public:
+  /// The type of data in the data source.  The data types currently supported
+  /// are volumetric data and image stacks representing tilt series.
+  enum DataSourceType { Volume, TiltSeries };
   /// \c dataSource is the original reader that reads the data into the
   /// application.
-  DataSource(vtkSMSourceProxy* dataSource, QObject* parent=NULL);
+  DataSource(vtkSMSourceProxy* dataSource, DataSourceType dataType = Volume,
+             QObject* parent=NULL);
   virtual ~DataSource();
 
   /// Returns the data producer proxy to insert in ParaView pipelines.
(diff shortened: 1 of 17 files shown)


mathturtle:
@cryos @utkarshayachit 
Here is a start on handling tilt angles.  Since the above comment, tilt angles are saved and loaded with state if the data source is of type tilt series.  I'm about to remove the work in progress marker.  How does this look?

@Hovden This adds support for tilt angles in the GUI; you can view them and modify them but they don't do anything yet.

tomviz/DataSource.h
```diff
@@ -37,9 +37,13 @@ class DataSource : public QObject
   typedef QObject Superclass;
 
 public:
+  /// The type of data in the data source.  The data types currently supported
+  /// are volumetric data and image stacks representing tilt series.
+  enum DataSourceType { VOLUME, TILT_SERIES };
```
cryos: As this is more Qt/C++ it might be nice to consider using camel-case as these are not C macros. I think we can do this later though, and it doesn't need to be modified in order to get this merged now.

tomviz/MainWindow.ui
```diff
@@ -158,7 +160,7 @@
         <bool>true</bool>
        </property>
        <property name="columnCount">
-        <number>2</number>
+        <number>1</number>
```
cryos: This must be 2, why has this been changed to 1 again? This introduced a subtle bug before where we couldn't delete operators.

cryos:
The column count on the operators has been changed to 1 again, this took a little while to track last time after it was changed. This needs to be fixed before we merged, not sure if this is some kind of bug in the designer tool, if you are changing it on purpose please don't make unrelated changes without a commit message - the second column is for the delete action.

Once that is fixed I think this branch looks good enough to merge to me.

mathturtle:
I don't remember changing it... I suppose I might have, but I try not to manually edit the ui files.  Looking at the file again, it looks like the version of designer that I am using assumes the widget will have a `<column>` tag for each column and the number of `<column>` tags overrides `columnCount` (so since there is only one `<column>` tag it reset the number of columns to 1 when I loaded the file).  I'd go with this is a bug in Qt Designer version 5.2.1 for Ubuntu.  I'll fix it.

mathturtle:
Since I had to rebase anyways, I went ahead and fixed the camel-case thing too.

PR #130, opened 2015-07-23
Hide non orthogonal slice arrow

mathturtle:
This branch adds a checkbox to the Slice module to allow hiding the arrow (and disabling interaction with the widget similar to filters with widgets in ParaView).

@Hovden @utkarshayachit @cryos

PR #133, opened 2015-07-24
Fix non-orthogonal slice save state

mathturtle:
The non-orthogonal slice now saves the location of the slice plane and the visibility of the arrow (added in PR #130).  Since @Hovden wanted #130 merged, I merged that and based this off of the merge commit so that I could save the arrow visibility as well as the plane location.

@cryos @utkarshayachit


utkarshayachit:
+2

PR #143, opened 2015-07-29
Downsample x2 added using SciPy

Hovden:
Downsample by two is a common downsampling routine.

In the future we will also add a generic interpolation method for upsampling or downsampling - however this will probably be in addition to the "one-click" downsample x2.


cryos:
Looks good to me.

PR #145, opened 2015-07-30
Axis finding

cryos:


cryos:
@Hovden said he would like to hold off on this, but pushing the cleaned up branch that can be used as a basis for further modifications if desired.

cryos:
I am closing this as it no longer seems relevant, I will leave the branch in case we want to resurrect it.

PR #150, opened 2015-07-31
Save state with raw file now uses relative path

mathturtle:
During testing, Elliot found that the raw file reader uses a different
name for its file name property that was missed by the relative path
handling code.  This adds the property name 'FilePrefix' to the list of
properties that will be changed to relative path.


cryos:
Looks good to me!

PR #153, opened 2015-08-07
Ensure ITK is in the python path for installs

mathturtle:
@cryos If you know a better way to do this, let me know.  Only the superbuild knows about the path ITK will be installed at relative to tovmiz, so hardcoding it for now.


mathturtle:
Updated: ITK's location is different in MacOSX bundles that it will be on other systems.

PR #159, opened 2015-08-13
Disable save data menu item when no data is selected

mathturtle:
Fixes a segfault when no data is loaded and this menu item is clicked.

@cryos This prevents it from happening.  And I put in an assertion where the missing NULL check should be.


cryos:
LGTM

PR #161, opened 2015-08-13
Data properties panel: diplay data type

mathturtle:
@cryos I added a new label since putting the data type in the same line as the dimensions and range was too wide for the default panel width.  Well, for 'unsigned char' at least.


cryos:
Looks good to me, and will help spotting issues associated with type.

PR #163, opened 2015-08-14
Change output of reconstruct to be float not double

mathturtle:
I don't know why it was using doubles to begin with, but it is casting
the input to double deliberately.  However, double is an unsupported
format when writing tiff files, so cast the output back to float before
returning it.

@cryos

@Hovden This should prevent your tiff writing error.  I'm going to put in some error handling in the save data code too, but that will be another branch.

### DIFF
--- a/tomviz/python/Recon_DFT.py
+++ b/tomviz/python/Recon_DFT.py
@@ -93,7 +93,7 @@ def dfm3(input,angles,Npad):
     v = np.fft.ifftshift(v)
     recon = np.real(np.fft.ifftn(v))
     recon = np.fft.fftshift(recon)
-    return recon
+    return recon.astype(np.float32)
 
 
 def bilinear(kz_new,ky_new,sz,sy,N,p):


cryos:
Looks good to me. We should help the user more, and issue a warning if the volume is a type that is not well supported. There are other formats we are adding support for that will happily support doubles, so ultimately we may want to add a dialog to help more when saving data to formats with limited support for different types.

mathturtle:
I'm looking at trying to add that dialog right now, but it doesn't seem straightforward.  Essentially I need to bake in a huge table of what formats support which datatypes and then ask what to convert to...  And then the conversion will be a monstrosity using vtkTemplateMacro unless I can figure out a better way to do it.  For now I'll pop up a warning dialog.  New branch for that soon.

cryos:
I don't think that is the case, we are only concerned with a subset of file formats, with TIFF being a prime example. The warning dialog is a good start, we should talk before you start trying to "bake in a huge table" as I don't think that is necessary.

PR #165, opened 2015-08-14
Warn and cancel when writing a non-supported data type

mathturtle:
The tiff writer does not support double precision floating point and the
writer creates invalid files when trying to write this type.  Check for
this error condition and tell the user about the error.

@cryos

### DIFF
--- a/tomviz/SaveDataReaction.cxx
+++ b/tomviz/SaveDataReaction.cxx
@@ -25,6 +25,10 @@
 #include "pqProxyWidgetDialog.h"
 #include "pqFileDialog.h"
 #include "pqWriterDialog.h"
+#include "vtkDataObject.h"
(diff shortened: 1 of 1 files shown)


tomviz/SaveDataReaction.cxx
```diff
@@ -108,6 +114,21 @@ bool SaveDataReaction::saveData(const QString &filename)
     qCritical() << "Failed to create writer for: " << filename;
     return false;
     }
+  if (strcmp(writer->GetClientSideObject()->GetClassName(),"vtkTIFFWriter") == 0)
```
cryos: Missing space after comma

cryos:
Super minor whitespace issue, but patch looks good to me.

PR #167, opened 2015-08-14
Change assert to warning dialog -- fixes #162

### LINKED ISSUE #162: Error when running tomviz under 64 bit arch linux.
After compiling tomviz under 64 bit arch linux it threw the `SIGABORT` signal and exited.  This is the output of the command.

```
ASSERT: "ainfo->GetNumberOfComponents() == 1" in file /home/aaron/test/tomviz/tomviz/Utilities.cxx, line 191
fish: “./bin/tomviz” terminated by signal SIGABRT (Abort)
```

mathturtle:
The assert that number of components of an array is equal to one can
fail from user error.  This means it should not be an assertion, but a
warning to the user.

### DIFF
--- a/tomviz/Utilities.cxx
+++ b/tomviz/Utilities.cxx
@@ -32,6 +32,7 @@
 
 #include <QString>
 #include <QDir>
+#include <QMessageBox>
 
 namespace {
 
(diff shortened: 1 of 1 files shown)


cryos:
LGTM

PR #168, opened 2015-08-19
non-ortho slice fixes

mathturtle:
@Hovden @cryos

### DIFF
--- a/tomviz/vtkNonOrthoImagePlaneWidget.cxx
+++ b/tomviz/vtkNonOrthoImagePlaneWidget.cxx
@@ -314,7 +314,11 @@ void vtkNonOrthoImagePlaneWidget::SetEnabled(int enabling)
     }
 
     // Add the plane
-    this->CurrentRenderer->AddViewProp(this->PlaneOutlineActor);
+    if (this->PlaneProperty->GetOpacity() != 0 ||
+        this->SelectedPlaneProperty->GetOpacity() != 0)
+      {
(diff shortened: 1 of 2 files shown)


Hovden:
Compiles and fixes rendering problems.
screen shot 2015-08-19 at 2 24 02 pm

tomviz/vtkNonOrthoImagePlaneWidget.cxx
```diff
@@ -362,7 +366,7 @@ void vtkNonOrthoImagePlaneWidget::SetEnabled(int enabling)
     this->Interactor->RemoveObserver(this->EventCallbackCommand);
 
     // turn off the plane
-    this->CurrentRenderer->RemoveViewProp(this->PlaneOutlineActor);
+    //this->CurrentRenderer->RemoveViewProp(this->PlaneOutlineActor);
```
cryos: If this is no longer needed then this line and the previous line should be removed. If you think we might need it then let's resolve that before merging.

cryos:
I think you should either remove the code you commented out, or decide if it is still needed. Other than that question the patch looks reasonable to me, and if fixes an irritating bug.

mathturtle:
@cryos Technically it wouldn't matter since that actor is no longer added by default and we never change the defaults. But I uncommented it anyways.  Commenting it was from my earlier tests where I just disabled it entirely instead of checking if it is fully transparent.

cryos:
I just hate seeing commented code, and prefer to decide whether it be kept or not before merging. Thanks for the clarification, LGTM.

Issue #46, opened 2015-05-13
Make module and data properties context aware


cryos:
The module and data properties as implemented are confusing, it is not clear which module the module properties refer to, and the connection to the data is loose. We always have a selected data item or module - show an integrated properties dialog that shows the properties for the highlighted item.

mathturtle:
@cryos Do you want me to take a stab at this for 0.6.1?

cryos:
Sure, that would be great.

mathturtle:
Making a new panel that shows one or the other based on the selection isn't too hard, but I don't know exactly what you were planning as far as merging the two.  Which things should be visible in both cases?

Hovden:
There will be a single panel that uses the pipeline for context. When a visualization module is selected in the pipeline, the current  "module properties" panel is displayed in the panel. When a data module is selected in the pipeline, the current "data properties" panel is displayed in the panel.

This was our thought toward reducing the confusion of having a module and data properties panel.

If you have UX ideas to further help the user understand how they are modifying data and module properties, we should try it out.

cryos:
This patch has the expected behavior for me here - please take a look and let me know if you spot any issues.

cryos:
Closing this out, as I said in the pull request we may want to revisit panels in the future but I think this change makes it more obvious what you are looking at and saves space only having one of the other panels visible at any given time.

PR #169, opened 2015-08-25
Make the data/module properties context aware

cryos:
It is confusing to see the data properties and module properties when
only one of the two can be selected at any given time. Added some new
signals to indicate when a given item is active, and some logic to the
main window to switch in the appropriate panel.

This fixes issue #46.


mathturtle:
+1

Code changes look ok.  I've tested it and it works...

Hovden:
Also tested on OSX.

cryos:
Merged, thanks for taking a look. We may want to revisit the panels in general in the future, but this feels better than showing both and saves us some screen real estate.

Hovden:
Definitely cleans up the real estate noticeably.

Issue #173, opened 2015-08-27
Fatal crash when using data transform.


Hovden:
System crashes (OSX) when calling a simple transform (square root)

Reproduce:
1) Open sample tilt series
2) Data Transforms > Reconstruct
3) Data Transforms > Square Root

mathturtle:
Your sample tilt series wouldn't happen to have negative values would it?  I think I've seen a crash with applying sqrt to a dataset with negative values before...

Hovden:
You're right, I'll fix this in the Python code.

The system shouldn't crash though. Users will probably make mistakes when coding.

mathturtle:
Digging into the code a bit, it seems that the python code is correctly producing NaN when taking the square root of a negative number.  However, our histogram code is then converting this into an index of -2147483648.

My proposed solution to this is to have the create histogram code ignore NaN (or infinite to be thorough) floating point values.  Does that seem reasonable?

Hovden:
Ignore NaN is a good option.

mathturtle:
@Hovden I put up a branch to stop tomviz from crashing even if there is a NaN value in the dataset.

PR #177, opened 2015-08-28
Histogram computation now ignores NaN and infinite values in dataset

mathturtle:
This addresses the other half of #173.  If a python script results in a dataset with NaN or Inf/-Inf values, tomviz should not crash.  It now detects these when creating the histogram and ignores them.  This will slow histogram creation, but is better than crashing.  This is based before #174 was merged so that I could use sqrt to generate bad values to test with.


Hovden:
Cool. This is the correct solution. NaN values are permitted without crashing the system, however they appear odd (which is good) in the rendering.

Tested on OSX.

cryos:
Looks good to me.

PR #178, opened 2015-08-31
Limit available readers/writers

mathturtle:
@cryos This is part of my effort to get readers working the way we want them to.  Utkarsh suggested adding a mechanism to ParaView to limit the available readers for an application.  This depends on [this ParaView MR](https://gitlab.kitware.com/paraview/paraview/merge_requests/341).

I'm open to suggestions about where to store the list of readers we want available, they're just hard-coded in Behaviors.cxx for now.  With the change to ParaView, if no readers are specified you get the old behavior, but if *any* readers are specified this way, then *only* those readers will be available.  We can do the same thing with writers, I just need a list of which ones we care about.

### DIFF
--- a/tomviz/Behaviors.cxx
+++ b/tomviz/Behaviors.cxx
@@ -28,6 +28,7 @@
 #include "pqViewStreamingBehavior.h"
 #include "ProgressBehavior.h"
 //#include "ScaleActorBehavior.h"
+#include "vtkSMReaderFactory.h"
 #include "vtkSMSettings.h"
 
 #include <QMainWindow>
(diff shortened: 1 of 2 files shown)


tomviz/Behaviors.cxx
```diff
@@ -50,6 +51,12 @@ Behaviors::Behaviors(QMainWindow* mainWindow)
   : Superclass(mainWindow)
 {
   Q_ASSERT(mainWindow);
+  vtkSMReaderFactory::AddReaderOfInterest("sources","JPEGSeriesReader");
```
cryos: Missing spaces after , in function calls.

cryos:
AddReaderOfInterest sounds a little vague, very minor comment on whitespace. This is good, we can add more to the list, I might have gone more direct with AddAuthorizedReader or AddReaderToWhitelist, but reading the ParaView documentation it becomes clear what is meant by interest.

mathturtle:
You're probably right about the name for that function.  We could probably
still change it in ParaView (I like AddReaderToWhitelist) if you want.
We're still the only ones who use it.  @utkarsh.ayachit?

On Tue, Sep 1, 2015 at 11:30 AM, Marcus D. Hanwell <notifications@github.com

mathturtle:
Also, this branch is the base for #179, so it shouldn't need to get merged separately.  I finished #179 much faster than I thought.  (This was supposed to be an incremental step that would still be useful even without the custom plugin)

cryos:
I think it would be more immediately obvious what it does, so I would vote to change it if @utkarshayachit is open to it before others start using it. The functionality is great, just minor suggestions on API naming/whitespace. Thanks for pushing on this functionality!

mathturtle:
Ok, [function rename MR](https://gitlab.kitware.com/paraview/paraview/merge_requests/347) is up.  I'll update this branch.

PR #182, opened 2015-09-03
Fix a bug in "shift uniformly" to accept negative inputs

yijiang1:
Fix a bug in "shift uniformly" to accept negative inputs


mathturtle:
@yijiang1 Typically we wait to merge branches until someone else has looked over the change and tested it themselves to make sure there aren't corner cases that it fails to handle...

PR #183, opened 2015-09-03
Revert "Fix a bug in "shift uniformly" to accept negative inputs"

Hovden:
Reverts OpenChemistry/tomviz#182


Hovden:
Sorry about the merge and revert. Just getting Yi up to speed with the github process as we fix a bug he identified.

mathturtle:
@yijiang1 @Hovden Looking at the code he was fixing, does it really make sense to limit the user to a range of values to input?  And if so, I think something like +/-(extent[1]-extent[0]) with a default of 0 should be the new range.  (Extents can be negative too, so +/- the size of the image makes more sense as a shift range)

Hovden:
@mathturtle You are right that, for the shift operation, the extents should be +/-(extent[1]-extent[0]). This should be the same as +/-extent[1] since extent[0] is zero, however your suggestion is more generic and better.

It doesn't make sense for the user to be able to put in values beyond the range of the data, I slightly favor limiting the range.

mathturtle:
@Hovden Whether the extends are zero, positive or negative depends on the origin of the image and the spacing and what the reader chose to do with these three quantities.  I know of at least one reader that does not guarantee that extents[0] is 0.

Hovden:
Good to know. Really appreciate the necessary improvement.

PR #184, opened 2015-09-03
Fix a bug in shift uniformly

yijiang1:
I rewrote the code as @mathstuf suggested.


mathturtle:
Sorry about the confusion Ben... our usernames *are* similar.  He means me.

PR #185, opened 2015-09-04
Fix a bug in shift uniformly to accept negative inputs

yijiang1:


mathturtle:
+1

mathturtle:
@yijiang1 +1 is shorthand for 'I have looked over these changes and they seem alright'.  Generally I use +2 for 'I have looked at and *and tested* these changes and I approve'.

PR #186, opened 2015-09-04
Add a Sobel filter in Data Transform

yijiang1:
Add a Sobel filter in Data transform.
It applies scipy.ndimage.filters.sobel() function to a 3D dataset.
Sobel filter highlights high intensity variations in the image.


tomviz/python/SobelFilter.py
```diff
@@ -0,0 +1,16 @@
+def transform_scalars(dataset):
+    """Apply a Sobel filter to dataset."""
+    """Sobel Filter hgihtlights high intensity variations."""
```
mathturtle: nitpick: spelling *highlights

mathturtle:
This change makes sense to me.  I have no idea what a Sobel filter is, but if @Hovden thinks it is useful enough to warrant an item in the menu then this change looks good.

Hovden:
The Sobel filter has a non-isotropic kernal. The user needs to input the Sobel axis.

Currently this uses the default of Axis=-1, which I believe refers to the last axis dimension (however SciPy Libraries are poorly documented).

Hovden:
Note, the Laplace operator has an isotropic kernel.

yijiang1:
I added an UI, in which users now can choose the Sobel axis.

tomviz/AddPythonTransformReaction.cxx
```diff
@@ -219,6 +219,35 @@ OperatorPython* AddPythonTransformReaction::addExpression(DataSource* source)
...
   }
+  else if (scriptLabel == "Sobel Filter") //UI for Sobel Filter
+  {
+      QDialog dialog(pqCoreUtilities::mainWidget());
+      QHBoxLayout *layout = new QHBoxLayout;
+      QLabel *label = new QLabel("Axis:");
+      layout->addWidget(label);
+      QSpinBox *axis = new QSpinBox;
+      axis->setRange(0, 2);
+      axis->setValue(0);
```
mathturtle: If this represents axis, we should hide the numbers from the user.  They probably don't care that 0 is X, 1 is Y and 2 is Z.  We should probably make this a QComboBox with the items "X", "Y" and "Z" in the drop down.

Hovden: Agreed. I need to go back and do that for my "Rotate" code as well—probably set a bad example.

Hovden:
@mathturtle if everything looks OK to you, then I suggest we merge. I have put in a request for the qcombobox here #188 . Our qt programming experience is very low, so trivial things like comboboxes take some work.

mathturtle:
Compiles, code looks fine (I'll do the combo box after this is merged).

PR #192, opened 2015-09-08
Added Resample function in Data Transform

yijiang1:
Added a resample function, which calls scipy.ndimage.interpolation.zoom() and change the size of the dataset.
In UI, users can specify scale factors in each dimension.
The default factors are 1 (no scaling). The minimum input allowed is 0, which will produce an empty dataset but won't crash the program (on Mac).


Hovden:
Tested on OSX. Can't break yet ... yet.

PR #193, opened 2015-09-09
Use combo box for axis

mathturtle:
@Hovden @cryos This updates the Rotate and Sobel transforms to use a combo box of "X" "Y" and "Z" for the axis selection.  I also added a commit that fixes up the parents of all the widgets created in the AddPythonTransformReaction::addExpression function.  I'm not 100% sure it matters since they may be set automatically, but I'm not sure.

@yijiang1 This shows how to use a combo box for the axis if you have any more filters where you would like to do that.

### DIFF
--- a/tomviz/AddPythonTransformReaction.cxx
+++ b/tomviz/AddPythonTransformReaction.cxx
@@ -84,15 +84,15 @@ OperatorPython* AddPythonTransformReaction::addExpression(DataSource* source)
 
     QDialog dialog(pqCoreUtilities::mainWidget());
     QHBoxLayout *layout = new QHBoxLayout;
-    QLabel *label = new QLabel("Shift to apply:");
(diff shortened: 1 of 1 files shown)


Hovden:
I am getting the following error when running Rotate and Shift Uniformly (OSX):

QLayout: Attempting to add QLayout "" to QDialog "", which already has a layout
QLayout: Attempting to add QLayout "" to QDialog "", which already has a layout
QLayout::addChildLayout: layout "" already has a parent
QWidget::setLayout: Attempting to set QLayout "" on QDialog "", which already has a layout

mathturtle:
Hmmm. I'll take a look.

mathturtle:
@Hovden I goofed when setting parents (forgot that layouts have special rules). I've pushed a new version that should fix the errors  you saw.

Hovden:
Combo boxes work on OSX.

Issue #92, opened 2015-06-25
Slice Tool: Display Angles


Hovden:
In the slice tool properties panel, show the user the current angle of the slice plane.

mathturtle:
What format would be the most useful for this?  Angle of the plane doesn't make much sense to me: do you want the normal vector and a point on the plane?  Or something else?

I know, this is a 0.7 issue, but my 0.6 issues are done (pending bug reports) and I have to do something...

Hovden:
Great. 
The normal vector and the point on the plane would work.

mathturtle:
This should be fixed now.  Both the ParaView and tomviz patches are merged.

PR #195, opened 2015-09-10
Display slice plane

mathturtle:
This PR is work to address #92.  It displays a point on the plane and the normal vector of the non-orthogonal slice in the module properties panel.  @Hovden how is this?


Hovden:
This is great and very useful. Several researchers at Cornell are using tomviz for analysis daily and non-ortho slicing is popular.

Some minor comments:
1) When changing the point on plane, the slice is shifted and the render is not properly refreshed until the slice it is moved.
2) It is not clear which is the x,y,z coordinate.
3) There are a lot of significant digits displayed.

To solve 2&3, it may be worth breaking the everything into individual rows and labeling "x", "y", "z" to the left of each input box. The added space will accommodate larger boxes and better address point 3.

Hovden:
It would be sufficient to simply label x, y, z across the top and merge.

mathturtle:
@Hovden 

1) Good catch, I'll dig into this one.
2/3) This is going to require changing ParaView to support labeling the components of this kind of property.  It will actually be easier to put "x", "y" and "z" to the left of each input box than across the top.

mathturtle:
Ok, pushed a new version that fixes the update issue.  I was using the wrong function to update the plane.

Do you want to wait on the ParaView changes to add the coordinate labels before merging this?

Hovden:
I'm happy to merge if the code looks good. +1

mathturtle:
I have the ParaView changes working [here](https://gitlab.kitware.com/paraview/paraview/merge_requests/379).  They may be changed some in code review though.  Let's wait until they are finalized and I can add a 'use component labels' flag to those properties.  I suspect @utkarshayachit may want me to change how that flag is formatted.  Here is a preview:
component_labels

mathturtle:
We updated the ParaView code a bit so that the labels are aligned vertically with the property label on the left.  It still looks bad, but at least it is consistent.  I've added the commit here that will cause labels to appear when ParaView is updated.

New preview :(
component_labels2

mathturtle:
The ParaView fixes are merged and [this superbuild branch](https://github.com/OpenChemistry/tomviz-superbuild/pull/51) adds the fix.

tomviz/ModuleSlice.cxx
```diff
@@ -48,7 +48,9 @@ namespace tomviz
 {
 
 //-----------------------------------------------------------------------------
-ModuleSlice::ModuleSlice(QObject* parentObject) : Superclass(parentObject)
+ModuleSlice::ModuleSlice(QObject* parentObject)
+  : Superclass(parentObject)
+  , IgnoreSignals(false)
```
cryos: This looks odd, I think the comma should always be at the end of the line, not the start of the following.

tomviz/ModuleSlice.cxx
```diff
@@ -318,6 +324,12 @@ bool ModuleSlice::deserialize(const pugi::xml_node& ns)
 //-----------------------------------------------------------------------------
 void ModuleSlice::onPropertyChanged()
 {
+  // Avoid recursive clobbering of the plane position
+  if (this->IgnoreSignals)
+    {
```
cryos: Indentation

cryos:
Looks good to me, tiny niggle on comma, fix in the follow up.

PR #197, opened 2015-09-14
Make sure to install the tomvizExtensions library

mathturtle:
This fixes OSX binaries.  The Windows binaries require a superbuild change too.

### DIFF
--- a/tomviz/pvextensions/CMakeLists.txt
+++ b/tomviz/pvextensions/CMakeLists.txt
@@ -5,3 +5,9 @@ add_paraview_plugin(tomvizExtensions "1.0"
   SERVER_MANAGER_XML tomvizExtensions.xml
   SERVER_MANAGER_SOURCES
 )
+
+install(TARGETS tomvizExtensions
+        RUNTIME DESTINATION "bin"
+        LIBRARY DESTINATION "lib"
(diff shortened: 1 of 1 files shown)


tomviz/pvextensions/CMakeLists.txt
```diff
@@ -5,3 +5,5 @@ add_paraview_plugin(tomvizExtensions "1.0"
   SERVER_MANAGER_XML tomvizExtensions.xml
   SERVER_MANAGER_SOURCES
 )
+
+install(TARGETS tomvizExtensions DESTINATION lib COMPONENT runtime)
```
cryos: If we use the correct install command we should be able to address all platforms correctly in the install command. Something like,

install(TARGETS tomvizExtensions
    RUNTIME DESTINATION "bin"
    LIBRARY DESTINATION "lib"
    ARCHIVE DESTINATION "lib"
    COMPONENT "runtime")

Should install the everything in the right place, a DLL will go to the RUNTIME_DESTINATION while the .so/.dylib will be correctly placed in lib.

mathturtle:
Nice.  Your install command works on both platforms.  The cmake install documentation didn't really explain how to do that very well.

PR #200, opened 2015-09-15
Reorganize menus and add generated sample data

mathturtle:
Note: This branch *works* but I'm not happy with how the code is organized yet, please do not merge without asking me.  I'm putting this up so that @yijiang1 and @Hovden can try it out and give feedback on the UI and @cryos can give suggestions about improving the code.  I'd like to figure out a way to give parameters other than size to the dataset generators too.

### DIFF
--- a/tomviz/MainWindow.ui
+++ b/tomviz/MainWindow.ui
@@ -20,7 +20,7 @@
      <x>0</x>
      <y>0</y>
      <width>1024</width>
-     <height>30</height>
+     <height>27</height>
     </rect>
    </property>
    <widget class="QMenu" name="menu_File">
@@ -49,7 +49,7 @@
    </widget>
    <widget class="QMenu" name="menuModules">
     <property name="title">
-     <string>Visualization Modules</string>
+     <string>Visualization</string>
     </property>
    </widget>
    <widget class="QMenu" name="menuTools">
@@ -63,10 +63,7 @@
     <property name="title">
      <string>Data Transforms</string>
     </property>
-    <addaction name="actionAlign"/>
-    <addaction name="actionReconstruct"/>
     <addaction name="separator"/>
-    <addaction name="actionToggleDataType"/>
     <addaction name="separator"/>
     <addaction name="actionClone"/>
     <addaction name="actionDeleteData"/>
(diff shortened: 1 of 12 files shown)


yijiang1:
Thanks for quick update. I have a few suggestions:
1. The "Zero Dataset" function crashes my program (on Mac).
2. Could you move the "Mark As Tilt Series" function to the "Tomography" menu as well?
3. Could you disable functions in "Tomography" for non-tilt series dataset?
For 2 and 3, here is the UI we want to have:
When no dataset loaded:
nodata
 When dataset is Not marked as a tilt series:
nonts
When dataset is a tilt series:
ts

Thanks.

mathturtle:
It crashes?!  On OSX I get an error message about the dataset not having a filename and what look like some OpenGL errors but no crash.  Do you know how to get a stacktrace of the crash?  Can you run tomviz.app/Contents/MacOSX/tomviz directly on the command line and send me what it prints out? (this second one is not the stack trace, but it may be useful).

Menu items: Ok, easy enough.  I'll add a commit.

mathturtle:
I've updated the menu items to disable on non-tilt series data and moved the 'Mark As Tilt Series' menu item.

yijiang1:
@mathturtle I get a very long problem report like this
screen shot 2015-09-15 at 3 36 26 pm
Is this what you need?

mathturtle:
Hmmm.  I think it might have some useful information.  That might be a stack trace at the bottom of the screenshot.  Can you email me the contents of that scroll area?

mathturtle:
Ok, the warnings that I am getting saying 'this data source doesn't have a filename' are causing your tomviz to crash.  It is crashing when it goes to print the warning message.  Unfortunately I still have no idea why trying to print a warning message is causing it to crash.

mathturtle:
I figured out the crash with some help from Utkarsh.  I'll have a new version that works around it up soon.

mathturtle:
That was easier than I thought... This one shouldn't hit that crash.

yijiang1:
@mathturtle Yep, my program doesn't crashes anymore.

Hovden:
Everything seems to work on OSX. I tested the functionality of many menu functions including the zero data.

+1 (tested OSX)

Hovden:
@cryos or @utkarshayachit - OK to merge?

mathturtle:
Rebased to fix conflicts.  Pending @cryos's opinion, I think this is ready to merge.  I would have liked to put in a mechanism to send parameters to the data generation scripts, but that can wait for later.

tomviz/MainWindow.ui
```diff
@@ -160,7 +167,7 @@
         <bool>true</bool>
        </property>
        <property name="columnCount">
-        <number>2</number>
+        <number>1</number>
```
cryos: This column thing again - can you fix it up so that it stops changing this down to 1? Failing that just remove this change for now.

cryos:
One thing I would like fixing on the unintended change side, other than that it looks reasonable. I just glanced through, and we can certainly iterate on this once it is merged. Thanks Shawn.

tomviz/MainWindow.ui
mathturtle: I don't know.  When I set up my new machine I made sure to install only the Qt Designer that came with Qt 4.  It says version 4.8.6.  I really don't know why it keeps changing this.

mathturtle:
Updated to fix the column thing.

cryos:
Looks good - merging now.

Issue #189, opened 2015-09-08
Add a delete slice


Hovden:
Users should be able to delete a single slice from 3D data. This especially applies to tiltseries data.

Issue #199, opened 2015-09-14
Delete (zero) data cube for removing data in volume.


Hovden:
Jade is preparing a figures using tomviz and has requested a delete volume feature. I think it would be easy.

Graphically it would work just like the crop box, but it would replace the data inside the box with zeros instead of cropping. This can be done with a transparent python routine.

If the crop code is an easily callable function, we may use it in several upcoming python gui's.

mathturtle:
re: crop ui: It is pretty tightly coupled to the crop feature right now.  It doesn't have to be and I can refactor it to be a 'select volume of interest' widget that you could use in your python guis.  If this is a high-ish priority assign it to me and I'll look at it later this week.

Hovden:
@yijiang1 Now that a crop widget (#208) has been added, this feature (deleting selected data) would be a simple introduction to using this selection tool (which we will probably use for background subtraction too).

mathturtle:
I think we'll have to figure out how to make the python operator dialogs non-modal before this will be "simple".  I'll think about how to do this.

Hovden:
To be clear, this can be modal c++ code for now. It will operate just like the crop tool to get bounds then it calls python to delete the data within that range.

We don't need generic python code.

mathturtle:
Sure, it can be modal and will *mostly* still work.  But if the dialog is modal you will only get the six input boxes to type in numbers and not the widget to select the bounds interactively.

Hovden:
You're correct, my statement was wrong. It cannot be modal—my point was that it only needs to run one time and can be a pop-up prompt where the user clicks once for a one time operation.

PR #201, opened 2015-09-15
Added "Delete Slices" function

yijiang1:
For #189 , add "Delete slices" that allows users to delete a range of slices along a given axis.
If dataset is marked as a tilt series, the function also deletes corresponding tilt angles.
UI:
screen shot 2015-09-15 at 2 29 29 pm
The range for "start" and "End" Spinbox updates automatically based on axis.

### DIFF
--- a/tomviz/AddPythonTransformReaction.cxx
+++ b/tomviz/AddPythonTransformReaction.cxx
@@ -60,6 +60,19 @@ void AddPythonTransformReaction::updateEnableState()
         ActiveObjects::instance().activeDataSource() != NULL);
 }
 
+    
+//Update minimum value allowed for last slice
+void AddPythonTransformReaction::updateLastSliceMin(int min){
+    lastSlice->setMinimum(min);
+}
+
+//Update maximum value allowed for first and last slices
+//a=0,1,2 represents axis
+void AddPythonTransformReaction::updateSliceMax(int a){
+    firstSlice->setMaximum(shape[2*a+1]-shape[2*a]);
+    lastSlice->setMaximum(shape[2*a+1]-shape[2*a]);
+}
+
 //-----------------------------------------------------------------------------
 OperatorPython* AddPythonTransformReaction::addExpression(DataSource* source)
 {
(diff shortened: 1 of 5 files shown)


mathturtle:
I can confirm that it works... We will want to discuss how the UI is implemented.  Having variables for the UI of all the python data transforms in `AddPythonTransformationReaction.h` is not good design.  I'll need to think about how best to do this.  @cryos do you want to weigh in here?

mathturtle:
I've pushed a branch for how I would reorganize this to [my fork](https://github.com/mathturtle/tomviz/tree/deleteSlices).  Essentially, I created a custom widget class that takes care of selecting a range of slices.  This avoids filling up the reaction with fields and slots that are specific to certain operations and you can reuse this if you need to do 'select a range' again for another transform.

yijiang1:
@mathturtle I like that you grouped everything (first slice, last slice, and axis) into a class. 
I was wondering if it's better to have a class that just includes first slice and last slice? It may be used by other UI such as crop and reconstruction functions.

mathturtle:
Hmmm... I don't really see how that would be more useful.  If we need a version without the axis dropdown I can make that part of the UI optional later.  But for first and last slice to make sense you need an axis, even if it is hard-coded.  I don't see many cases where we would want that.

Crop already has a UI that handles 6 way bounds rather than just first and last slice.  I'm going to be pulling this out into a helper widget class that can be reused for other UIs soon.  @Hovden asked for it in #199.

yijiang1:
For crop: tomviz crashes (on OSX) if the start slice is greater than the end slice of the same axis, so we need to update the minimum value allowed in end slice.

For reconstruction: we might want to ask users to specify a range of slices they want to reconstruct. Since projections are stacked in z-direction by default, we no longer need to specify the axis. I'll be working on the UI for reconstruction once the new menu branch (#200 ) is merged.

mathturtle:
You found a new bug.  Can you open an issue about that crop widget crash?  I'll fix it at the same time that I work on #199 to factor out the 'select box' widget.

Re: for reconstruction: I'll add an option to hide the axis selector.  Since it already defaults to the z-axis, that should work fine for the reconstruction UI.  When I've pushed the new version, I'll make a new pull request for it.

mathturtle:
Closing since the PR #205 supersedes this and is merged.

Issue #198, opened 2015-09-14
Histogram / Colormap should not be zoomable


Hovden:
The colormap and histogram should not be zoomable, or at least not in the way it functions now.

I find myself losing view of the histogram / colormap because I accidentally zoom on my touchpad.

Hovden:
@mathturtle I say place a comment with the Paraview team letting them know that the colormap zooming is bad and should be turned off or improved. Then we can close this issue.

@utkarshayachit @cryos

mathturtle:
@Hovden I've actually got [this fix](https://gitlab.kitware.com/paraview/paraview/merge_requests/392#note_34966) most of the way through ParaView, just working out the last few issues.

mathturtle:
ParaView is having issues right now (someone merged something that broke most of the builds), so this is taking longer to get my change merged than I thought.

mathturtle:
This should be fixed in the latest binaries.

yijiang1:
Tested on OSX.

PR #202, opened 2015-09-16
Disable zooming on the histogram

mathturtle:
@Hovden This addresses part of #198.  Disabling zoom on the color map may be trickier.


cryos:
LGTM

Issue #137, opened 2015-07-28
Histogram does not update after scale is changed on OSX


Hovden:
On Mac OSX only: "Rescale to custom range" does not update the histogram colormap. However, if I move the curser of the histogram, it will refresh with the proper colormap.

From #127 :
"I can confirm that this histogram update issue happens on MacOS but not on my Linux machine... weird. However this happens regardless of the non-orthogonal slice being there. I'm building with the old ParaView now to see if this is a ParaView regression or just a bug somewhere.

This histogram update should be opened as a separate issue. I can confirm that the update to ParaView did not affect it. Building against older ParaView gave the same behaviour. When the colormap is updated via Rescale to Custom Range, the histogram update is not triggered until the user moves the mouse over the histogram."
-@mathturtle

mathturtle:
This may be another Qt4/OSX bug.  Debugging on linux, I found that the view was being refreshed automatically by Qt rather than by any events.  Whenever the window had focus, the view was getting refreshed.  If I did something silly like adjust the color map and switch applications I could reproduce it, but as soon as the tomviz window had focus, the view was automatically updated.

mathturtle:
I can't debug on OSX since my account on bigmac apparently is not in the 'Developer Tools' group?

mathturtle:
This should be fixed now.

Hovden:
Tested on OSX cdash build.

PR #203, opened 2015-09-16
Signal the histogram to update when the colormap is updated

mathturtle:
@Hovden can you try this and see if it fixes your histogram update issue on OSX? ( #137 )  It fixed it for me.


Hovden:
Nice. Things seem to be behaving properly after python transforms, new colormaps, and cropping are applied.

+1 (tested on OSX)

Issue #113, opened 2015-07-09
Histogram colormap transparency


Hovden:
For a little while, my local build showed the histogram with the colormap transparency applied. I found the transparency informative. Can we turn this on?

mathturtle:
This should probably be just an option (and not the default).  I'll take a look at how to add it.

mathturtle:
After some digging, I asked @utkarshayachit.  This happens when you turn on the 'Enable opacity mapping for surfaces' checkbox below the color map.  Note that it suffers from the same 'histogram doesn't update' issue as #137, even on Linux (I may be able to use this to debug that one).  I think we can close this one as 'Already fixed'.

Hovden:
I see. The opacity maps to volume rendering though.

The best solution is to have a context aware histogram. This will handle the different (local) colormaps (not just global) and also transparency. For example, when a contour is selected that has its own colormap (by checking "Use separate color map for this module") the histogram will display that colormap.

mathturtle:
The branch for what you suggested is merged now.  Is that it for this issue?

Hovden:
Very nice!

PR #204, opened 2015-09-17
Histogram use module's detached colormap

mathturtle:
This builds on #203 to add the changes discussed in issue #113 where the histogram uses a module's detached colormap when that module is selected. @Hovden @cryos


cryos:
LGTM

Issue #206, opened 2015-09-17
Crop crashes (on OSX) when start slice is greater than end slice.


yijiang1:
The Crop transform crashes (on OSX) when start slice is greater than end slice of the same axis.

PR #208, opened 2015-09-17
Factor out crop widget

mathturtle:
This changes the UI for the crop widget into a reusable widget that selects a volume as requested by #199.  It also fixes the crashes caused by inputing invalid volumes by preventing such volumes from being created by updating the opposite end of the dimension if the minimum is ever greater than the maximum (reported in #206).

@yijiang1 Can you test this and make sure it fixes it?

### DIFF
--- a/tomviz/CMakeLists.txt
+++ b/tomviz/CMakeLists.txt
@@ -44,8 +44,8 @@ set(SOURCES
   CropReaction.h
   CropOperator.cxx
   CropOperator.h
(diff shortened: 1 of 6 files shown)


yijiang1:
I tested it and the UI works properly now.

mathturtle:
@cryos If this looks good to you we should merge it.

cryos:
LGTM

PR #212, opened 2015-09-22
Cancelling 'change background color' now does nothing

mathturtle:
This fixes a bug where cancelling the change background dialog set the
background color to black.


Hovden:
Nice catch.

PR #214, opened 2015-09-24
Reorganize python sample data to allow custom UIs

mathturtle:
This should allow them to fairly simply add custom UIs for the python sample data.  It still feels pretty hacky though.

@cryos


mathturtle:
@Hovden @yijiang1 This is the branch that will allow custom GUIs for python generated sample data.  The last commit in the branch is a simple example of how to create these.

yijiang1:
LGTM. The update works well and now I can add new functions under sample data.

cryos:
LGTM

Issue #158, opened 2015-08-13
Add ITK


Hovden:

mathturtle:
The most recent binaries should have ITK bundled.

Hovden:
Great. Can you provide a simple example(s) to demonstrate and get us started with tinkering?

cryos:
If you can provide a simple example @mathturtle that would be great, we should also talk to Matt McCormick now that ITK is bundled about some fuller examples.

mathturtle:
I'm working on it.  I emailed Matt yesterday afternoon asking what some good filters to demonstrate might be.  We may have to update the bundled ITK though to turn more things on in its build process.

cryos:
Sounds like whilst it is packaged there are some teething issues. We will leave it in the binary so that people can experiment but I don't want to delay our bug fix release. I will move this to the 0.7 milestone as a tracker, and we can continue working on some examples.

mathturtle:
@Hovden here is a sample ITK script for segmentation that can be used in the tomviz 'custom python transform' dialog.  I don't know much about ITK, so this may not be optimal, but I think it works (as far as I can tell anyways).
```
def transform_scalars(dataset):
    """Define this method for Python operators that 
    transform the input array"""

    from tomviz import utils
    import numpy as np
    import itk

    # Get the current volume as a numpy array.
    array = utils.get_array(dataset)

    # Set up some ITK variables
    itk_image_type = itk.Image.F3
    itk_converter = itk.PyBuffer[itk_image_type]

    # Read the image into ITK
    itk_image = itk_converter.GetImageFromArray(array)

    # ITK filter (I have no idea if this is right)
    filter = itk.ConfidenceConnectedImageFilter[itk_image_type,itk.Image.SS3].New()
    filter.SetInitialNeighborhoodRadius(3)
    filter.SetMultiplier(3)
    filter.SetNumberOfIterations(25)
    filter.SetReplaceValue(255)
    filter.SetSeed((24,65,37))
    filter.SetInput(itk_image)
    filter.Update()


    # Get the image back from ITK
    result = itk.PyBuffer[itk.Image.SS3].GetArrayFromImage(filter.GetOutput())

    # This is where the transformed data is set, it will display in tomviz.
    utils.set_array(dataset, result)
```

Hovden:
Thanks. @yijiang1 and I will dig into what ITK can provide for segmentation.

yijiang1:
[...]

PR #215, opened 2015-09-25
Add an ITK segmentation module

mathturtle:
This adds a module that runs an ITK-python script and contours the
resulting data.  I'm just putting this up as something for people to play with right now.  A sample script that can be used with this module is below.  The contour doesn't default right yet...

```python
def run_itk_segmentation(itk_image, itk_image_type):
    # should return the result image and result image type like this:
    # return outImage, outImageType
    itk_filter = itk.ConfidenceConnectedImageFilter[itk_image_type,itk.Image.SS3].New()
    itk_filter.SetInitialNeighborhoodRadius(3)
    itk_filter.SetMultiplier(3)
    itk_filter.SetNumberOfIterations(25)
    itk_filter.SetReplaceValue(255)
    itk_filter.SetSeed((24,65,37))
    itk_filter.SetInput(itk_image)
    itk_filter.Update()
    return itk_filter.GetOutput(), itk.Image.SS3
```

### DIFF
--- a/tomviz/CMakeLists.txt
+++ b/tomviz/CMakeLists.txt
@@ -77,6 +77,8 @@ set(SOURCES
   ModuleOutline.h
   ModulePropertiesPanel.cxx
   ModulePropertiesPanel.h
+  ModuleSegment.cxx
+  ModuleSegment.h
   ModuleSlice.cxx
   ModuleSlice.h
   ModuleThreshold.cxx
--- a/tomviz/ModuleContour.cxx
+++ b/tomviz/ModuleContour.cxx
@@ -178,7 +178,7 @@ bool ModuleContour::serialize(pugi::xml_node& ns) const
   node = ns.append_child("ContourRepresentation");
   if (tomviz::serialize(this->ContourRepresentation, node, contourRepresentationProperties) == false)
   {
-    qWarning("Failed to serialize ContourFilter.");
+    qWarning("Failed to serialize ContourRepresentation.");
     ns.remove_child(node);
     return false;
   }
--- a/tomviz/ModuleFactory.cxx
+++ b/tomviz/ModuleFactory.cxx
@@ -23,6 +23,7 @@
 #include "ModuleContour.h"
 #include "ModuleOrthogonalSlice.h"
 #include "ModuleOutline.h"
+#include "ModuleSegment.h"
 #include "ModuleSlice.h"
 #include "ModuleThreshold.h"
 #include "ModuleVolume.h"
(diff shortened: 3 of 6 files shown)


Hovden:
Is this ready for testing? We can start playing with ITK.

mathturtle:
It is certainly ready for testing.  I'd like some feedback before I spend too much more time on it.  But it is definitely not in its final form and there are a few things I haven't implemented yet (like saving the script in the state file).  That said, this will make it much easier to start playing with ITK since you won't have to worry about any of the VTK to ITK conversion code.

Hovden:
Ok. I'm sitting down with Yi and Elliot and we're trying to understand the ITK documentation.

ITK installation via macports is failing: "Error: python25 is not supported on Yosemite or later." It appears that ITK is made for python25?

It's really tough to follow the c++ ITK docs, is there Python docs / tutorials. For example SetSeed() takes in a variable "const IndexType & 	seed" which is very cryptic. In your example however, it is simple a vector for the x,y,z coordinates.

Also, is this SimpleITK or ITK?

mathturtle:
Right now it is ITK (SimpleITK may be a possibility later).  It looks like the ITK macports recipe is really old and probably hasn't been updated since python 2.5 was the default of OSX.  It looks like homebrew may have an up-to-date recipe, but I'm not sure.  The easiest thing for you to do may be to grab the version I built for the superbuild [here](http://www.tomviz.org/files/itk-4.8.0-macosx10.8.tar.gz).  You'll have to set the PYTHONPATH and DYLD_LIBRARY_PATH so that python can find it.  I posted directions on this issue #158.

>It's really tough to follow the c++ ITK docs

I agree completely.  They use a lot of templates and typedefs to make the types "simpler".  I modified the code for my example from one of their C++ examples.

Hovden:
Thanks for the discussion today, really cleared up some things. No need for SimpleITK at this point. ITK looks really great, but the python docs are really sparse.

It would be nice to have to python ITK tutorials (we can do this to promote ITK and tomviz) and also update the macport build.

Should we merge this pull request so that we can test ITK on the tomviz build? Or wait till it is further developed?

cryos:
I think we can merge it if you feel it would be useful. I am not sure about the value of putting energy into macports, I had multiple issues and have found brew to be far superior. End users will get a binary with everything included.

mathturtle:
I don't have a problem with merging it, but I'd like to at least get the 'save to state file' part working first.  @cryos should probably code review it first too.

If you just want to test ITK, you should be able to do that from the 'Custom Transform' menu item too (this will give you access to a more diverse set of visualization techniques to look at the resulting data too).

mathturtle:
I've finished the serialization code now @cryos, so if you want to merge this it is fine with me.

cryos:
Looks good to me, merging it.

PR #220, opened 2015-09-28
Added sample data function that generates random particles

yijiang1:
Added a function that generates random structures such as nano particles.

The basic idea is to generate random phases in Fourier space and combine it with an exponential decay function as magnitude. The "decay rates" can be specified by users and control the complexity of the objects as well as their shape. Users can also specify the percentage of non-zero voxels. 

Because the function uses Numpy's FFT, the data size should be set to a power of two to maximize performance. Also, the maximum size allowed right now is 512 for each dimension, which takes more than 5 mins on my Mac (2.8 GHz Intel Core i7, 16 GB memory).

### DIFF
--- a/tomviz/PythonGeneratedDatasetReaction.cxx
+++ b/tomviz/PythonGeneratedDatasetReaction.cxx
@@ -227,9 +227,9 @@ public:
 //----------------------------------------------------------------------------
   void setSpinBoxValue(int x, int y, int z)
   {
-    this->xSpinBox->setValue(128);
-    this->ySpinBox->setValue(128);
-    this->zSpinBox->setValue(128);
+    this->xSpinBox->setValue(x);
+    this->ySpinBox->setValue(y);
+    this->zSpinBox->setValue(z);
   }
 //----------------------------------------------------------------------------
 
--- a/tomviz/PythonGeneratedDatasetReaction.cxx
+++ b/tomviz/PythonGeneratedDatasetReaction.cxx
@@ -38,6 +38,7 @@
 #include <QLabel>
 #include <QSpinBox>
 #include <QVBoxLayout>
+#include <QGridLayout>
 
 #include <limits>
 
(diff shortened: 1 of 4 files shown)


tomviz/PythonGeneratedDatasetReaction.cxx
```diff
@@ -328,6 +346,78 @@ void PythonGeneratedDatasetReaction::addDataset()
...
     this->dataSourceAdded(generator.createDataSource(shape));
   }
+  else if (this->Internals->scriptLabel == "Random Particles")
+  {
+    QDialog dialog;
+    dialog.setWindowTitle("Random Particles");
+    QVBoxLayout* layout = new QVBoxLayout; //overall layout
+    //Guide
+    QLabel *guide = new QLabel;
+    guide->setText("Generate many random 3D \"particles\" using the Fourier Noise method. You can increase the \"Internal Complexity\" of particles and their average \"Particle Size\". You can also specify the sparsity (percentage of non-zero voxels) of the generated dataset. Note: 512x512x512 may take a couple minutes to run.");
```
mathturtle: I would add a note that sizes that are powers of two will be faster to run too.

tomviz/PythonGeneratedDatasetReaction.cxx
```diff
@@ -328,6 +346,78 @@ void PythonGeneratedDatasetReaction::addDataset()
...
+    guide->setWordWrap(true);
+    layout->addWidget(guide);
+
+    //Shape Layout
+    ShapeWidget *shapeLayout = new ShapeWidget(&dialog);
+    shapeLayout->setSpinBoxValue(128,128,128);
+    shapeLayout->setSpinBoxMaximum(512,512,512);
```
mathturtle: Do we want to set the maximum or just warn the user that the operation will be extremely long-running (>5 minutes)?

@Hovden @cryos

Hovden: 1024 takes a long time (~1hr). It is not reasonable to let the user wait this long. 512 seems reasonable for now.

mathturtle: What if the user wants to create a non-cubic volume (say 1024x256x256)?

Hovden: True. For now, 512x512x512 is plenty given that we can upsample (interpolate) to larger volumes. Let's merge and move forward. We can raise a new issue and expand our approach if there is a demand for larger random volumes.

Issue #221, opened 2015-09-28
Crashes when applying data transform to sample data. 


yijiang1:
Tomviz sometimes crashes when applying data transforms to sample data.

One example:
1. Open Tomviz.
2. Click "Sample Data"->"Constant Data". Use default parameter.
3. Click "Data Transforms"->"Rotate". Use default parameter.

Note: 
1. This does not happen to real data loaded from tiff files.
2. Doing other transforms (for example, "Downsample by 2") before "Rotate" can prevent crashes.

mathturtle:
This may actually be a bug in scipy.  Here is the line of code causing the problem:
`from scipy import ndimage`
As far as @mathstuf and I can tell it is a binary incompatibility issue with python and scipy.  It is somehow asking for the data size of one of the scipy types and getting a pointer back (instead of the size variable which as far as I can tell is correctly set).  Then it can't allocate an array big enough and it crashes.  I have no idea what is going on here, since scipy is built against the same python I am using to run this.

mathturtle:
@cryos For some reason it is getting a pointer as the size of a scipy type... That's really about as far as I got with the debug package for python on linux.  PyObject_GC_Malloc is failing when it is passed an argument of: basicsize=140735892870160 (0x7fffa0e6d410).  The one fairly consistent way to work around this is to open the python console and import ndimage there.  Sometimes applying other transforms such as 'Downsample by 2' can avoid this.  And as Yi mentions, it doesn't seem to happen with real data loaded.

mathstuf:
Other notes:

 * It isn't a SciPy type; it's a `PyCFunction` which is defined in Python itself.
 * The pointer appears heap-allocated, but doesn't appear to be stored near the `PyCFunction_Type` object.
 * It occurs when adding the `fourier_filter` method to the `nd_image` module which is 9th in the list.

Has valgrind or asan/ubsan been run over the failing test case?

mathturtle:
When I forced it to use the superbuild numpy/scipy I think I got a slightly different stack.  I think yesterday it was using the pip installed versions in the virtualenv (which *shouldn't* be a problem...)  I didn't see the scipy module code in there anymore.  Valgrind had only one error in python and it is the invalid read that causes the segfault as far as I can tell.  But the getting a pointer for the size of the object is still the real problem according to my gdb run.

Valgrind error:
==12770== Invalid read of size 1
==12770==    at 0xC66EAE4: visit_decref (gcmodule.c:360)
[...]

PR #224, opened 2015-10-02
Added function that generates tilt series from volume

yijiang1:
Added a function that generates tilt series from volume.
The function is disabled for non-volume dataset. It automatically marks output as tilt series and save the tilt angles, which are specified by the user through UI.

Because the current implementation uses scipy's rotate function, it has same issue as the #221 for sample data. The crash can be avoided by applying other data transform (for example, shift uniformly) beforehand.

The function is very basic and slow but enough for me to test reconstruction functions at current stage. In the future we should reimplement it in C++.


mathturtle:
This looks fine to me.  I've just done a few basic tests that it works.

I haven't had much luck tracking down #221... I'll have to get back to you on that one.

Issue #82, opened 2015-06-19
Auto save state and restore when tomviz crashes.


Hovden:
Tomviz crashes frequently. To alleviate this issue, we should have an auto save state that runs every so often and restores the last known state in the event of a crash. It could prevent a user from losing hours of work.

cryos:
We also need to get recipes to recreate the crashes, and fix the issue. Autosave or not, most people will get frustrated with an application that "crashes frequently". I would like to prioritize getting crash reports so that we can work on fixing the underlying bugs.

Hovden:
Crash reporting would make crashes more transparent and noticed. We'll definitely report re-creatable bug recipes when noticeable. This would just be the ad-hoc "Microsoft Office" solution for everything we don't catch.

Hovden:
Basic use case tested on OSX and Windows 32-bit.

PR #225, opened 2015-10-05
Add autosave capabilities

mathturtle:
This is still pretty rough, but you autosave/load works.  It should
probably autosave more often (or maybe less, I was able to create an
autosave that would re-create the crash that it was recovering from...)

### DIFF
--- a/tomviz/MainWindow.cxx
+++ b/tomviz/MainWindow.cxx
@@ -72,9 +72,11 @@
 #include "ConstantDataset.h"
 #include "RandomParticles.h"
 #include "TiltSeries.h"
+
+#include <QDir>
 #include <QFileInfo>
 #include <QMessageBox>
-#include <QDir>
+#include <QTimer>
 
 //we are building with dax, so we have plugins to import
 #ifdef DAX_DEVICE_ADAPTER
@@ -102,6 +104,7 @@ public:
   Ui::MainWindow Ui;
   Ui::AboutDialog AboutUi;
   QDialog *AboutDialog;
+  QTimer *Timer;
 };
 
 //-----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


mathturtle:
This is to address #82.  There may still be a few issues to work out though.

Hovden:
Cool! Maybe less on the autosave would better solve the re-create crash problem—say every five or ten minutes?

mathturtle:
Do you think 5 minutes would be often enough?  I like the autosave timer idea and it should be simple enough to implement.  As a user how often do you think it should autosave?

Hovden:
My thought; the simplest away to avoid re-crashing is by extending the window of autosave. If a user loses 5 minutes of work that is not a big deal. I'm worried that a user will work for hours without saving and then tomviz crashes (this has happened to me).

How will we handle the long computations though? If there is no crash after completion it saves?

cryos:
I think a timer would be irritating, and is not something I have seen good examples of elsewhere. We could make it configurable, but five minutes seems reasonable to me.

cryos:
Unless you meant a QTimer, non-graphical, in which case - this is the right way to go.

mathturtle:
QTimer is what I thought of when timers were mentioned.  Long computations, as long as they are done on the event loop thread, should block the timer events so autosave will only run if the operation completes.

cryos:
Precisely, with a 5-10 minute interval being pretty reasonable. We should consider adding a settings dialog at some point, and it would be reasonable to expose the number of minutes between autosaves there.

mathturtle:
Rebased to avoid conflicts.  This branch has the 'save every 5 minutes' timer now.

mathturtle:
ping

cryos:
Looks good - let's merge it.

PR #226, opened 2015-10-06
Add python source information to state file

mathturtle:
This allows saving and loading of the state of tomviz when there are
python generated datasets present.  Previously the load state operation
would crash tomviz.

I found this while working on #225 and decided it was high priority.

### DIFF
--- a/tomviz/ModuleManager.cxx
+++ b/tomviz/ModuleManager.cxx
@@ -20,6 +20,7 @@
 #include "Module.h"
 #include "Utilities.h"
 #include "ActiveObjects.h"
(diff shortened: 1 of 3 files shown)


mathturtle:
@Hovden @cryos

cryos:
Looks good to me, haven't had a chance to test it, but I am good with merging it.

Hovden:
How do we test this?

mathturtle:
Do something with one of the new generated datasets in the sample data menu.  Save the state and try to load it again.  Without this patch trying to load it will crash.

mathturtle:
ping

PR #230, opened 2015-10-08
Add a 'Set Tilt Angles' menu item

mathturtle:
This menu item allows the tilt angles to be set by querying the user for
the first and last slice and angle and iterpolating the angle over the
intervening slices.

### DIFF
--- a/tomviz/SetTiltAnglesReaction.cxx
+++ b/tomviz/SetTiltAnglesReaction.cxx
@@ -1,0 +1,157 @@
+/******************************************************************************
+
+  This source file is part of the tomviz project.
+
+  Copyright Kitware, Inc.
+
+  This source code is released under the New BSD License, (the "License").
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+
+******************************************************************************/
+
+#include "SetTiltAnglesReaction.h"
+
+#include "ActiveObjects.h"
+#include "DataSource.h"
+#include "vtkDataArray.h"
+#include "vtkImageData.h"
+#include "vtkSMSourceProxy.h"
+#include "vtkTrivialProducer.h"
+
+#include <QDialog>
+#include <QDoubleSpinBox>
+#include <QGridLayout>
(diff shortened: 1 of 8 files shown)


tomviz/SetTiltAnglesReaction.cxx
```diff
@@ -0,0 +1,144 @@
...
+    QVector<double> tiltAngles = source->getTiltAngles();
+    int start = startTilt->value();
+    int end = endTilt->value();
+    if (end == start)
+    {
+      tiltAngles[start] = startAngle->value();
+    }
+    else
+    {
+      double delta = (endAngle->value() - startAngle->value()) / (end - start + 1);
```
yijiang1: The "End Angle" is not set to the last tilt. The angle increment should be 
delta = (endAngle->value() - startAngle->value()) / (end - start)

mathturtle: Hmmm. I think you're right.  I shouldn't be adding 1 to the difference in extents where I'm computing the max angle I expect since I want the number of gaps between slices.  This +1 was to make the 'increment by 2' thing come out right.

yijiang1:
Nice work! The UI looks good to me. A couple of things:
1. It should be disabled for non-tilt series dataset (just like alignment and reconstruction functions). 
2. The UI doesn't work if Start Tilt > End Tilt.This should be allowed.
For example, say 
Start Tilt = 2, Start Angle = -2.
End Tilt = 0, End Angle = 2.
The result should be 
screen shot 2015-10-08 at 5 46 53 pm

3. The UI should pop up when users click "Mark as Tilt Series" too.

mathturtle:
I am working on 1.
For 2: You want that to work so that you can input the slice range backwards?  It isn't hard, I'm just surprised since other places where you select a slice range you didn't want that (the delete slices transform for example).

Getting it to pop up on Mark as Tilt series is last after I get it working.  I designed it so that it will be easy to do that.

ElliotPadgett:
Some people take data with the angles in descending order and might want to
stick with that in the software, so it would be good if imputing the range
backwards gave descending order.

mathturtle:
I think I have it working for both cases (the input range backward or forward).

mathturtle:
And it should show up when marking a dataset as a tilt series for the first time now.

Hovden:
Works well. Tested OSX. +1

mathturtle:
@cryos?

PR #231, opened 2015-10-14
Add a python transform to clear a part of the volume

mathturtle:
This transform replaces a volume of interest with zeros.  This is my solution to #199 .  The code for the non-modal dialog probably isn't ideal for non-programmers to copy and reuse.

### DIFF
--- a/tomviz/AddPythonTransformReaction.cxx
+++ b/tomviz/AddPythonTransformReaction.cxx
@@ -19,6 +19,7 @@
 #include "DataSource.h"
 #include "OperatorPython.h"
 #include "pqCoreUtilities.h"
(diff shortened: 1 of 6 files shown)


yijiang1:
Works well. Tested OSX.
One minor thing: the UI still say "**Crop** data start/end".

mathturtle:
Good catch @yijiang1.  I've put up a new version that changes it to 'Volume Selected start/end' for both crop and this operation.

mathturtle:
Rebased to avoid conflicts.

Hovden:
Merge?

Hovden:
Merging so it can be used by colleagues (non-programmers).

PR #232, opened 2015-10-15
Added weighted back projection reconstruction method

yijiang1:
Added weighted back projection reconstruction method. 
This is the most traditional and popular method.
Users can specify the reconstruction size and different weighting functions and interpolation methods.


yijiang1:
@Hovden

Issue #170, opened 2015-08-26
Improve Python Development Code Structure


Hovden:
Python development is now ramping up. It is worth thinking about the best way to organize the tomviz code for development and addition of many new python routines and menu items.

mathturtle:
@cryos Essentially we want to install the python files instead of compiling them in right?

cryos:
Pull request #272 makes a few tweaks to the Python files - switch to symlinks on OSes that support them for the Python files in the build trees, and make the Python files show up in IDEs.

I have a simple JSON sample that would be read in to construct the menu entries below, and a branch where some of this is being read in.

  {
    "name": "Tilt Series",
    "description": "Do a tilt series",
    "category": "Tomography",
    "priority": 5
  }

How are you determining the ordering for the menu entries right now - they look somewhat arbitrary. I thought each top level menu could simply fall into a category, but what about separators, absolute ordering? We could insert a separator if the entry passes a 100 line, use the priority number, and alphanumeric sorting as a secondary.

This pushes towards having drag and drop Python files, with separate JSON files (same name, .json and .py file endings).

cryos:
I think we addressed the major points I was concerned about here.

PR #233, opened 2015-10-15
Read in python scripts from files rather than compiling them in

mathturtle:
This is a start on #170, based on a discussion with @cryos a few weeks ago.  This causes the python scripts for the python transforms and sample data to be installed instead of compiled in.  They are installed in X/share since they are not really scripts (they have to have code substituted in to be useful).  If X/share is the wrong place to put them it shouldn't be hard to change.

### DIFF
--- a/tomviz/CMakeLists.txt
+++ b/tomviz/CMakeLists.txt
@@ -146,21 +146,10 @@ set(python_files
   RandomParticles.py
   TiltSeries.py 
   )
-unset(python_h_files)
 foreach(file ${python_files})
   set(file python/${file})
-  get_filename_component(file_we ${file} NAME_WE)
-  set(src  ${CMAKE_CURRENT_SOURCE_DIR}/${file})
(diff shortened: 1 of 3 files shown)


mathturtle:
@cryos

cryos:
LGTM

yijiang1:
I can't run any of my python script now and always get this error message:
raise IOError("Couldn't read script file")

mathturtle:
Which OS?  OSX?

Is this in the binaries or in a local build?  I tested all the superbuilds when I first made the branch to ensure that binaries would work.  Has something broken since then?

yijiang1:
@mathturtle It's in my local build and OSX.

mathturtle:
I figured it out.  Local builds on OSX are different from other OSes...  I'll have a patch up in a minute.

mathturtle:
#245 should handle this.  Try it and feel free to merge if it succeeds.

Issue #229, opened 2015-10-08
Need UI to set tilt angles


yijiang1:
Tomviz need an UI to set angles of a tilt series. This is essential for reconstruction.
The UI allows users to specify the range of tilts for which they want to assign angles. The user also needs to specify the first and last angle. The function assumes equal-angle increment. 
Let N = total number of tilts. The default start and end tilt should be 0 and N-1. The UI always assign 0 degree to the middle tilt (floor(N/2)+1) and have different default angle-increment depending on the number of tilts: 
If N>=120, default angle-increment = 1
If 80<=N<120, default angle-increment = 1.5
If 60<=N<80, default angle-increment = 2
If N<60, default angle-increment = 3

Here is a mock-up of the UI:
![uisettiltangles](https://cloud.githubusercontent.com/assets/13088588/10371400/35812b72-6db1-11e5-99cc-eef7166d68dc.png)
Description:
A tomographic “tilt series” is a set of projection images taken while rotating (“tilting”) the specimen. Setting the correct angles is needed for accurate reconstruction. 
Set a linearly spaced range of angles by specifying the start, end tilt and start, end angle. Note, tilt angles can also be set in “Data Properties” panel or within Python.

The UI should be called when users click "Mark as Tilt Series" and "Set Tilt Angles" in Tomography menu.
![menusettiltangles](https://cloud.githubusercontent.com/assets/13088588/10371537/e38ead66-6db1-11e5-83cc-ce060a8ccc81.png)
It'd be great if we can put similar UI in "Data Property".
![uisettiltangles2](https://cloud.githubusercontent.com/assets/13088588/10371738/c6cb6196-6db2-11e5-8d46-7eff6d8ec8da.png)

mathturtle:
Did that branch take care of all of this?  Or is this last bit about the tilt angles table still left.  What exactly do you want for that?

yijiang1:
I think @Hovden  wants something like this
![uisettiltangles3](https://cloud.githubusercontent.com/assets/13088588/10525355/46847b4e-7352-11e5-99a0-e33d4d0ab02f.png)

mathturtle:
Hmmm. You would at least want some sort of apply button there too.  And keeping the values in those text fields synced with the values in the table would be tricky.  Much easier would be a button that throws up the other UI (equivalent to clicking the menu item).

yijiang1:
[...]

PR #235, opened 2015-10-15
DataPropertiesPanel: Add button to set range of tilt angles

mathturtle:
@yijiang1 How is this for what we discussed on #229?


yijiang1:
LGTM. Test on OSX.

PR #236, opened 2015-10-21
Added manual background subtraction function for tilt series

yijiang1:
Added a function that subtracts background of a tilt series.
The background for each tilt image is the average value in a window specified by the user.
Note:
The function does NOT impose non-negativity constraint. It may results in negative pixels.


Hovden:
+1

Issue #101, opened 2015-06-27
Manual axis shift and rotation finding tool (Basic Framework)


ElliotPadgett:
tomviz should have a tool to help users manually find the rotation and shift of the tilt axis in a tilt series for fine adjustments and when automatic routines do not work.

This could work by showing the user back-projection reconstruction slices from select parts of the tilt series, and the letting them position the axis to minimize artifacts in the reconstruction visually.

yijiang1:
Here is a basic mock-up of the UI. I'll be writing a reconstruction function in C++ so the UI can use it.
![axisalignui](https://cloud.githubusercontent.com/assets/13088588/10553948/d8a71290-742c-11e5-806a-e56ee415c1fa.png)

Hovden:
For this issue, lets just get the basic GUI framework in place.

mathturtle:
The UI shouldn't be too hard to throw together.  Connecting up all the actions and the reconstruction will be the tricky part.

So if I have this right:
The large image in the upper left is an x-y slice of the current data with the z-value controlled by the 'Projection No.' text field.
The yellow line shows the 'center'.  So is this just a static line I can hard code as halfway up the scene or does it move?
The three red dashed lines positions are controlled by the 'slice No.' text fields.  So these text fields are values along the horizontal (X?) axis?
Rotation axis position is a vertical (Y?) offset to the rotation axis (which presumably is a parameter to your reconstruction).
Rotation angle is a rotation about the Z axis applied before reconstruction?
Are rotation angle and rotation axis different for each slice?
What information do you need to generate the reconstruction slices?
What does the create aligned data button do?  Just generate a new dataset that is a copy with this applied similar to the translation align already present in tomviz?

Hovden:
Pretty close, which is good given the minimal set of instructions.

Yes. Large image is x-y slice with z-value controlled by Proj. Number.
No. Yellow line will move up and down by the axis position (in pixels).
Yes. Three dashed lines are controlled by the slice number (in pixels).
Yes. Rotation axis is vertical offset and moves yellow line.
No. Rotation angle will change the angle of the yellow line.
We need the axis position and angle to reconstruct. The gui will run mini reconstructions of the three slices (dashed red lines).
Create aligned data will call a python transform that uses axis position and angle to shift and rotate the data.

Every time the yellow line is adjusted (i.e. rotation axis and angle) we will refresh the three reconstructions using C++ reconstruction code.

[...]

PR #238, opened 2015-10-22
Manual align ui

mathturtle:
This is work in progress toward #101.  Putting it up for feedback on the current setup.

### DIFF
--- a/tomviz/RotateAlignWidget.cxx
+++ b/tomviz/RotateAlignWidget.cxx
@@ -18,6 +18,7 @@
 
 #include "ActiveObjects.h"
 #include "DataSource.h"
+#include "TomographyReconstruction.h"
 
 #include "QVTKWidget.h"
 #include "vtkCamera.h"
@@ -118,12 +119,12 @@ class RotateAlignWidget::RAWInternal
 public:
   Ui::RotateAlignWidget Ui;
   vtkNew<vtkImageSlice> mainSlice;
+  vtkNew<vtkImageData> reconImage[3];
   vtkNew<vtkImageSlice> reconSlice[3];
   vtkNew<vtkImageSliceMapper> mainSliceMapper;
   vtkNew<vtkImageSliceMapper> reconSliceMapper[3];
   vtkNew<vtkRenderer> mainRenderer;
   vtkNew<vtkRenderer> reconRenderer[3];
-  vtkNew<vtkImageData> reconImage;
   QPointer<DataSource> Source;
   vtkNew<vtkLineSource> rotationAxis;
   vtkNew<vtkActor> axisActor;
@@ -207,6 +208,41 @@ public:
     this->Ui.sliceView->update();
   }
 
+  void updateReconSlice(int i)
+  {
+    vtkTrivialProducer *t =
+        vtkTrivialProducer::SafeDownCast(this->Source->producer()->GetClientSideObject());
+    if (!t)
+    {
+      return;
+    }
+    vtkImageData *imageData = vtkImageData::SafeDownCast(t->GetOutputDataObject(0));
+    if (imageData)
+    {
+      int extent[6];
+      imageData->GetExtent(extent);
+      int dims[3] = { extent[1] - extent[0] + 1,
+                      extent[3] - extent[2] + 1,
+                      extent[5] - extent[4] + 1 };
+      QSpinBox *spinBoxes[3] = { this->Ui.spinBox_1, this->Ui.spinBox_2, this->Ui.spinBox_3 };
+      int sliceNum = spinBoxes[i]->value();
+      std::vector<float> sinogram(dims[1] * dims[2]);
+      TomographyReconstruction::tiltSeriesToSinogram(imageData, sliceNum, &sinogram[0]);
+      this->reconImage[i]->SetExtent(0, dims[1] - 1, 0, dims[1] - 1, 0, 0);
+      this->reconImage[i]->AllocateScalars(VTK_FLOAT, 1);
+      vtkDataArray *reconArray = this->reconImage[i]->GetPointData()->GetScalars();
+      float *reconPtr = static_cast<float*>(reconArray->GetVoidPointer(0));
+
+      vtkDataArray *tiltAnglesArray = imageData->GetFieldData()->GetArray("tilt_angles");
+      double *tiltAngles = static_cast<double*>(tiltAnglesArray->GetVoidPointer(0));
+
+      TomographyReconstruction::unweightedBackProjection2(&sinogram[0], tiltAngles, reconPtr, dims[2], dims[1]);
+      this->reconSliceMapper[i]->SetInputData(this->reconImage[i].GetPointer());
+      this->reconSliceMapper[i]->SetSliceNumber(0);
+      this->reconSliceMapper[i]->Update();
+    }
+  }
+
   void updateSliceLines()
   {
     vtkTrivialProducer *t =
@@ -220,12 +256,15 @@ public:
     {
       double bounds[6];
       imageData->GetBounds(bounds);
+      int extent[6];
+      imageData->GetExtent(extent);
+      double maxSlices = extent[1] - extent[0] + 1;
       QSpinBox *spinBoxes[3] = { this->Ui.spinBox_1, this->Ui.spinBox_2, this->Ui.spinBox_3 };
       for (int i = 0; i < 3; ++i)
       {
         double p1[3], p2[3];
-        p1[0] = bounds[0] + (bounds[1] - bounds[0]) * (spinBoxes[i]->value() / (bounds[1] - bounds[0]));
-        p2[0] = bounds[0] + (bounds[1] - bounds[0]) * (spinBoxes[i]->value() / (bounds[1] - bounds[0]));
+        p1[0] = bounds[0] + (bounds[1] - bounds[0]) * (spinBoxes[i]->value() / maxSlices);
+        p2[0] = bounds[0] + (bounds[1] - bounds[0]) * (spinBoxes[i]->value() / maxSlices);
         p1[1] = bounds[2];
         p2[1] = bounds[3];
         p1[2] = bounds[5];
@@ -259,10 +298,6 @@ RotateAlignWidget::RotateAlignWidget(DataSource *source, QWidget *p)
       this->Internals->reconRenderer[1].Get());
   this->Internals->Ui.sliceView_3->GetRenderWindow()->AddRenderer(
       this->Internals->reconRenderer[2].Get());
-
-  this->Internals->reconSliceMapper[0]->SetOrientation(0);
-  this->Internals->reconSliceMapper[1]->SetOrientation(0);
-  this->Internals->reconSliceMapper[2]->SetOrientation(0);
 
   vtkNew<vtkInteractorStyleRubberBand2D> interatorStyle;
   interatorStyle->SetRenderOnMouseMove(true);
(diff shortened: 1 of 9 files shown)


tomviz/RotateAlignWidget.cxx
```diff
@@ -0,0 +1,319 @@
...
+    vtkImageData *imageData = vtkImageData::SafeDownCast(t->GetOutputDataObject(0));
+    if (imageData)
+    {
+      int extent[6];
+      imageData->GetExtent(extent);
+      double spacing[3];
+      imageData->GetSpacing(spacing);
+      int projections = extent[5] - extent[4] + 1;
+      this->Ui.projection->setMaximum(projections - 1);
+      this->Ui.rotationAxis->setRange(-projections * spacing[1] / 2.0, projections * spacing[1] / 2.0);
```
yijiang1: The range of the rotationAxis should be  +/-(extent[3] - extent[2])/2 instead of +/-(extent[5] - extent[4])/2.
We are shifting the images in the y-direction.

yijiang1:
Looks good. A couple of things:
1. Can you add three vertical lines indicating the slices (see red lines in #101 ). They should move horizontally whenever the user changes the slice No.
2. It's better to add labels indicating the dimension of the tilt image.
3. In the reconstruction viewing window, can you show three y-z slices corresponding to the their slice No.? The image in y-z plane (often referred as a "sinogram") will be the input of the reconstruction function.
4. The rotation center of the yellow line is not at (x=0,y=0)  (the lower-left corner). It should be the center of the yellow line. See figure below:
axisalignui2

tomviz/RotateAlignWidget.cxx
```diff
@@ -0,0 +1,319 @@
...
+    if (!tform)
+    {
+      vtkNew<vtkTransform> t;
+      t->PreMultiply();
+      tform = t.Get();
+      this->axisActor->SetUserTransform(t.Get());
+    }
+    tform->Identity();
+    tform->Translate(0, this->Ui.rotationAxis->value(), 0);
+    tform->RotateZ(this->Ui.rotationAngle->value());
```
yijiang1: Need to specify proper rotation center.

tomviz/RotateAlignWidget.cxx
mathturtle: Good catch!

tomviz/RotateAlignWidget.cxx
mathturtle: This is what I get for working with a volume that is a cube centered at the origin... I don't see the interesting cases.

mathturtle:
Ok:
1. Done.  I can't figure out how to make the lines dashed.  It looks like that is unsupported in OpenGL2 at the moment.
2. I'll see what I can do without breaking the rest of the UI.  These labels (at least the way you have them positioned) will be tricky.
3. Are you referring to the three views at the bottom?  I thought they were supposed to be slices from the output of a reconstruction algorithm so I left them showing a blank image for now.  Are they just y-z slices of the original image?  If so, that isn't hard...
4. Done.

I also fixed some display issues for large images where the top and bottom slices were getting chopped off by the camera's clipping range.

Known issues:
* The 'Create aligned data' button does nothing at the moment.
* I'm still trying to figure out what to display in the three views at the bottom.
* Red lines aren't dashed.

ElliotPadgett:
Ultimately I think it would be best if the user could toggle between y-z
slices (sinograms) and reconstructions of those slices.  For now just the
sinograms should be fine as as place holders.

mathturtle:
New version:
I got as close as I could with the size labels to what you had in your mock-up above.  Let me know if it is ok.  Anything closer will probably require hacky Qt code.

Discovered new issue:
* If the y dimension of the data is larger than the x dimension, the data gets clipped at the top and bottom so that the x dimension fits to the screen size and the extremes in y are offscreen.  I'm looking into this now.

yijiang1:
> 1. Done. I can't figure out how to make the lines dashed. It looks like that is unsupported in OpenGL2 at the moment.

The line is dashed on my computer like this
screen shot 2015-10-22 at 3 36 58 pm

> The 'Create aligned data' button does nothing at the moment.

Don't worry about it for now. We'll implement it after we have everything else.

tomviz/RotateAlignWidget.cxx
```diff
@@ -0,0 +1,401 @@
...
+    sMapper->SetInputConnection(this->Internals->reconSliceLine[i]->GetOutputPort());
+    this->Internals->reconSliceLineActor[i]->SetMapper(sMapper.Get());
+    this->Internals->reconSliceLineActor[i]->GetProperty()->SetColor(1, 0, 0);
+    this->Internals->reconSliceLineActor[i]->GetProperty()->SetLineWidth(2.0);
+    this->Internals->reconSliceLineActor[i]->GetProperty()->SetLineStipplePattern(0xFF00);
+    this->Internals->mainRenderer->AddActor(this->Internals->reconSliceLineActor[i].Get());
+  }
+  this->Internals->Ui.spinBox_1->setValue(25);
+  this->Internals->Ui.spinBox_2->setValue(50);
+  this->Internals->Ui.spinBox_3->setValue(75);
```
yijiang1: The range of the three spinboxes above should be from 0 to extent[1]-extent[0] (number of y-z slices - 1). Right now the range is 0-99 (by default).

mathturtle:
> The line is dashed on my computer like this
> screen shot 2015-10-22 at 3 36 58 pm

When you configured ParaView in CMake, did you set VTK_RENDERING_BACKEND to OpenGL2? (You can tell by looking for a line with that looks like this [1] in the CMakeCache.txt file in the ParaView build folder)  If not, then you are still on the old backend.  That's not necessarily bad, but you are losing out on a huge volume rendering improvement among various other fixes.  The tomviz binaries use OpenGL2 right now.

[1]: `VTK_RENDERING_BACKEND:STRING=X `
X should be OpenGL or OpenGL2

tomviz/RotateAlignWidget.cxx
mathturtle: That's because I was assuming slices from a small (100^3) reconstruction volume was going to be displayed there.  I'll change it when I add the sinograms.

yijiang1:
@mathturtle I see. My VTK_RENDERING_BACKEND is still set to OpenGL.

mathturtle:
New version.  I found the last bug in my 'show sinograms' code just now.  I'll work on the stuff from the meeting now.  If you want to try this out in the meantime...

yijiang1:
@mathturtle The new version looks good. Thanks!

mathturtle:
@yijiang1 I've put up a new version that uses your reconstruction code.  Can you look at it/try it out and give me feedback please?

tomviz/RotateAlignWidget.cxx
```diff
@@ -0,0 +1,503 @@
...
+      vtkTrivialProducer::SafeDownCast(source->producer()->GetClientSideObject());
+  vtkImageData *imageData = vtkImageData::SafeDownCast(t->GetOutputDataObject(0));
+  DataSource* output = source->clone(true,true);
+  QString name = output->producer()->GetAnnotation("tomviz.Label");
+  name = "Rotation_Aligned_" + name;
+  output->producer()->SetAnnotation("tomviz.Label", name.toAscii().data());
+  t = vtkTrivialProducer::SafeDownCast(output->producer()->GetClientSideObject());
+  vtkImageData *recon = vtkImageData::SafeDownCast(t->GetOutputDataObject(0));
+
+  TomographyReconstruction::weightedBackProjection3(imageData, recon);
```
yijiang1: No need to perform a 3D reconstruction. This UI is only for generating aligned tilt series. Users may want to use other reconstruction methods.

yijiang1:
@mathturtle The UI works well. :thumbsup:
One more thing:
Can you make the three reconstruction windows square instead of rectangular? 
For now you can assume that the size the each 2D reconstruction is always 256 by 256. I'm going to modify the "tiltSeriesToSinogram" function so that its output has size 256 regardless of the tilt series.

yijiang1:
Can you also remove the top size label?
screen shot 2015-11-09 at 2 44 16 pm
Thus we'll have more space for recon windows.

tomviz/RotateAlignWidget.cxx
mathturtle: Do we have a function to generate an aligned tilt series?  For now I'll make it do nothing again.

yijiang1: Nope. I'll add that function later.

mathturtle:
I've moved the size label a bit... you'll have to let me know if it is better.

I've made all the reconstruction windows bigger (256 x 256 by default).  I would recommend making the size of the reconstruction a parameter of the function rather than hard-coding it if that is something easy to do.

yijiang1:
> I would recommend making the size of the reconstruction a parameter of the function rather than hard-coding it if that is something easy to do.

I will work on this in the future. For the rotation alignment UI, we want to restrict the reconstruction size to 256 to save time.

yijiang1:
I talked to @Hovden and he has some additional suggestions to the UI:
1. Change the title to "Rotation Axis Alignment".
2. Make the starting projection number to total number of tilts(zDim)/2 instead of 0.
3. Change color map to gray scale (if this is easy to do) for all images.
4. Move the tilt series window to the left. Also add a label for description. The UI can look like this:
axisalignui3
We'll add detailed instructions later. For now, just create a label as placeholder. Thanks!

ercius:
I find it easier to do this alignment if you show the sum of all
projections rather than a single projection. The average shows a blur that
is perpendicular to the tilt axis.

Hovden:
Good idea, I often use the sum of projections as well. @yijiang1 @mathturtle could you implement a checkbox to the right of "projection no" so that when it is checked the sum of all projections is displayed instead of one specific particular from the tiltseries?

Great to see this Axis Alignment tool is come together.

yijiang1:
> could you implement a checkbox to the right of "projection no" so that when it is checked the sum of all projections is displayed instead of one specific particular from the tiltseries?

I think we should implement this feature in future releases. We are running out of time to finish the basic UI for v0.7

Hovden:
OK. I've made it request #252 so we can get to it in a future release.

mathturtle:
> 1. Change the title to "Rotation Axis Alignment".
> 2. Make the starting projection number to total number of tilts(zDim)/2 instead of 0.
> 3. Change color map to gray scale (if this is easy to do) for all images.
> 4. Move the tilt series window to the left. Also add a label for description. The UI can look like this:

I've done all of these except the color map so far.  The color map will be a separate branch that sets the tomviz-wide default color map to grayscale.  This UI doesn't do anything special, it just uses whatever colormap is associated with that dataset to display the data, so if we change the tomviz default it will change.

yijiang1:
The UI looks pretty good. Let's merge it? @Hovden

Hovden:
It is ready to merge for testing and feedback. Someone familiar with the code should merge.

mathturtle:
Ok.  Keep in mind that while all the pieces work, the final 'Create aligned data' button doesn't do anything yet since I don't have a create aligned data function.  But I'm fine with merging the branch as it stands.  Let @cryos look over the code for little things I missed though.

Hovden:
The 'create align button' will create a shift and rotate python data transform in the pipeline. The shift will be a global shift (applied first) and then a global rotation (applied second).

I'm happy to merge with or without this button working.

yijiang1:
> the final 'Create aligned data' button doesn't do anything yet since I don't have a create aligned data function

I'm working on this now. I'll add all the missing parts in a separate branch.

Issue #237, opened 2015-10-21
Add axis labels to SelectVolumeWidget


yijiang1:
Can we add x,y,z labels in the SelectVolumeWidget UI like this?
<img width="390" alt="screen shot 2015-10-21 at 6 01 48 pm" src="https://cloud.githubusercontent.com/assets/13088588/10651558/2eb7191a-781e-11e5-8c3b-492d2d43ef30.png">

yijiang1:
Tested on cdash build.

PR #242, opened 2015-10-26
Add axis labels to select volume widget dialog

mathturtle:
@yijiang1 How's this for #237?


yijiang1:
Looks good. Tested OSX.

PR #243, opened 2015-10-27
C++11 use override and nullptr

mathturtle:
This branch updates tomviz's minimum required CMake to the version present on Ubuntu 14.04 and turns on C++11 support in the CMake files so that the project is built with C++11.

Three following commits switch the project to use `override` and `nullptr` and remove `virtual` from functions with `override`.

@cryos


cryos:
Looks great, thanks for getting this working.

PR #245, opened 2015-10-27
OSX: local builds find script files now

mathturtle:
OSX local builds are special in that the executable is 3 levels down in
bin/tomviz.app/Contents/MacOS instead of being directly in bin.  Handle
this.

### DIFF
--- a/tomviz/MainWindow.cxx
+++ b/tomviz/MainWindow.cxx
@@ -382,6 +382,22 @@ QString MainWindow::readScript(const QString &scriptName)
     QByteArray array = file.readAll();
     return QString(array);
   }
+// On OSX the above doesn't work in a build tree.  It is fine
+// for superbuilds, but the following is needed in the build tree
+// since the executable is three levels down in bin/tomviz.app/Contents/MacOS/
+#ifdef __APPLE__
+  else
+  {
+    path = QApplication::applicationDirPath() + "/../../../../share/tomviz/scripts/";
+    path += scriptName + ".py";
+    QFile file2(path);
+    if (file2.open(QIODevice::ReadOnly))
+    {
+      QByteArray array = file2.readAll();
+      return QString(array);
+    }
+  }
+#endif
   qCritical() << "Error: Could not find script file: " << scriptName;
   return "raise IOError(\"Couldn't read script file\")\n";
 }


yijiang1:
I'm getting a new error:
screen shot 2015-10-27 at 5 40 02 pm

mathturtle:
Have you re-run CMake since the change to python files went in?  CMake is copying them to the build tree and that may be the problem.  You shouldn't be getting all those errors at the top: the 'Could not find script file' errors.  The IOError is just a side effect.

yijiang1:
> Have you re-run CMake since the change to python files went in?

@mathturtle Can you be more specific about what should I do to "re-run cmake"?

mathturtle:
> @mathturtle Can you be more specific about what should I do to "re-run cmake"?

The easiest way is to go to the build directory and run `cmake .`, but you may not have the command line tools for CMake installed.  So re-run the cmake gui and click Configure and Generate to force it to re-run the CMake code.  It probably already did this though, but it's better to double check.

yijiang1:
> So re-run the cmake gui and click Configure and Generate to force it to re-run the CMake code.

I did this and still get the " raise IOError("Couldn't read script file") " error.

mathturtle:
Hmmm.  Look in the build folder under `share/tomviz/scripts`.  All the script files *should* be copied there.  Make sure they are there.  Also let me know the path from the build folder to the tomviz executable.  It should be either `bin/tomviz` or `bin/tomviz.app/Contents/MacOS/tomviz`.  If either one of these are wrong, then I'll have to figure out why it is wrong on your machine while working on our mac.

yijiang1:
> Look in the build folder under share/tomviz/scripts. All the script files should be copied there.

Yes. All python scripts are there.


> Also let me know the path from the build folder to the tomviz executable.

My tomvia.app is in  .../tomviz_build/bin/tomviz

mathturtle:
So maybe I wasn't clear enough.  OSX .app files are really folders (the UI hides this from you, but on the terminal you can look around inside them).  What is the path that starts with .../tomviz_build/... and ends in .../tomviz.app/Contents/MacOS/tomviz?  (This is the executable that is run when you open that .app)

yijiang1:
I see. The path is .../tomviz_build/bin/tomviz.app/Contents/MacOS/tomviz 
When I double click the tomviz executable in MacOS folder, I get a bunch of "Could not find script file..." errors in terminal.

mathturtle:
That's what I thought.  Building this branch should fix it unless your Qt is doing something funny with the applicationDirPath.  If building this branch does not fix the error...
(see inline comment)

tomviz/MainWindow.cxx
```diff
@@ -382,6 +382,21 @@ QString MainWindow::readScript(const QString &scriptName)
     QByteArray array = file.readAll();
     return QString(array);
   }
+// On OSX the above doesn't work in a build tree.  It is fine
+// for superbuilds, but the following is needed in the build tree
+// since the executable is three levels down in bin/tomviz.app/Contents/MacOS/
+#if __APPLE__
+  else
+  {
+  path = QApplication::applicationDirPath() + "/../../../../share/tomviz/scripts/";
```
mathturtle: ...add a line here `qDebug() << QApplication::applicationDirPath();` and tell me what it says (you can shorten it to start with .../tomviz_build if you want).

mathturtle: To be clear, this is a debugging print statement using Qt's debugging macros, so when you run the program it should print a path.  If you are building in release you may have to change the first part to `qWarning()` to get it to show anything.

yijiang1: It prints .../tomviz_build/bin/tomviz.app/Contents/MacOS

I can "fix" my problem by changing the directory in line 377 from "/../share/tomviz/scripts/" to "/../../../../share/tomviz/scripts/".
Maybe Tomviz never run the code between line 388 to line 400?

mathturtle:
Try this new version.  I think I messed up my 'if on a Mac' condition.

yijiang1:
It still doesn't work.

mathturtle:
What compiler are you using?

yijiang1:
I think it's the default one in Xcode. Is there a way to check?

cryos:
Shawn is just taking a look at this on a Mac build here, he should hopefully have an updated patch for you shortly - he was able to reproduce the failure locally. Thanks Yi!

mathturtle:
Somehow a line of the change got lost when I moved the patch to my main machine...  Sorry about that.  Try this version.

yijiang1:
Yup it works now!

Issue #239, opened 2015-10-23
Reconstruction operation in C++


cryos:
Need to provide capabilities to perform reconstruction in C++. For the first pass provide a C++ operation that takes in the image stack, converts it to a float, and passes it to the reconstruction routine. The reconstruction routine will return a float volume. Convert vtkImageData to a C array pointer, show layout/give example of going through values in the volume.

cryos:
Let's do the widget in-place reconstruction first, and then look at C++ operations later perhaps.

mathturtle:
@yijiang1 @cryos Is this enough of an example?  I tried to show how to get the raw data pointer, allocate the output and access the data at an index.  We should talk about the exact signature of your reconstruction function soon so that I will know what input it needs.

```c++
#include "vtkImageData.h"

void ReconstructionAlgorithm(vtkImageData *input, vtkImageData *output)
{
  if (input->GetScalarType() != VTK_FLOAT)
  {
    // print error
    return;
  }
  int extents[6];
  input->GetExtent(extents);
  int xDim = extents[1] - extents[0] + 1;
  int yDim = extents[3] - extents[2] + 1;
  int zDim = extents[5] - extents[4] + 1;
  float *data = static_cast<float*>(input->GetScalarPointer());

  // Creating the output volume and getting a pointer to it
  int outputSize[3] = {100, 100, 100};
  output->SetExtent(0, outputSize[0] - 1, 0, outputSize[1] - 1, 0, outputSize[2] - 1);
  output->AllocateScalars(VTK_FLOAT, 1); // 1 is for one component (i.e. not vector)
  float * outPtr = static_cast<float*>(output->GetScalarPointer());

  // Accessing data:
  for (int z = 0; z < zDim; ++z)
  { 
    for (int y = 0; y < yDim; ++y)
    {
      for (int x = 0; x < xDim; ++x)
      {
        outPtr[z * outputSize[0] * outputSize[1] + y * outputSize[0] + x] =
          data[z * yDim * xDim + y * xDim + x];
        // obviously you want to be a bit more careful about out of bounds access than this...
      } 
    } 
  } 
} 
```

yijiang1:
LGTM. Another basic input the reconstruction function need is the tilt angles.

mathturtle:
Right, tilt angles.  You would do the following:
```c++
#include "vtkFieldData.h"
#include "vtkDataArray.h"

.... // skipping into function from other example
[...]

PR #246, opened 2015-10-28
Basic tomography reconstruction in C++

yijiang1:
Added basic framework for C++ tomography reconstruction.
I made two new classes: "TomographyReconstruction" and "ReconstructionReaction".

Class "TomographyReconstruction" holds all the reconstruction functions. 
Right now it has a simple back projection method. Other recon methods will be added later, including 2D reconstruction functions, which will be useful for axis alignment UI (#238 ).
The class also contains code for converting data type to float (#239 ). These functions should be implemented somewhere else.

Class "ReconstructionReaction" connects UI (Tomography menu) to TomographyReconstruction.

### DIFF
--- a/tomviz/TomographyReconstruction.cxx
+++ b/tomviz/TomographyReconstruction.cxx
@@ -37,50 +37,40 @@ TomographyReconstruction::~TomographyReconstruction()
 
 }
 
-  // conversion code
-  template<typename T>
-  vtkSmartPointer<vtkFloatArray> convertToFloatT(T *data, int len)
+// conversion code
+template<typename T>
+vtkSmartPointer<vtkFloatArray> convertToFloatT(T *data, int len)
+{
+  vtkSmartPointer<vtkFloatArray> array = vtkSmartPointer<vtkFloatArray>::New();
+  array->SetNumberOfTuples(len);
+  float *f = static_cast<float*>(array->GetVoidPointer(0));
+  for (int i = 0; i < len; ++i)
   {
-    vtkSmartPointer<vtkFloatArray> array = vtkSmartPointer<vtkFloatArray>::New();
-    array->SetNumberOfTuples(len);
-    float *f = static_cast<float*>(array->GetVoidPointer(0));
-    for (int i = 0; i < len; ++i)
-    {
-      f[i] = (float) data[i];
-    }
-    return array;
+    f[i] = (float) data[i];
   }
+  return array;
+}
   
-  vtkSmartPointer<vtkFloatArray> convertToFloat(vtkImageData* image)
+vtkSmartPointer<vtkFloatArray> convertToFloat(vtkImageData* image)
+{
+  vtkDataArray *scalars = image->GetPointData()->GetScalars();
+  int len = scalars->GetNumberOfTuples();
+  vtkSmartPointer<vtkFloatArray> array;
+  switch(scalars->GetDataType())
   {
-    vtkDataArray *scalars = image->GetPointData()->GetScalars();
-    int len = scalars->GetNumberOfTuples();
-    vtkSmartPointer<vtkFloatArray> array;
-    switch(scalars->GetDataType())
-    {
-        vtkTemplateMacro(array = convertToFloatT(static_cast<VTK_TT*>(scalars->GetVoidPointer(0)), len););
-    }
-    return array;
+    vtkTemplateMacro(array = convertToFloatT(static_cast<VTK_TT*>(scalars->GetVoidPointer(0)), len););
   }
+  return array;
+}
   
 void TomographyReconstruction::reconWBP(vtkImageData *tiltSeries,vtkImageData *recon)
 {
-  //Tomviz currently can't change data type. Enable this in the future.
-  /*
-  if (tiltSeries->GetScalarType() != VTK_FLOAT)
-  {
-    qDebug() << "Wrong datatype :-(";
-    return;
-  }
-  */
-  
   int extents[6];
   tiltSeries->GetExtent(extents);
   int xDim = extents[1] - extents[0] + 1;
   int yDim = extents[3] - extents[2] + 1;
   int zDim = extents[5] - extents[4] + 1;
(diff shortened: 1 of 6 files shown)


tomviz/TomographyReconstruction.cxx
```diff
@@ -0,0 +1,137 @@
...
+          if (t >= -yDim/2 && t <= yDim/2 )	//check if ray is inside projection
+          {
+            int rayIndex = floor((t + yDim/2));
+            if (rayIndex >= 0 && rayIndex <= yDim-2)
+            {
+              //Linear interpolation
+              double Q1 = data[ang*xDim*yDim + rayIndex * xDim + x ];
+              double Q2 = data[ang*xDim*yDim + (rayIndex+1) * xDim + x ];
+              double QDash = Q1 + (t-double(rayIndex-yDim/2))*(Q2-Q1);
+              reconPtr[iz*outputSize[0]*outputSize[1] + iy*outputSize[0] + x] += QDash;
```
yijiang1: The way I'm updating the reconstruction is not very efficient, because x is varied in the outermost loop. Is there a good way to reindex the array?

tomviz/TomographyReconstruction.h
```diff
@@ -0,0 +1,40 @@
...
+
+class TomographyReconstruction
+{
+    
+public:
+  TomographyReconstruction();
+  ~TomographyReconstruction();
+  static void weightedBackProjection3(vtkImageData *tiltSeries,vtkImageData *recon); //3D WBP recon
+  static void unweightedBackProjection2(float *sinogram, double *tiltAngles, float *recon,int numOfTilts, int numOfRays ); //2D WBP recon
+  static void tiltSeriesToSinogram(vtkImageData *tiltSeries, int, float* sinogram); //Extract sinograms from tilt series
```
yijiang1: Function tiltSeriesToSinogram takes a tilt series and integer (x index) and extract a sinogram (y-z plane).
Then you can pass the sinogram pointer, along with other parameters, to "unweightedBackProjection2" and obtain the reconstruction.

@mathturtle Are these two function enough for you to finish the axis alignment UI #238 ?

mathturtle: Some questions about your functions:
* How do I tell it the dimensions of the sinogram?  Or of the reconstruction for that matter?  Or is that something that it should tell me?
* What is the parameter numOfRays?
* If I am supposed to display the sinogram then having it as a vtkImageData rather than just a float* would be helpful.  I don't mind passing you the raw pointer, but we need to figure out the dimension issue.

yijiang1: @mathturtle 
Let's say the dimension of a tilt series is (xDim,yDim,zDim). Because a sinogram is just a y-z plane of the tilt series, the pointer *sinogram points to a 1D (float) array of size yDim*zDim. 
Right now, the size of 3D reconstruction is always set to (xDim, yDim, yDim). For the 2D recon function "unweightedBackProjection2", the *recon points to a 1D array of size yDim*yDim. 
The parameter "numOfRays" is just yDim. Together with "numOfTilts", which is zDim, they specify the size of sinogram.

Because we have 2D reconstruction function now, there is no need to display sinograms in the UI. We can just display the 2D reconstructions.

mathturtle: Can you add comments to the code to explain that?  Those are good things to know.  I'll try to use this with my branch and let you know how it goes.

tomviz/TomographyReconstruction.h
```diff
@@ -0,0 +1,40 @@
...
+namespace tomviz
+{
+class DataSource;
+
+class TomographyReconstruction
+{
+    
+public:
+  TomographyReconstruction();
+  ~TomographyReconstruction();
```
mathturtle: If all your functions are public and static then there is no need to create a class at all in C++.  You can just put all the functions in the tomviz namespace (they don't need to be static this way) or if you want to access them with `tomviz::TomographyReconstruction::X`, then you can do a nested `namespace TomographyReconstruction` instead of a class declaration.

tomviz/TomographyReconstruction.cxx
```diff
@@ -0,0 +1,169 @@
...
+    TomographyReconstruction::tiltSeriesToSinogram(tiltSeries, s, sinogram);
+    //2D back projection
+    TomographyReconstruction::unweightedBackProjection2(sinogram, tiltAngles, recon2d, zDim, yDim);
+    //Put recon into
+    for (int iy = 0; iy < outputSize[1]; ++iy) //loop through all pixels in reconstructed image (y-z plane)
+      for (int iz = 0; iz < outputSize[2]; ++iz)
+      {
+        reconPtr[iz*outputSize[0]*outputSize[1] + iy*outputSize[0] + s] = recon2d[iy*outputSize[0] + iz];
+      }
+  }
```
yijiang1: @mathturtle Line 87 to 102 shows how I currently do the 3D reconstruction.
First, I allocate two pointers as placeholders for sinograms and 2D reconstructions (line 88 and 89).
Next, I loop through each slice (y-z plane) in the tilt series (line 90). For each slice, I extract the corresponding sinogram (line 93) and use it to obtain a 2D reconstruction (line 95). Lastly, I copy the 2D reconstruction to the 3D volume (reconPtr). 

I think we can implement the alignment UI in similar fashion:
Every time user specify the a slice No., 
1. Extract the corresponding sinogram (using "tiltSeriesToSinogram")
2. Reconstruct the new sinogram (using "unweightedBackProjection2")
3. Update recon window in the UI.

mathturtle:
One more question: where do I input the axis position and angle from the recon UI?

yijiang1:
@mathturtle Good question! My plan is to make another function that's similar to "tiltSeriesToSinogram" but take two additional inputs (axis position and angle). The new function will interpolate a sinogram based on three parameters (slice number, axis position and axis angle) instead of taking it directly from the tilt series.
For now, let's just focus on displaying reconstructions in the alignment UI. I'll add the new function in the future.

Issue #248, opened 2015-11-04
Change default color map


ercius:
The cool-to-warm color map is the default color map, but not very useful for volumetric tomography data.I suggest changing the default color map to gray scale.

Hovden:
Also, issue #44 would make changing the colormap more accessible.

cryos:
Thanks for the reminder, we should make this change before the next release (and make the color map easier to change).

Hovden:
I know gray scale has been advocated. But there are great colormaps that have the same intensity transfer profiles as gray but provide more dynamic range to the human eye (even colorblind people).

Here are the new colormaps used in matplotlib: http://matplotlib.org/1.5.0/users/whats_new.html#new-colormaps

They've been designed by color theorists, a great discussion is given here: https://bids.github.io/colormap/

There are four colormaps released under a CC0 license. I personally like "magma" and "viridis".

PR #254, opened 2015-11-11
Change application-wide default color map to grayscale

mathturtle:
@Hovden @cryos @yijiang1 This changes the default color map as we discussed last week.


yijiang1:
LGTM.

cryos:
Looks good, we want to make sure users can change the default and save that if it is not currently possible too.

mathturtle:
> we want to make sure users can change the default and save that if it is not currently possible too.

It is not currently possible (which is why I didn't feel bad about hardcoding the new default for now).

cryos:
Great, thanks for confirming.

mathturtle:
So can we close #248?

Issue #253, opened 2015-11-10
Autosave crashed when restoring (Mac OSX)


Hovden:
Autosave crashes when restoring the saved state.
1) Open reconstruction sample data.
2) Click volume render.
3) Wait 5 minutes for autosave.
4) Force Quit to create crash.
5) Open tomviz. Restore state. Then software crashes.

mathturtle:
The autosaves were going to the system temporary directory, which on OSX is a nested path in /private/var/... and is symlinked to /var/...  The symlink was causing the state file relative paths to fail when reloading the data files which meant that nothing was loaded and the output of the reader was an empty image.  This was causing the histogram creation code to crash since it didn't expect an empty image.

Hovden:
Now works when restoring visualization pipeline on OSX. Now testing with basic data transforms.

Hovden:
Data transforms also working. Going to close this issue and the milestone issue. If something new pops up, we can start a new bug report.

PR #255, opened 2015-11-11
Fix autosave paths

mathturtle:
@Hovden This should fix your autosave crash (#253).


cryos:
LGTM - I will let @Hovden verify.

Hovden:
Having compiler issues after upgrading to El Capitan. Merging; then testing.

PR #256, opened 2015-11-12
Reorganize C++ tomography functions

yijiang1:
Reorganize C++ tomography functions:
Function "tiltSeriesToSinogram" in file "TomographyReconstruction" is renamed to "getSinogram" and moved to a new file called "TomographyTiltSeries". 
In the future, all non-reconstruction operations, such as extracting sinogram, generating tilt series from a volume, and summing over projections will be implemented in "TomographyTiltSeries".
File "TomographyReconstruction" will hold reconstruction-related functions only.

Added an alternative "getSinogram" function that takes additional parameters (number of rays per projection and rotation axis position). It generates a rescaled and shifted sinogram by linear interpolation. This is useful for the manual rotation axis alignment UI.

The manual axis alignment UI now displays reconstructions from shifted tilt series. The reconstruction size is fixed to 256 by 256. The update is not very fast for realistic image size, so I approximate the in-plane as additional shifts in y-direction. I'll address the speed issue in separate branches.

### DIFF
--- a/tomviz/CMakeLists.txt
+++ b/tomviz/CMakeLists.txt
@@ -116,6 +116,8 @@ set(SOURCES
   ToggleDataTypeReaction.cxx
   TomographyReconstruction.h
   TomographyReconstruction.cxx
+  TomographyTiltSeries.h
+  TomographyTiltSeries.cxx
   Utilities.cxx
   Utilities.h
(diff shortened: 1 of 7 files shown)


yijiang1:
Merging this so I can work on reconstruction and axis alignment separately.

PR #258, opened 2015-11-12
Finish "Create aligned data" button in manual axis alignment UI

yijiang1:
Finishing up the "Create aligned data" button in manual axis alignment UI.
As required by @Hovden , The 'create align button' now "create a shift and rotate python data transform in the pipeline. The shift will be a global shift (applied first) and then a global rotation (applied second)."

@mathturtle  I'm not sure what's the optimal way to assess the python scripts, but my code works.

Also, the existing "shift uniformly" python function is not suitable for alignment, because it assumes periodic boundary condition and only takes integer inputs. I'll write a new one later.

### DIFF
--- a/tomviz/RotateAlignWidget.cxx
+++ b/tomviz/RotateAlignWidget.cxx
@@ -522,8 +522,8 @@ void RotateAlignWidget::onFinalReconButtonPressed()
                        QString("SHIFT = [%1, %2, %3]").arg(0)
                        .arg(-this->Internals->Ui.rotationAxis->value()).arg(0));
   
-  QString scriptLabel = "Shift Uniformly";
-  QString scriptSource = readInPythonScript("Shift_Stack_Uniformly"); //TODO:Rewrite python script
+  QString scriptLabel = "Shift";
+  QString scriptSource = readInPythonScript("Shift3D"); //TODO:Rewrite python script
   AddPythonTransformReaction::addPythonOperator(this->Internals->Source, scriptLabel, scriptSource, substitutions);
   
   //Apply in-plane rotation
--- a/tomviz/RotateAlignWidget.cxx
+++ b/tomviz/RotateAlignWidget.cxx
@@ -523,6 +523,7 @@ void RotateAlignWidget::onFinalReconButtonPressed()
   QString scriptLabel = "Shift";
   QString scriptSource = readInPythonScript("Shift3D");
   AddPythonTransformReaction::addPythonOperator(this->Internals->Source, scriptLabel, scriptSource, substitutions);
+  substitutions.clear();
   
   //Apply in-plane rotation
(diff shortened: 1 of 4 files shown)


mathturtle:
I don't really think we want to call python from a C++ operation like this.  Ideally we should write C++ code for the rotate and shift that you want to apply.  I think vtkImageReslice is able to do what you want, but it is complicated to use...

Hovden:
Right now placing it in the python pipeline is good. It provides a transparent workflow. We can revisit this as a c++ interface later. (similarly, we need to address the manual translation alignment c++ approach since cloning is not the best approach).

mathturtle:
@yijiang1 Let me add some helper function in that you can use.  We don't really want to duplicate all the logic for python operators here...

yijiang1:
@mathturtle Sure. Thanks

yijiang1:
New changes:
1. Use the new helper functions introduced in #261 to read python scripts.
2. Added a new python function "Shift3D" to shift aligned data. It does not warp the image and can do sub-pixel shifts.
3. Rotation angle spinbox's step size is set to 0.5.

tomviz/RotateAlignWidget.cxx
```diff
@@ -509,9 +510,29 @@ void RotateAlignWidget::onFinalReconButtonPressed()
...
+  //Apply shift (in y-direction) first
+  QMap<QString, QString> substitutions;
+  substitutions.insert("###SHIFT###",
+                       QString("SHIFT = [%1, %2, %3]").arg(0)
+                       .arg(-this->Internals->Ui.rotationAxis->value()).arg(0));
+
+  QString scriptLabel = "Shift";
+  QString scriptSource = readInPythonScript("Shift3D");
+  AddPythonTransformReaction::addPythonOperator(this->Internals->Source, scriptLabel, scriptSource, substitutions);
+
```
mathturtle: You may want to add
```c++
substitutions.clear();
```
call here.  It doesn't really matter unless `###SHIFT###` appears in your second script though.

tomviz/RotateAlignWidget.h
```diff
@@ -49,6 +49,7 @@ protected slots:
 
 private:
   Q_DISABLE_COPY(RotateAlignWidget)
+  QString readScript(const QString &scriptName);
```
mathturtle: You can take this out now that the helper functions are in.

tomviz/RotateAlignWidget.cxx
yijiang1: Yeah it doesn't matter. But it's good to know the clear function.

tomviz/RotateAlignWidget.h
yijiang1: Good catch!

Issue #260, opened 2015-11-13
Axis Alignment Testing and Polish


Hovden:
The axis alignment tool looks good. It needs testing and validation plus a little polish on the UX.

Hovden:
Instructions: Adjust the "Shift Rotation Axis" and the "Tilt Rotation Axis" such that the three reconstructions on the bottom look accurate. The yellow line shows the current axis of rotation. The three vertical red lines show where the displayed reconstructions are taken from. It is often useful to reconstruct different slices of the tomogram by changing the corresponding "Slice No.". When complete, click "Transform Data to Axis of Rotation" button

Widget Title: Determine the Axis of Rotation

@yijiang1

yijiang1:
@mathturtle Could you add the instructions above? I'm not sure how to edit UI in Qt.

yijiang1:
@Hovden The basic UI is finished now.

Hovden:
Great. It appears to be working properly.

PR #262, opened 2015-11-13
Update text in UI for rotation axis alignment

mathturtle:
@yijiang1 Here is the UI update requested in #260. I updated the buttons/labels to match the instructions too.


yijiang1:
The UI looks good to me.
What about the text? @Hovden @ElliotPadgett 
Screen shot:
screen shot 2015-11-13 at 3 16 28 pm

PR #269, opened 2015-11-20
Reorganize tomography menu

yijiang1:
Reorganize tomography menu as suggested by @ElliotPadgett
 
1. Rename some functions in the menu and clean up the code.
2. Add a simple UI to Direct Fourier Method reconstruction.
3. Disable the simple back projection function for now since the c++ code is not very optimized.
4. Move the "Generate Tilt Series" function to the end of the menu.

### DIFF
--- a/tomviz/AddPythonTransformReaction.cxx
+++ b/tomviz/AddPythonTransformReaction.cxx
@@ -333,7 +333,7 @@ OperatorPython* AddPythonTransformReaction::addExpression(DataSource* source)
     v->addLayout(layout2);
     v->addWidget(buttons);
     dialog.setLayout(v);
-    
+
     if (dialog.exec() == QDialog::Accepted)
     {
       QMap<QString, QString> substitutions;
@@ -344,7 +344,6 @@ OperatorPython* AddPythonTransformReaction::addExpression(DataSource* source)
       addPythonOperator(source, this->scriptLabel, this->scriptSource, substitutions);
     }
   }
-    
   else if (scriptLabel == "Delete Slices")
   {
     vtkTrivialProducer *t = vtkTrivialProducer::SafeDownCast(
@@ -381,7 +380,6 @@ OperatorPython* AddPythonTransformReaction::addExpression(DataSource* source)
       addPythonOperator(source, this->scriptLabel, this->scriptSource, substitutions);
     }
   }
-
   else if (scriptLabel == "Sobel Filter") //UI for Sobel Filter
   {
     QDialog dialog(pqCoreUtilities::mainWidget());
@@ -414,7 +412,6 @@ OperatorPython* AddPythonTransformReaction::addExpression(DataSource* source)
       addPythonOperator(source, this->scriptLabel, this->scriptSource, substitutions);
     }
   }
-    
   else if (scriptLabel == "Resample")
   {
     QDialog dialog(pqCoreUtilities::mainWidget());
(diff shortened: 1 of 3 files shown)


tomviz/AddPythonTransformReaction.cxx
```diff
@@ -522,6 +519,41 @@ OperatorPython* AddPythonTransformReaction::addExpression(DataSource* source)
...
+    connect(buttons, SIGNAL(rejected()), &dialog, SLOT(reject()));
+
+    v->addLayout(layout);
+    v->addWidget(buttons);
+    dialog.setLayout(v);
+    dialog.layout()->setSizeConstraint(QLayout::SetFixedSize); //Make the UI non-resizeable
+    if (dialog.exec() == QDialog::Accepted)
+    {
+      QMap<QString, QString> substitutions;
+      addPythonOperator(source, this->scriptLabel, this->scriptSource, substitutions);
```
mathturtle: Out of curiosity: this UI doesn't appear to do anything.  Did you mean to ask for the dimensions to reconstruct into?  Or is this just a 'Warning: this operation will take a while' dialog?

yijiang1: The latter. We want to provide a simple description that's similar to the other two reconstruction methods.

tomviz/AddPythonTransformReaction.cxx
```diff
@@ -522,6 +519,41 @@ OperatorPython* AddPythonTransformReaction::addExpression(DataSource* source)
       addPythonOperator(source, this->scriptLabel, this->scriptSource, substitutions);
     }
   }
+  else if (scriptLabel == "Reconstruct (Direct Fourier)")
+  {
+    vtkTrivialProducer *t = vtkTrivialProducer::SafeDownCast(source->producer()->GetClientSideObject());
+    vtkImageData *data = vtkImageData::SafeDownCast(t->GetOutputDataObject(0));
+    int *extent = data->GetExtent();
```
cryos: These three lines should just be removed - the result is never used.

tomviz/MainWindow.cxx
```diff
@@ -199,37 +199,32 @@ MainWindow::MainWindow(QWidget* _parent, Qt::WindowFlags _flags)
...
   ui.menuTomography->addSeparator();
-    
-  QAction *generateTiltSeriesAction = ui.menuTomography->addAction("Generate Tilt Series");
-  ui.menuTomography->addSeparator();
 
   QAction *subtractBackgroundAction = ui.menuTomography->addAction("Background Subtraction (Manual)");
-  QAction *alignAction = ui.menuTomography->addAction("Translation Align");
-  QAction *autoAlignAction = ui.menuTomography->addAction("Translation Align (Auto)");
-  QAction *rotateAlignAction = ui.menuTomography->addAction("Rotation Align");
+  QAction *alignAction = ui.menuTomography->addAction("Tilt Series Registration (Manual)");
```
cryos: Is registration really the right word here? I thought alignment communicated it more clearly, and registration risks being confused with other things.

yijiang1: According to wikipedia, "Image registration is the process of transforming different sets of data into one coordinate system", which is what we are doing here. 
However, I do think some users may not be familiar with the term.

cryos: I found the alignment term more immediately obvious, I think alignment is more intuitive to me about the purpose and registration while it may apply is not a word I would immediately associate with this functionality.

ElliotPadgett: @cryos @yijiang1 I suggested "Tilt Series Registration" mostly because I don't like "Translation Alignment", since the "Rotation Axis Alignment" also does translation.  I agree that calling it registration may be confusing, and is not standard for us either.

Perhaps a better option would be to just call it "Image Alignment" to clarify that it is operating on images individually rather than the stack as a whole, which "Rotation Axis Alignment" does.  @Hovden what do you think?

cryos: Yes, I think image and rotation axis alignment would make more sense, and if it isn't standard for you then I think keeping the word alignment is far more likely to guide people to the thing they want than registration is.

ercius: Alignment is the typical word used instead of registration.

Hovden: Ok. Let's call it "Image Alignment (Manual)".

Also "Rotation Axis" should be called "Tilt Axis Alignment"

tomviz/AddPythonTransformReaction.cxx
ElliotPadgett: One other minor improvement would be to put the automatic alignment above the manual one in the menu, since users are most likely to do the automatic one first, and then refine manually.

PR #271, opened 2015-11-23
Create default orbit camera animation on load

### LINKED ISSUE #131: Make a default animation when loading.
Currently the user must create an animation in order for it to show any functionality. Having a default animation, such as a simple rotation, will add a lot of value to new users trying to learn the animation tool. It will also make the play button functional when the software is first loaded.

Suggestion:
Make a default animation that rotates about each axis sequentially.

mathturtle:
@cryos This works mostly as we discussed now.  The only issue still remaining is this: when you delete all the data sources in the scene and then open another one it creates a second orbit animation.

 Fixes #131

### DIFF
--- a/tomviz/Utilities.cxx
+++ b/tomviz/Utilities.cxx
@@ -16,7 +16,16 @@
 #include "Utilities.h"
 
 #include "DataSource.h"
+#include "pqAnimationCue.h"
+#include "pqAnimationManager.h"
+#include "pqAnimationScene.h"
+#include "pqPVApplicationCore.h"
+#include "pqSMAdaptor.h"
+#include "vtkBoundingBox.h"
+#include "vtkCamera.h"
+#include "vtkImageData.h"
 #include "vtkNew.h"
+#include "vtkPoints.h"
 #include "vtkPVArrayInformation.h"
 #include "vtkPVDataInformation.h"
 #include "vtkPVDataSetAttributesInformation.h"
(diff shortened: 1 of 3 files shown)


tomviz/Utilities.cxx
```diff
@@ -240,4 +252,56 @@ QString readInPythonScript(const QString &scriptName)
   return "raise IOError(\"Couldn't read script file\")\n";
 }
 
+void createCameraOrbit(vtkSMSourceProxy *data, vtkSMRenderViewProxy *renderView)
```
cryos: This file is beginning to look like a dumping ground for 'misc'. I think we should just merge this as-is, but consider bringing some pieces out into their own more focused source files. I will be pulling the Python loading code into a separate class, I think the animation code would probably benefit from the same.

tomviz/Utilities.cxx
```diff
@@ -240,4 +252,56 @@ QString readInPythonScript(const QString &scriptName)
...
+{
+  // Get camera position at start
+  double *normal = renderView->GetActiveCamera()->GetViewUp();
+  double *origin = renderView->GetActiveCamera()->GetPosition();
+
+  // Get center of data
+  double center[3];
```
cryos: You might consider classes like vtkVector in future, can be called with the standard methods (pass the pointer to the start of the double array), but have more API and more clearly express what the object is.

cryos:
Looks good to me, few inline comments. High level - it might be nice to have more frames in the initial orbit - this is an enormous improvement.

Hovden:
Yeah. More frames is better.

PR #272, opened 2015-11-23
Symlinks should be preferred on OSes that support them

cryos:
This leads to less confusion when editing the Python files, and clearly
expresses our intent when the operating system support is available.

Add the Python files to our SOURCES list so that they show up in IDEs.

### DIFF
--- a/tomviz/CMakeLists.txt
+++ b/tomviz/CMakeLists.txt
@@ -171,7 +171,7 @@ foreach(file ${python_files})
   execute_process(COMMAND ${CMAKE_COMMAND} -E ${script_cmd}
     "${tomviz_SOURCE_DIR}/tomviz/python/${file}"
     "${tomviz_BINARY_DIR}/share/tomviz/scripts/${file}")
-  install(FILES ${file}
+  install(FILES "python/${file}"
     DESTINATION "${tomviz_data_install_dir}/scripts"
     COMPONENT "Scripts")
 endforeach()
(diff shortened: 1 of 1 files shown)


tomviz/CMakeLists.txt
```diff
@@ -158,10 +158,22 @@ set(python_files
...
+make_directory("${tomviz_BINARY_DIR}/share/tomviz/scripts")
 foreach(file ${python_files})
-  set(file python/${file})
-  file(COPY ${file} DESTINATION "${tomviz_BINARY_DIR}/share/tomviz/scripts")
-  install(FILES ${file} DESTINATION "${tomviz_data_install_dir}/scripts" COMPONENT "Scripts")
+  list(APPEND SOURCES "python/${file}")
+  execute_process(COMMAND ${CMAKE_COMMAND} -E ${script_cmd}
+    "${tomviz_SOURCE_DIR}/tomviz/python/${file}"
+    "${tomviz_BINARY_DIR}/share/tomviz/scripts/${file}")
+  install(FILES ${file}
```
mathturtle: This should be `install(FILES python/${file}`.  Right now `make install` fails since it cannot find the file.

cryos: Yes, I forgot to test make install. Good catch - thanks Shawn!

cryos:
Should work now

mathturtle:
`make install` works now.  Hopefully this is all that is required to avoid breaking superbuild.

cryos:
It should be, this topic is much more about the build tree when actively developing, and far less about the install tree.

Issue #62, opened 2015-05-31
Manual alignment: zoom to any point


ElliotPadgett:
Currently the widget allows the user to zoom in and out in the data view, but the zoom is always centered on the data center point.  It would be useful to be able to zoom in on any part of the data.

mathturtle:
The zoom is about the 'center' of the scene.  The center of the scene is also the center of rotation and can be set with the button with this icon:
![pqpickcenter24](https://cloud.githubusercontent.com/assets/553788/10405090/bcdb3b8c-6ea4-11e5-88aa-1d792c57ddb3.png)
Click on this button, then on the part of the scene where you want the new center.  It isn't the best iterface, but it may help with this problem.

Another option would be the zoom to box button which lets you select the part of the scene to zoom in on.  It looks like this:
![pqzoomtoselection24](https://cloud.githubusercontent.com/assets/553788/10405192/51d7c278-6ea5-11e5-8a8b-b4c0bdf245e2.png)

Do either of these solve this, or do I need to look into another solution?

Hovden:
@mathturtle the tools you are talking about do not exist in the manual alignment widget. They are for 3D visualization.

mathturtle:
Oh.  I can get this effect in the alignment tool by click-and-dragging with the middle mouse button to pan the image so that the portion of interest is in the center and then zooming.  (Of course on OSX you may not have a middle mouse button...)  I can look into changing the zoom behavior in this type of view.  It is a VTK level change that will have to be discussed with the VTK community.

yijiang1:
#278 works well. 
Do we need a button or something that switch back to the initial view (i.e. the entire image)? @ElliotPadgett  Thus it's easier for users to look around and find the best feature for alignment.

ElliotPadgett:
Yes, zooming back out to the initial view is a requirement.

mathturtle:
Please try out this new branch where I have added this button

Hovden:
Works on OSX.

ElliotPadgett:
This works well, but the buttons are somewhat hidden up at the top.  Can we move them down to be near the other controls?

PR #278, opened 2015-11-24
Add zoom to selection button to manual align UI

mathturtle:
This is to partially address #62.  It adds a button that allows the user to select a box of interest and then zooms to that box similar to the button for the 3D view.  This is still not ideal, but for a quick fix to make this UI easier to use it will work.


mathturtle:
@ElliotPadgett @Hovden

Hovden:
OK. Let's merge and test. @ElliotPadgett

PR #281, opened 2015-11-24
Added labels to Tomography menu

yijiang1:
@Hovden @ElliotPadgett How does this look?
screen shot 2015-11-24 at 4 49 20 pm

### DIFF
--- a/tomviz/MainWindow.cxx
+++ b/tomviz/MainWindow.cxx
@@ -63,7 +63,6 @@
 #include <QFileInfo>
 #include <QMessageBox>
 #include <QTimer>
-#include <QFont>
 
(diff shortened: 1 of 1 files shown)


ElliotPadgett:
Visually those are nice.  I think "Image Alignment" and "Tilt Axis
Alignment" should be grouped under a subheading called "Alignment:".  "Set
Tilt Angles" and "Background Subtraction" could probably go together in a
subheading called "Preprocessing" or something like that.

Hovden:
Just change "Data Processing" to "Pre-reconstruction Processing" or "Preprocessing".

When this menu grows substantially we will split it as @ElliotPadgett suggested.

Issue #273, opened 2015-11-23
Slowdown (memory leak?) when using data transforms.


Hovden:
The program gets slower and slower then crashes when using data transforms.

To repeat, simply crop the data. Then delete the crop transform and crop again and delete. Keep repeating and after 2 or three times it will be slow and probably crash.

mathturtle:
I've seen this before.  I've never actually gotten it to crash (I usually killed it once the transforms were taking 10 minutes each or so).

mathturtle:
It doesn't appear to be a memory leak (at least not on Linux).  I have seen a few crashes while testing it, but I am unable to reproduce them consistently.  It appears to be a memory corruption issue of some sort for the crashes (maybe something to do with crop itself?).  But the slowdown seems to be a separate problem.

mathturtle:
I figured out the crash.  It was a threading bug.  It has to do with our model of shallow copying the output of a transform into the input vtkImageData.  If you have a large dataset and a fast operation like crop, it is possible for the operation to finish before the histogram building code is done with the dataset.  Then you are shallow copying over something that another thread is using (and deleting the old data out from under the background thread, which was causing the crash).

mathturtle:
As for the slowness, all of the times I have interrupted the code in the debugger it has been in VTK's garbage collection/destructor code.  I'm not sure why that would take longer the more the program runs though.  Maybe @cryos has an idea?

mathturtle:
CentralWidget::refreshHistogram is the culprit for the slowness.  All the stack traces pass through there.

cryos:
We may need to add a QTimer that delays any refresh, and extend the code to be more interrupt friendly if it is being invalidated frequently. It should stay in a background thread, but if it is causing slowdowns enabling it to exit early when the result may not be used would be advantageous.

ElliotPadgett:
I think this is a high priority issue.  The normal use case will be 1024X1024X100 data sets with 5-10 data transforms.  This slowdown is making it practically impossible for me to get through a realistic data set now.

mathturtle:
@ElliotPadgett I agree.  Is there an older version of tomviz that doesn't have this bug?  I can guess, but if you have any information about when this was introduced please let me know.

PR #282, opened 2015-11-24
Histogram minor improvements

mathturtle:
@Hovden I think this should fix the crash you are seeing in #273, the slowdown is more complicated and will probably have to wait.


cryos:
LGTM

PR #286, opened 2015-11-25
Change default animation to have 200 frames -- fixes #284

### LINKED ISSUE #284: Default Animation Needs More Frames (200)
When a user presses play using the default animation, the object frantically spins. The default number of frames is much too low (10) and should be increased. I suggest something at or around 200 frames.

mathturtle:
@Hovden


mathturtle:
I'm not 100% sure this is the right way to do this, but it works for now.

Issue #292, opened 2015-11-25
Python Generated Data Crashes


Hovden:
The "Random Particle" and "Constant Dataset" crashes after a couple uses.

mathturtle:
This should be fixed now.

PR #295, opened 2015-11-30
Fix python-related crashes in generating custom datasets

mathturtle:
@Hovden @yijiang1 Can you test if this branch fixes the crashes you were seeing in the python custom sample data (#292)?


yijiang1:
Yup. It doesn't crash anymore.

Hovden:
Great. Once these python bugs are fixed we should update to a 0.7.1 release.

mathturtle:
@Hovden I think we are planning a 0.7.1 release, if only for this and #296.

PR #296, opened 2015-12-01
Fix transform slowdown

### LINKED ISSUE #273: Slowdown (memory leak?) when using data transforms.

mathturtle:
@cryos I found it.  I was so sure that refreshHistogram was only called once, but after talking to you I checked in kcachegrind... It was called 11 million times.  I was calling QObject::disconnect on the wrong object (see first commit message) and the number of connections on the signal dataModified was growing exponentially every time something changed the data.  And git blame points straight at me and @utkarshayachit for the lines of code in question.

The second commit in this branch does the event collapsing we talked about with a QTimer.  It also fixed the problem, but it wasn't the real fix.

@Hovden @ElliotPadgett @yijiang1 This should fix #273.

### DIFF
--- a/tomviz/CentralWidget.cxx
+++ b/tomviz/CentralWidget.cxx
@@ -320,7 +320,7 @@ void CentralWidget::setActiveDataSource(DataSource* source)
 {
   if (this->AModule)
   {
-    this->disconnect(this->AModule);
+    this->AModule->disconnect(this);
     this->AModule = nullptr;
   }
   this->setDataSource(source);
@@ -331,7 +331,7 @@ void CentralWidget::setActiveModule(Module* module)
 {
   if (this->AModule)
   {
-    this->disconnect(this->AModule);
+    this->AModule->disconnect(this);
   }
   this->AModule = module;
   if (this->AModule)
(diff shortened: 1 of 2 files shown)


cryos:
Glad you managed to track it down, like I said the only way I could see that level of allocation/deallocation is if we were calling it thousands (minimum) but millions would do it. Thanks for persevering.

tomviz/CentralWidget.cxx
```diff
@@ -432,7 +437,12 @@ void CentralWidget::setDataSource(DataSource* source)
 
 void CentralWidget::onColorMapUpdated()
 {
-  this->refreshHistogram();
+  this->onDataSourceChanged();
+}
+
+void CentralWidget::onDataSourceChanged()
+{
+  this->Internals->Timer.start();
```
cryos: Might be useful to add a comment explaining this for those not knowing the API so well
// Start (or restart) the one shot timer for delayed histogram refresh.

cryos:
Looks good to me, like I said inline it might be worth adding a comment to make the intent clear (to start/restart the one shot timer). Glad you dug down to the real cause, but squashing many updates down to one or two is useful for many reasons.

yijiang1:
Yup doing multiple data transforms is pretty fast now. :thumbsup:

mathturtle:
I've added the inline comment about what starting that timer in the slot does.

Issue #288, opened 2015-11-25
Add New Matplotlib Colormaps


Hovden:
Here are the new colormaps used in matplotlib: http://matplotlib.org/1.5.0/users/whats_new.html#new-colormaps

They've been designed by color theorists, a great discussion is given here: https://bids.github.io/colormap/

There are four colormaps released under a CC0 license. I personally like "magma" and "viridis".

mathturtle:
I have a work in progress branch for this, but I'm going to need to talk to @utkarsh.ayachit when he gets back to figure out the last piece of it.

mathturtle:
The branch for this is merged... I also have a ParaView branch to add these upstream.  We'll have to take them out when that is merged (or we can have duplicates...)

PR #303, opened 2015-12-04
Import matplotlib colormaps

mathturtle:
@Hovden This branch adds the colormaps from matplotlib that you mentioned in #288.

@cryos These are CC0 licensed, but I'd like to figure out how to at least mention where we got them.  You can't comment Json files, and ParaView uses Json files to store colormaps.  We also need to figure out an appropriate color for 'Not a Number' values (right now I have is set to white for all of them).  I'm also abusing the python script copy and install mechanisms to get the json file included...


cryos:
Won't ParaView ignore any keys it doesn't recognize? So you could add a key "source" with a value of a URL, even an "attribution" key. Not familiar with ParaView's JSON parsing, but I would imagine these additional keys will be ignored (and could in future be added as optional keys displayed in the interface).

mathturtle:
Good idea.  I've updated the file to attribute the color maps and mention that they are CC0 licensed...

mathturtle:
I was talking to @utkarshayachit yesterday and he said that we should import these into ParaView as well.  So we can either merge this or wait for ParaView to include them and update ParaView...

cryos:
I think merging this now so people can play with it is fine, we can always remove them once they appear in ParaView.

Issue #299, opened 2015-12-03
2D sobel filter


ercius:
The sobel filter in the Data Transforms menu does not do what a user would want. Scipy's sobel is only a 1D filter, but most image analysis should be done in 2D. See attached python script (TXT format) to calculate a 2D sobel filter on a input data set in tomviz.

[2Dsobel.txt](https://github.com/OpenChemistry/tomviz/files/51270/2Dsobel.txt)

ercius:
Removed a couple of extraneous lines of code. Also, I cast the array to np.float32 before operating to ensure that the values can be stored.
[2Dsobel.txt](https://github.com/OpenChemistry/tomviz/files/51286/2Dsobel.txt)

yijiang1:
> Scipy's sobel is only a 1D filter
You are right. "scipy.ndimage.filters.sobel" calculates an approximation of the gradient of the image along a given direction.

> but most image analysis should be done in 2D
Why can't we do this in 3D? You can do this by simply calling 
```python
result = scipy.ndimage.filters.generic_gradient_magnitude(array, scipy.ndimage.filters.sobel)
```

ercius:
There are two different use cases here:

1 use case is to sobel filter the 3D volumetric data set. I think generic_gradient_magnitude will accomplish this. Thanks for mentioning this.

The other use case is to sobel filter each image separately in a tilt series to aid with automated alignment. In this case you want to loop over all projection images and apply a 2D sobel filter to each image then return the filtered stack. My code implements this second use case.

yijiang1:
I've changed the Sobel filter in "Data Transforms" menu to a 3D transform and added the 2D filter to "Tomography" menu.

ercius:
I think that is a good option. Thanks.

yijiang1:
Tested well on binaries.

PR #304, opened 2015-12-04
Added 3D Sobel Filter

yijiang1:
Changed the "sobel filter" in Data Transforms menu to 3D instead of 1D.
This addresses case 1 in #299 .

### DIFF
--- a/tomviz/AddPythonTransformReaction.cxx
+++ b/tomviz/AddPythonTransformReaction.cxx
@@ -379,40 +379,6 @@ OperatorPython* AddPythonTransformReaction::addExpression(DataSource* source)
       addPythonOperator(source, this->scriptLabel, this->scriptSource, substitutions);
     }
   }
-
-  else if (scriptLabel == "Sobel Filter") //UI for Sobel Filter
-  {
-    QDialog dialog(pqCoreUtilities::mainWidget());
-    QHBoxLayout *layout = new QHBoxLayout;
-    QLabel *label = new QLabel("Axis:", &dialog);
-    layout->addWidget(label);
-    QComboBox *axis = new QComboBox(&dialog);
-    axis->addItem("X");
-    axis->addItem("Y");
-    axis->addItem("Z");
-    axis->setCurrentIndex(2);
-    layout->addWidget(label);
-    layout->addWidget(axis);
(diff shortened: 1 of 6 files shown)


Hovden:
Does the gradient magnitude use the same laplace kernel? If not, do we need to change the name from "Sobel" to "Gradient Magnitude"?

We can merge this if it's ready.

yijiang1:
> Does the gradient magnitude use the same laplace kernel?

Laplace operator (filter) approximates the second derivative of the image.
Sobel operator (filter) approximates the first derivative. 

For edge detection, we often do two things:
1. Calculate the gradient image, which depends on the operator we use.
2. Calculate the magnitude of the gradient image.

Hovden:
My question, pertained more to the algorithm being used. But looking more closely I see that the function takes in an argument that specifies that it will calculated using the sobel filter:

scipy.ndimage.filters.generic_gradient_magnitude(array, scipy.ndimage.filters.sobel)

But I'm still confused as to what the axial direction of the sobel kernel is and how the user selects it.

yijiang1:
It first approximates the gradients along all three directions (say, Gx, Gy, Gz) using Sobel operator and then calculates the magnitude of the gradient image (sqrt(Gx^2+Gy^2+Gz^2)). Thus the user doesn't need to select anything.

In the old version it basically calculates Gx, Gy, or Gz, which can be specified by the user.

Based on #299 , I thought we want the gradient magnitude image instead of a gradient along a particular direction.

ercius:
I think the new 3D function accomplishes what most users would want. I.e.
along all directions simultaneously.

yijiang1:
@Hovden Can we merge this now?

Hovden:
I'm still confused why it is being called a Sobel filter when it using the gradient magnitude.

yijiang1:
So we call it "Gradient Magnitude using Sobel Filter"?

Hovden:
We could call it "Gradient Magnitude" and then in the Python file, add a comment explaining how it is calculated.

yijiang1:
OK.

yijiang1:
I changed "sobel filters" to "gradient magnitude" and added descriptions in python code.

PR #305, opened 2015-12-04
Added 2D sobel filter to tilt series

yijiang1:
Added 2D sobel filter to tilt series, as required in #299 .
The function is listed in "Tomography" menu and apply a 2D sobel filter to each image in the series.


Hovden:
Merge AFTER 0.7.1 release.

PR #312, opened 2015-12-07
Added auto background subtraction for tilt series

yijiang1:
Added a simple auto background subtraction for tilt series.
For each tilt image, the method calculates its histogram and then chooses the one with highest peak as the background level. It does NOT set negative pixels to zero.


Hovden:
Great. Can you add your description to the python code?

i.e. "#For each tilt image, the method calculates its histogram and then chooses the highest peak as the background level and subtracts it from the image. It does NOT set negative pixels to zero."

yijiang1:
Sure. I'll add the description on a separate branch.

PR #320, opened 2015-12-11
Make align an operator

### LINKED ISSUE #310: Make Manual Image Alignment generate python transform
The manual alignment widget should generate a python transform to shift the data rather than cloning the data and creating a new shifted data set.

The python script should display the shifts that it is applying.

mathturtle:
This branch fixes #310.  It changes the manual image alignment into an operator that applies to a data source instead of creating a new data source.  I also add a table to the manual alignment UI to show all of the offsets rather than just the ones for the current slice.

@ElliotPadgett @ercius @Hovden @cryos

### DIFF
--- a/tomviz/AlignWidget.cxx
+++ b/tomviz/AlignWidget.cxx
@@ -50,6 +50,7 @@
 #include <QKeyEvent>
 #include <QTableWidget>
 #include <QTableWidgetItem>
+#include <QHeaderView>
 #include <QButtonGroup>
 
 namespace tomviz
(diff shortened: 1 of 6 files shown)


yijiang1:
It may be useful to add another column that shows tilt angles. Some users like to start from the 0 degree tilt image.

tomviz/AlignWidget.cxx
```diff
@@ -47,14 +48,16 @@
...
 #include <QButtonGroup>
 
 namespace tomviz
 {
 
-AlignWidget::AlignWidget(DataSource* d, QWidget* p, Qt::WindowFlags f)
-  : QWidget(p, f), timer(new QTimer(this)), frameRate(5),
-    unalignedData(d), alignedData(nullptr)
+AlignWidget::AlignWidget(TranslateAlignOperator *op, QWidget* p)
+  : EditOperatorWidget(p), timer(new QTimer(this)), frameRate(5),
```
yijiang1: I think the default frameRate is actually 10 instead of 5.
In line 251, you have timer->start(100) and 100 = 1000/10.

mathturtle:
Ok, I've added that column.  And fixed the starting timer interval to match the frame rate of 5.

yijiang1:
Looks pretty good to me.

Issue #222, opened 2015-09-29
Add new paraview axis grid


Hovden:
We need good scale markers for scientific visualization. It looks like the new axis grid implemented in Paraview could address this: http://www.kitware.com/blog/home/post/871

mathturtle:
It turns out that the axis grid breaks volume rendering somehow.  I've reported the bug and they are working on it.

mathturtle:
Ok, that was easy... after hours of searching how to enable it: It's already there!  It is hiding in the view properties panel.  There is a checkbox with no label that has a button to it's right instead.  The button says Edit Axes Grid and can be used to change parameters of the axes grid.  The checkbox turns it on or off.

Note: there is a bug where volume rendering fails when the Axes Grid is enabled and will not work again in that view even if the Axes Grid is turned off.  I have reported it and the volume rendering folks are looking into it right now.

mathturtle:
![checkbox](https://cloud.githubusercontent.com/assets/553788/10404337/f270c848-6e9f-11e5-89b6-625d48e6ee09.png)

This checkbox is the one I mentioned above.

mathturtle:
Ok, it is there... We want to expose it more.  Where does this go?  Menu item?  Button in the toolbars?  I don't think we can edit the buttons directly above the view...

PR #321, opened 2015-12-11
Add options for parallel and orthographic project to view menu

### LINKED ISSUE #114: Adjust camera perspective
It is important to be able to adjust the camera perspective. [probably good in both Paraview and tomviz view properties]

The best solution would be in the view properties having a perspective slider or input that allows the user to pick the level of perspective wanted. Exploring data with no perspective is a useful task for producing interpretable scientific visualization or looking down a high-symmetry zone axis.

For the simplest solution, this would be enabling rotation in the 2D camera view mode, where the user can flip between no perspective (2D mode) and some perspective (3D mode).

mathturtle:
@cryos @Hovden This fixes #114 by exposing the right way to set the projection type in the view menu.  We should think about disabling the tiny buttons above the view since we don't really support using them.

### DIFF
--- a/tomviz/ViewMenuManager.cxx
+++ b/tomviz/ViewMenuManager.cxx
@@ -71,11 +71,17 @@ void ViewMenuManager::buildMenu()
   {
     perspectiveProjectionChecked = this->perspectiveProjectionAction->isChecked();
   }
+  bool axisGridIsShowing = false;
+  if (this->showAxisGridAction)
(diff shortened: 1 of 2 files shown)


mathturtle:
Added a menu item to expose the axes grid.  This now also fixes #222.

Hovden:
Great. I agree we should discuss the tiny icons.

Issue #257, opened 2015-11-12
Deleting a data transform reset tilt angles to zero


yijiang1:
1. Load a dataset.
2. Mark it as tilt series. Auto fill the tilt angles. 
4. Do python data transform.
5. Delete the transform from the pipeline.
6. Now all tilt angles are reset to zero.

mathturtle:
I see what is happening here.  When the transform is deleted, the code doesn't know if the tilt angles came from the transform (i.e. your convert to tilt series transform) or from the user.  So they are removed.  Then, since the dataset is still marked as a tilt series a new tilt angles array is created with all 0s.

One solution would be to make the 'auto fill tilt angles' into a transform that stores what the user did to the tilt angles...

yijiang1:
> make the 'auto fill tilt angles' into a transform

I think @Hovden does not want this...

mathturtle:
To really address this we need to save where the tilt angles were added/modified in the sequence of transforms applied to a dataset.  The easiest way to do this would be to insert these events as new transforms.  I'll try to think of something else if @Hovden thinks this is a bad idea.

PR #322, opened 2015-12-13
Make tilt angles more permanent.

### LINKED ISSUE #316: Copy tilt angles when cloning datasets
When users clone a dataset that's marked as tilt series, tomviz should always carry the tilt angles to new dataset.

Right now, this happens only if users choose the "Transformed data only" option. However, it should do the same for "Original data only" and "Original data with transformations". This may be related to #257 

We should treat tilt angles as an intrinsic property of the dataset instead of a transform.

mathturtle:
They are kept when the original data is cloned or when transforms are
deleted now.  This should fix #316 and #257.  @yijiang1 can you please test this?  A lot.  This is a large change internally and I've tested all the cases I can think of but it is very possible that I missed a few.

The branch name is the way I initially planned to implement this, but it turned out to be easier to just make the tilt angles a separate part of the data source.

### DIFF
--- a/tomviz/DataPropertiesPanel.cxx
+++ b/tomviz/DataPropertiesPanel.cxx
@@ -220,20 +220,14 @@ void DataPropertiesPanel::update()
     this->Internals->TiltAnglesSeparator->show();
     ui.SetTiltAnglesButton->show();
     ui.TiltAnglesTable->show();
-    vtkDataArray* tiltAngles = vtkAlgorithm::SafeDownCast(
-        dsource->producer()->GetClientSideObject())
-      ->GetOutputDataObject(0)->GetFieldData()->GetArray("tilt_angles");
-    ui.TiltAnglesTable->setRowCount(tiltAngles->GetNumberOfTuples());
-    ui.TiltAnglesTable->setColumnCount(tiltAngles->GetNumberOfComponents());
-    for (int i = 0; i < tiltAngles->GetNumberOfTuples(); ++i)
+    QVector<double> tiltAngles = dsource->getTiltAngles();
+    ui.TiltAnglesTable->setRowCount(tiltAngles.size());
+    ui.TiltAnglesTable->setColumnCount(1);
+    for (int i = 0; i < tiltAngles.size(); ++i)
     {
-      double* angles = tiltAngles->GetTuple(i);
-      for (int j = 0; j < tiltAngles->GetNumberOfComponents(); ++j)
-      {
-        QTableWidgetItem* item = new QTableWidgetItem();
-        item->setData(Qt::DisplayRole, QString("%1").arg(angles[j]));
-        ui.TiltAnglesTable->setItem(i, j, item);
-      }
+      QTableWidgetItem* item = new QTableWidgetItem();
+      item->setData(Qt::DisplayRole, QString::number(tiltAngles[i]));
+      ui.TiltAnglesTable->setItem(i, 0, item);
     }
   }
   else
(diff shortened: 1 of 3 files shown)


yijiang1:
The branch works well except for some problems related to transforms that modify tilt angles (such as "delete slices" and "resample"). It seems that tomviz can't "remember" the tilt angles before transforms.

Problem 1:
1. Load a dataset, mark it as tilt series, auto fill the angles.
2. Do "Delete Slices" transform. Delete the first tilt image (in z direction).
3. Delete the transform in the pipeline, now the a random angle is added to the end of tilt series array. 
4. Moreover, now if you update the tilt angles using the "Set Tilt Angles" button, tomviz will crash.

Problem 2:
1. Load a dataset, mark it as tilt series, auto fill the angles.
2. Do "Delete Slices" transform. Delete the first tilt image (in z direction).
3. Clone the dataset by choosing "Original data only" or  "Original data with transformations" option.
4. Crash.

mathturtle:
I told @cryos that you cared about cases where transforms modified the tilt angles and then "remembering" the previous tilt angles when they are deleted.  These actually make this a very tricky bug.  The crashes are my fault (not resizing the tilt angles array when the data is resized by a transform and then that transform is deleted).

I can fix this case.  Essentially I'll add another level to the "remembering" so that it remembers the tilt angles that the user set manually and assume these are the base tilt angles without any transforms applied to the data.  That is what I intended with this branch... I just forgot a case.

There is a case that I don't know if we'll ever be able to do correctly which I'm not going to handle.  If the user auto fills the tilt angles, applies a transform such as delete slice that modifies the tilt angles and then manually edits the tilt angles, I don't know what to do with the tilt angles if the transform is deleted (I'm not sure you CAN know without asking the user).

mathturtle:
I've uploaded a new version.  Can you please test this one?

yijiang1:
> If the user auto fills the tilt angles, applies a transform such as delete slice that modifies the tilt angles and then manually edits the tilt angles

Right now this will make tomviz crash on new branch.

> I don't know what to do with the tilt angles if the transform is deleted

I think we don't need to worry about it for now. Just make sure we can restore the angles before transforms.

mathturtle:
> Right now this will make tomviz crash on new branch.

Good catch.  I've fixed this and uploaded a new version.

yijiang1:
It doesn't crash anymore. 
If I do a transform, manually edit tilt angles, then delete the transform, can I get the angles before I apply transforms?
For example, say we have three tilts with tilt angles = [1, 2, 3]. 
1. I delete the first tilt image, and the tilt angle array becomes [2, 3].
2. I manually change the angles to [4, 6].
3. Now if I remove the "delete slice" transform, is it possible to get [1, 2 ,3] back? Currently, I would get [4, 6, 3].

mathturtle:
> If I do a transform, manually edit tilt angles, then delete the transform, can I get the angles before I apply transforms?

Done.  Update pushed.  Just be aware that deleting any transform will remove changes you have made to the tilt angles after they were modified by transforms and reset them to the original tilt angles before reapplying the remaining transforms.

yijiang1:
Nice.
One more thing: Cloning "Original data with transformations" doesn't carry the tilt angles before transforms properly. 
Say we have three tilts with tilt angles = [1, 2, 3]. 
1. I delete the first tilt image, and the tilt angle array becomes [2, 3].
2. Clone the data with "Original data with transformations" option.
The newly cloned data has tilt angles = [0, 2], which is wrong.
3. Delete the "delete slice" transform in cloned data, the tilt angle becomes [1, 2, 0].

mathturtle:
Another indexing problem... Nice catch.  Updated version ready.

yijiang1:
Great! I can't find any more bugs.

mathturtle:
Alright, I'm going to merge this then.

PR #325, opened 2015-12-18
Change contour module to use flying edges

mathturtle:
@cryos I got sick of fighting with interaction and picking, so I decided to work on this today.  It was pretty easy to get it working.  (At least I assume it works.  The contour is reasonable and it uses all my cores at 100% for a second or two on the large test dataset which seems to imply tbb is working).

### DIFF
--- a/tomviz/ModuleContour.cxx
+++ b/tomviz/ModuleContour.cxx
@@ -63,7 +63,7 @@ bool ModuleContour::initialize(DataSource* data, vtkSMViewProxy* vtkView)
   vtkSMSessionProxyManager* pxm = producer->GetSessionProxyManager();
 
   vtkSmartPointer<vtkSMProxy> proxy;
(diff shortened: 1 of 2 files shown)


Hovden:
Very cool. Excited to test on large datasets.

cryos:
That is great, and it looks good to me. It is also easy to turn back off, I am on battery and so can't compile but one question is whether flying edges will work without TBB? Otherwise, we will either want to enable it optionally based upon build parameters, or figure something else out. Don't know flying edges well enough, but an ifdef would be simple to add here to make it dependent on whether we are building with TBB.

cryos:
This looks great, merging now.

Hovden:
Much faster.

mathturtle:
To answer Marcus's question: I think it will work even without TBB
enabled.  That is the goal of the SMPTools stuff.  It will just be slow
without TBB.

PR #326, opened 2015-12-18
Move objects in the scene

mathturtle:
This pull request starts work on #38 by using ParaView's picking to select what is being clicked on and moving it.  Right now the behavior is always on (for testing) and if you click on something it recognizes it will move that data source.  I'm not getting mouseMove events for some reason though, so the move will not be interactive.  There are also two issues with picking (that I know of): namely that orthogonal slice and volume datasets are ignored by it (some limitation in ParaView's picking).  Also, since we don't have a representation for the vtkNonOrthogonalSliceWidget, it is ignored by the picking mechanism.

Also we are waiting on this ParaView MR: https://gitlab.kitware.com/paraview/paraview/merge_requests/539

@cryos @utkarshayachit

### DIFF
--- a/tomviz/Behaviors.cxx
+++ b/tomviz/Behaviors.cxx
@@ -16,6 +16,7 @@
 #include "Behaviors.h"
 
 #include "AddRenderViewContextMenuBehavior.h"
+#include "MoveActiveObject.h"
 #include "pqAlwaysConnectedBehavior.h"
 #include "pqApplicationCore.h"
 #include "pqDefaultViewBehavior.h"
@@ -101,6 +102,8 @@ Behaviors::Behaviors(QMainWindow* mainWindow)
   //new tomviz::ScaleActorBehavior(this);
 
   new tomviz::AddRenderViewContextMenuBehavior(this);
+
+  new tomviz::MoveActiveObject(this);
 
   vtkNew<vtkSMTransferFunctionPresets> presets;
   bool needToAddMatplotlibColormaps = true;
--- a/tomviz/CMakeLists.txt
+++ b/tomviz/CMakeLists.txt
@@ -85,6 +85,8 @@ set(SOURCES
   ModuleThreshold.h
   ModuleVolume.cxx
   ModuleVolume.h
+  MoveActiveObject.cxx
+  MoveActiveObject.h
   Operator.cxx
   Operator.h
   OperatorPython.cxx
--- a/tomviz/DataSource.cxx
+++ b/tomviz/DataSource.cxx
@@ -31,6 +31,7 @@
 #include "vtkSMTransferFunctionManager.h"
 #include "vtkTrivialProducer.h"
 #include "vtkTypeInt8Array.h"
+#include "vtkVector.h"
 
 #include <vtk_pugixml.h>
 
@@ -47,8 +48,8 @@ public:
   QList<QSharedPointer<Operator> > Operators;
   vtkSmartPointer<vtkSMProxy> ColorMap;
   DataSource::DataSourceType Type;
-
   vtkSmartPointer<vtkDataArray> TiltAngles;
+  vtkVector3d DisplayPosition;
 
 //-----------------------------------------------------------------------------
 // Checks if the tilt angles data array exists on the given VTK data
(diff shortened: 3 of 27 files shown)


mathturtle:
@cryos I've updated this to use the same widget as ParaView's transform filter (or our crop operator).  When the data source (and not a module) is selected, the widget will appear and can be translated.  All of the modules support this kind of translation except for non-ortho-slice since that one is a custom vtk filter/representation/widget thing.  I'm going to work on adding support on that one now.  Feel free to try this out.  It is much less WIP now, but I'm leaving the title until I've fixed the non-ortho-slice.

mathturtle:
Turns out that translating the non-ortho-slice is much harder than I thought it was.  The arrow moves right, but the plane stays in the original bounds and moves around as you translate the data source...  There isn't (in the current model) any way to make the plane, texture it and then 'display it over there'...  I'll have to think about this one some more.

cryos:
I don't think there is a good reason to translate the non-orthoslice, or the orthoslice. We should work on using the right views to move volumes, and possibly adding some custom views such as a blended slice showing two volumes in the future.

cryos:
Misunderstood the first time... We can live with that one not working at first, if necessary, but we will need to ensure all of our modules support translation.

cryos:
It shouldn't be too hard to translate the plane that gets drawn, you would certainly want to let it do virtually everything in its normal coordinate space, and then just apply the transform matrix when we draw the plane/triangles that are being textured. We should take a look at this next week, the non-ortho could do with some attention to improve the widget used too.

mathturtle:
I agree that it *shouldn't* be too hard.  But somehow it is.  I have changes that add a UserTransform matrix to all of the actors the widget uses... but it doesn't work.  I have some work in progress commits I'll make into a new PR that results in very close to what we want.  The resulting plane is just clipped by the bounds of the original dataset rather than the translated one.  And when I change what I think affects the clipping bounds on the mapper, it doesn't do anything.

I'll update the title on this one though.  If you are OK with merging without that one working feel free after you have tested this.

mathturtle:
Ok, I finally got the non-orthogonal slice translating right.  I'm closing the WIP MR for that and just adding the final code as a commit on this branch.

@Hovden @yijiang1 Here is a branch that lets you move an object in the scene when you have the data source (not a module) selected on the left.  Do you want to try it out before it is merged?

Hovden:
We will check it out today. @yijiang1 want to meetup?

yijiang1:
@Hovden Sure. I'm be school around 10am.

Hovden:
@mathturtle The program builds and runs. The UX interactivity is not clear.

[1] When we clicked the dataset it took us a while to figure out that you have to click the center ball and that the other 6 balls are not active. Also, because we had a visualization in place, we didn't even see the center ball.

[2] I'm not sure what plane the object is translating in. Is it always perpendicular to the viewing direction? If so, maybe clicking the balls on the sides of the bounding box will translate only in the plane of that bounding box. This would be one way to solve [1] and [2]. Of course this all gets more confusing when we add rotation.

cryos:
Yeah, once you find the right ball to click it is OK, but I think the pickable zones need to simply be the faces of the outline ideally, otherwise it should at a minimum be any one of the spheres. Translation versus rotation should be similar to view navigation using mouse buttons/modifier keys. Can you just make all of the spheres pickable and have them perform the same interaction?

mathturtle:
I agree that the UI needs work.
> Can you just make all of the spheres pickable and have them perform the same interaction?

I think I'll have to write a custom box widget to do this.  I tried turning the spheres off, but then you can't interact at all.  I have rotation turned off right now but if it is on clicking inside the box but not on the spheres causes the object to rotate about its center.  We should talk next week about what widget we want and how we want it to behave.

@Hovden Yes, the object is translating in the plane of the viewing direction.  If I write a custom widget, I can implement your suggestion about the balls on the edges.

cryos:
Sounds good, I think both behaviors would be best served by clicking inside the box, with no spheres. This is a good starting point, but it feels pretty frustrating to interact with at this stage.

mathturtle:
So after 6+ hours of digging to find out how to modify the widget... There is already a way to configure the widget the way we want.  Try this patch.  I haven't enabled rotation yet, but a left click and drag anywhere on the widget should translate it.  Note that the widget is currently only enabled when the dataset (not a module) is selected on the left in the outline widget.  That probably has to change too.

mathturtle:
Rebased on the new Qt5 work.  The widget now translates whenever the user clicks anywhere in the box, but is still only enabled when the Dataset is selected on the left.

mathturtle:
@cryos @Hovden @yijiang1 
This is a new version without the annoying ball handles.  You can move the object by clicking anywhere on its bounding box now.  Can you please test and give me feedback?

yijiang1:
LGTM.

Hovden:
If the code is good. Let's merge for wider testing. We need to move forward with this feature.

PR #327, opened 2015-12-18
Convert to float

### LINKED ISSUE #241: Change data type
We should provide a way for the user to change the datatype.

Primarily, a user may want to go from an unsigned integer to a signed float.

mathturtle:
This adds an operator to convert the dataset to float to fix #241.

@Hovden since I didn't hear back I went ahead and implemented this...

### DIFF
--- a/tomviz/Operator.h
+++ b/tomviz/Operator.h
@@ -53,6 +53,7 @@ public:
   virtual bool deserialize(const pugi::xml_node& ns)=0;
 
   virtual EditOperatorWidget* getEditorContents(QWidget* parent) = 0;
+  virtual bool hasCustomUI() const = 0;
 
 signals:
   /// fire this signal with the operation is updated/modified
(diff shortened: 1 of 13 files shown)


Hovden:
OK. I think this is a reasonable solution.

tomviz/Operator.h
```diff
@@ -53,6 +53,7 @@ class Operator : public QObject
   virtual bool deserialize(const pugi::xml_node& ns)=0;
 
   virtual EditOperatorWidget* getEditorContents(QWidget* parent) = 0;
+  virtual bool hasCustomUI() const = 0;
```
yijiang1: This makes tomviz unable to be built because it conflicts with #320 , in which the "TranslateAlignOperator.h" file does not override the new virtual function.

cryos: Yeah, this seems a little over the top - it should default to false, and let you override it if you have custom UI. I will submit a patch.

Issue #328, opened 2015-12-31
Fail to build after merging #327 and #320


yijiang1:
Tomviz can not be built after merging both #320 and #327 
See inline comment in #327.

PR #329, opened 2015-12-31
Operator api

cryos:


cryos:
Fixes #328, return false by default from the new method.

PR #331, opened 2016-01-04
Added missing initialization of member variables

### LINKED ISSUE #330: Unable to launch Tomviz
I'm having trouble launching tomviz, both the build version and the binary version (from dashboards).
Sometime I'm able to open tomviz but it crashes when I mouse over the menus.

@Hovden @ElliotPadgett Can you download the latest binary version and see if you have the same issue? It may be my own computer's fault (for example, my paraview is outdated?) If not, I guess the bug is probably introduced in #321 .

cryos:
The manager was crashing when onViewPropertyChanged was called as the
member variables were not initialized in the constructor to null. I
also made the null check slightly more defensive in that member function
to check for both variables that are used. This fixes #330.

### DIFF
--- a/tomviz/ViewMenuManager.cxx
+++ b/tomviz/ViewMenuManager.cxx
@@ -37,7 +37,8 @@ namespace tomviz
 {
 
 ViewMenuManager::ViewMenuManager(QMainWindow* mainWindow, QMenu* menu)
-  : pqViewMenuManager(mainWindow, menu)
+  : pqViewMenuManager(mainWindow, menu), perspectiveProjectionAction(nullptr),
+    orthographicProjectionAction(nullptr), showAxisGridAction(nullptr)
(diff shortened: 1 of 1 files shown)


mathturtle:
I did test it, but I guess I got lucky with the memory that was there already being set to 0.

cryos:
That is possible, RelWithDebInfo can be good for retaining debug symbols but looking more like a release build in terms of performance and initialization. No big deal, it was a simple fix once I dug into the code a little. It seems odd to rebuild the menu every time it is shown too, I was surprised when I saw that - did you double check if the QAction objects were parented properly and deleted? It gets rebuilt every time you open the menu.

cryos:
I am going to merge this as it is a pretty simple fix, and it corrects a fairly consistent crash here.

Issue #19, opened 2015-05-01
Retina display (Mac) not supported; looks pixelated


Hovden:
GUI and icons look pixelated due to lack of retna display support. This may be resolvable with QT5 implementation.

Hovden:
@cryos has mentioned a desire to start building with QT5, so I'm marking as 0.8.

mathturtle:
Can someone test this with the new nightly binaries?  We are on Qt 5.6 now, but I don't have a retina display to test with.  Is there some Qt config thing we need to enable or should this just work?  @Hovden @cryos @cquammen

cryos:
Moving to Qt 5.6 is the first step, we need 2x versions of all icon assets, this is for Tomviz and ParaView icon assets. I think there were some code updates, I can go back to what we did in another project, but until the 2x icons are available the code changes will not do much. Google Qt 5 retinal display - their docs are pretty good.

Hovden:
I think there is a Qt config that needs to be set. The jump to high DPI monitors should be noticeable on most of the application (text etc.) even if the icons are not at 2x.

http://doc.qt.io/qt-5/highdpi.html

Hovden:
Any luck with setting QT_AUTO_SCREEN_SCALE_FACTOR to "1" ?

cquammen:
I didn't have luck with `QT_AUTO_SCREEN_SCALE_FACTOR `, but did have some success augmenting the `Info.plist` file with the following keys and values:

    <key>NSPrincipalClass</key>
    <string>NSApplication</string>
    <key>NSHighResolutionCapable</key>
    <string>True</string>

One must force the OS to update its cache of the `Info.plist` file by renaming the tomviz application in the Finder to something else, then naming it back to tomviz. This had me stuck for a little while. The good news is that the high-resolution option makes the UI elements render crisply. The bad news is that the VTK rendering only occurs in the bottom left quadrant of the VTK widgets:

![image](https://cloud.githubusercontent.com/assets/360056/17004203/f8f52910-4ea1-11e6-8790-c7bddc560c05.png)

More work will need to be done to expand the rendering to the full window. I don't have a good feel for how long this will take.

Hovden:
@utkarshayachit @cryos Has anyone got this to work in Paraview?

cquammen:
Not fully. See, for example, http://www.paraview.org/Bug/view.php?id=15826

Hovden:
It looks like point 1 and 2 have been addressed.
[...]

PR #335, opened 2016-01-11
Qt5 porting

cryos:


cryos:
@mathturtle @cquammen here is the first pass at Qt 5 porting, let me know if you see anything major missing. This branch will require changes to the superbuild before it can be merged.

cryos:
One issue, that I have seen in other applications, is the progress dialog popping up after a few seconds. I just tried delaying the initialization until it is used which looks better. The dialog never seems to get rendered though, which is not all that useful...

mathturtle:
I fixed the parenting issue on the colormap widget.  At least it appears to be fixed by the commit I just added to your branch.

cryos:
Thanks @mathturtle, that looks better here too.

cquammen:
I am testing against Qt5 (5.5.1) from homebrew on Mac. Everything is great except I get this warning message 12 times right at startup.

`0x7feeac460480 void QWindowPrivate::setTopLevelScreen(QScreen *, bool) ( QScreen(0x7feea9e60200) ): Attempt to set a screen on a child window. (:0, )`

Googling around suggests this is a Qt bug, and I'm looking for a solution.

cquammen:
Note that the warning also occurs in the ParaView build on top of which I am building tomviz.

cquammen:
There is a 5.6 tag in the Qt git repository that I checked out and built on El Capitan. In-source builds do not appear to work (QHelp was nowhere to be found). An out-of-source build does appear to work find. I built ParaView 5.0 and tomviz with it successfully.

The `QWindowPrivate::setTopLevelScreen` warning is gone, but the pixelization issue on Retina displays (#19) remains.

cryos:
Sounds good, we need to get retinal resolution pixmaps so that is expected.

cquammen:
Actually, it is rendered text in Qt widgets that are the most obviously aliased.

cryos:
We probably need to tweak the QApplication, I know there are some initialization parameters that I tweaked for Avogadro 2. We can look at this more next week.

cquammen:
Note for superbuild on Mac: if you *don't* use system Qt and you point to the qt5.git repo, then the build fails with

[ 62%] Performing update step for 'qt'
Current branch 5.6 is up to date.
No stash found.
No stash found.
No stash found.
CMake Error at /Users/cory.quammen/bin/tomviz-superbuild/qt/tmp/qt-gitupdate.cmake:136 (message):
  

  Failed to unstash changes in:
  '/Users/cory.quammen/bin/tomviz-superbuild/qt/src/qt/'.

  You will have to resolve the conflicts manually

There are a number of project files generated by Qt's `configure` script that prevent the git update step from working. These files make the submodules appear to be modified, causing the qt update script to attempt to stash the changes, but the files are not in the repository, so no stashing occurs. When it comes time to unstash the changes, unstashing fails because no stash was created. This causes the update script `qt/tmp/qt-gitupdate.cmake` to fail and exit.

mathturtle:
This is (I think) why we use Qt's source tarballs/zips instead of pointing at their repo.  But I would also add that the superbuilds have never really supported reusing a build tree.  For now I would build your own "system" Qt from Qt's repo and point the superbuild at that.

cquammen:
Yep, trying that out now.

cryos:
Qt doesn't support reusing their build trees well at all, you should just use homebrew's Qt, or download the official tarball, verifying the bug goes away with their tag is of course useful. We won't package from your machine anyway, so don't worry too much about building it there.

cryos:
We are planning on merging this shortly, @mathturtle will get the build bots ready, now that 0.7.2 is tagged we can start merging some of the larger changes.

PR #336, opened 2016-01-19
Automatically add the ITK Python wrapping to build tree Python path.

thewtex:
This was only tested locally on Linux, but it is also implemented for Mac OSX and Windows.

CC: @cryos @mathturtle

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -57,4 +57,11 @@ find_package(Qt4 REQUIRED)
 include_directories(SYSTEM "${QT_INCLUDES}")
 add_definitions(${QT_DEFINITIONS})
 
+find_package(ITK 4.9)
+if(ITK-FOUND AND NOT ITK_WRAP_PYTHON)
+  message(FATAL_ERROR
+    "Tomviz requires ITK_WRAP_PYTHON to be enabled. "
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -57,4 +57,11 @@ find_package(Qt4 REQUIRED)
 include_directories(SYSTEM "${QT_INCLUDES}")
 add_definitions(${QT_DEFINITIONS})
 
+find_package(ITK 4.9)
+if(ITK-FOUND AND NOT ITK_WRAP_PYTHON)
```
mathturtle: Did you mean an underscore in ITK_FOUND?

mathturtle: Also, can we test for the FFTW support as well?

thewtex: > Did you mean an underscore in ITK_FOUND?

Yes, thanks for checking that. I will add it to a follow-up.

> Also, can we test for the FFTW support as well?

Yes, ITK_USE_FFTWF and ITK_USE_FFTWD (for float and double) settings are exported by ITKConfig.cmake.  Are both needed?

cryos:
Let's queue this branch up, and merge after the next release too.

cryos:
I fixed this in #353, closing as it this has now been merged after rebase and some minor changes. Thanks @thewtex

PR #337, opened 2016-01-19
Remove ITK-4.8 Python path directory.

thewtex:
For ITK 4.9.0, this path addition should not be required because ITK now uses
a more standard installation location.

This has removed all ITK-4.8 references in the tomviz repository (as verified
with git grep).

This has *not been tested locally*.

### DIFF
--- a/tomviz/tomvizPythonConfig.h.in
+++ b/tomviz/tomvizPythonConfig.h.in
@@ -67,7 +67,6 @@ void PythonAppInitPrependPathWindows(const std::string& exe_dir)
 
     // we don't add anything for ParaView since, ParaView takes care of that.
     PythonAppInitPrependPythonPath(exe_dir + "/../lib/itk/");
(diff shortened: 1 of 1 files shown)


thewtex:
CC: @cryos @mathturtle

mathturtle:
@thewtex What is the 'more standard installation location'?  I'll have to update the superbuild to use ITK 4.9 and build new binaries for it to use.

@cryos Do we want to go ahead and update to ITK 4.9?

thewtex:
@mathturtle instead of `site-packages/itk.py` and the actual modules located outside of `site-packages`, it is now `site-packages/itk/__init__.py`, and the modules are in `site-packages/itk/`.

ITK 4.9.0 will be tagged over the next week or so, so it would be nice to upgrade then.

cryos:
We will upgrade after we make this next small release, which should happen in the next 4-5 days.

cryos:
@mathturtle Let's get this merged once the new ITK version is ready to go for the superbuild

PR #338, opened 2016-01-21
WIP: Translate non ortho slice

mathturtle:
@cryos This is as close as I have gotten to getting the non-ortho-slice to translate correctly.  It is somehow still cropped by the original bounds of the data though, and changing the clip_bounds variable doesn't seem to help.


mathturtle:
Wrapping into #326.  This MR was only to share WIP commits while allowing the other to be merged if desired.

PR #340, opened 2016-01-30
Color by labelmap

cquammen:
This branch introduces support for creating and rudimentarily visualizing label maps in tomviz. A label map is a special image in which each voxel has a label that has a special meaning. Usually, the label designates a voxel as belonging to part of a structure captured in the image.

Implementation-wise, a label map is stored as an additional point array in the vtkImageData in the DataSource output. This can be accomplished on the Python side with

`utils.set_label_map(dataset, result)`

where `result` is the array of label map values.

This PR also introduces a new Visualization Module, the Label Map Contour. Like the Contour module, this module enables the creation of a contour through the image data. It adds the capability of coloring by a label map in the image data if one is present.


cquammen:
@cryos I am happy with the ITK examples in this branch. Do you want to give it another try? The UI is the same as last time you tried it.

cryos:
Will do, thanks @cquammen

cryos:
Maybe I am forgetting something, but I do not see the colors for the sizes when trying this branch.

cquammen:
* Open LARGE_PtCu_NanoParticles.tiff
* Connected components with default arguments
* Create Label Map Contour visualization module
* Check "Use separate color map for this module"
* Pick a more colorful color map from the favorites (I like Rainbow Desaturated).
* You could change the contour value to 50 to see more blobs. You should see something like this:

image

Hovden:
Looks cool. Giving a tomviz talk in an hour. Last talk I got requests for segmentation.

cryos:
It is coming, just testing this out again now.

cryos:
I see what I missed - there is an exception because you used VTKImageToImageFilter, I guess I need to turn on another option in ITK and rebuild. Then I need to document our minimal set/ensure we have this in the superbuild. I see where you made the change now.

cryos:
It will make the build instructions a little more complex too as we need ITK to link to ParaView's VTK, but I can add that in if we need the ITKVtkGlue module.

cquammen:
How about that exception reporting, eh?

Yes, one needs to enable Module_ITKVtkGlue and point it to the VTK directory in the ParaView build directory. I can update the docs if you would like me to.

Another thing that needs to be defined is which pixel types we decide to wrap. I have enabled:

- ITK_WRAP_VECTOR_COMPONENTS  3
- ITK_WRAP_float ON
- ITK_WRAP_rgb_unsigned_char ON
- ITK_WRAP_rgba_unsigned_char ON
- ITK_WRAP_signed_char ON
- ITK_WRAP_signed_short ON
- ITK_WRAP_unsigned_char ON
- ITK_WRAP_unsigned_long ON
- ITK_WRAP_unsigned_short ON
- ITK_WRAP_vector_float ON

The dispatch from VTK image type to ITK image type could be improved to handle the ITK wrapped types that are available.

cryos:
It is easy to miss as I have other noise from the script. Do we need all of those types? The tomviz vtkImageData will be unsigned char, unsigned short, or float to be best of my knowledge. I would rather just build the types we are likely to use, and worked with Matt on minimizing that list.

cryos:
In addition we would want the likely output types for label maps, if that is not unsigned char, or unsigned short.

cryos:
Full output before rebuild (not I still have the warnings I mentioned):

Warning: Unknown parameter 'itk::FlatStructuringElement< 2 >' in template 'itk::BlackTopHatImageFilter'
Warning: Unknown parameter 'itk::FlatStructuringElement< 3 >' in template 'itk::BlackTopHatImageFilter'
Warning: Unknown parameter 'itk::FlatStructuringElement< 2 >' in template 'itk::ClosingByReconstructionImageFilter'
Warning: Unknown parameter 'itk::FlatStructuringElement< 3 >' in template 'itk::ClosingByReconstructionImageFilter'
Exception encountered while running ConnectedComponents
'LazyITKModule' object has no attribute 'VTKImageToImageFilter'

cquammen:
I agree we should minimize the type list. That will make ITK compile faster. I was just letting you know what I built with to potentially avoid a rebuild later on because a type was missing.

cryos:
I think the pain is worth it, works great with the addition of the glue module.

cquammen:
BTW, where are the ITK build instructions/configuration settings specified? Is this (https://github.com/OpenChemistry/tomviz-superbuild/blob/master/Generating_ITK_binaries.md) the place?

cryos:
That is for the superbuild, then there is the developer instructions page. Talking with Shawn, we may want to consider avoiding the glue module for now. I know things will be a little slower, but the build complexity goes up in the superbuild as ITK becomes coupled to VTK's ABI, and so it forces a rebuild every time ParaView is updated - that is not what we are doing right now.

If you have abstracted it away, is there any chance we can just do the copy for now, and then revisit this later? I would like to get these early results into people's hands soon. If it will take too long we can delay until the superbuild is modified to rebuild every time, and I have already asked them to enable incremental builds for tomviz builds (but that needs some testing to verify it works as expected) to avoid rebuilding all dependencies for every change.

cquammen:
Yeah, it's possible to change the implementation to avoid using the VTK glue. It will just double the connected components filter time. For the large image, creating the ITK image via numpy array took 12 seconds, just as long as the connected components filter itself. It may be possible to speed this up. I'll check with Matt.

cryos:
I think we can accept the slow down for now, and get it into people's hands. We can optimize/use the glue later, once we have some other bits of the superbuild working.

cryos:
Thanks Cory, once we have working binary builds for Mac and Windows we can merge this, @mathturtle is working on getting fixup bundle working for the Mac packages right now.

cquammen:
Two last things for this branch:

- I can add UI to set the binary threshold filter parameters
- I can add a checkbox for the regular contour filter to color by label map instead of image value. With that in place, I would remove the second contour icon in the toolbar.

Shall I do either of these?

cryos:
I am going to merge this now, and see how things work in the packaged binary. We can absolutely follow up with more commits in the future, sorry about the delays in merging. They were to allow us to finish patching tomviz, ParaView and the superbuild to package against Qt 5.

PR #349, opened 2016-02-22
Updates to some of our layouts for Qt 5

cryos:
This is to improve the use of space in the application, increase the
default size of the right panel, and will likely need more tweaking.

### DIFF
--- a/tomviz/CentralWidget.ui
+++ b/tomviz/CentralWidget.ui
@@ -14,6 +14,18 @@
    <string>Form</string>
   </property>
   <layout class="QGridLayout" name="gridLayout">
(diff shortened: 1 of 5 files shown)


Hovden:
Great.

Also, not sure where to make this comment, but Cdash builds for Mac haven't been running.

mathturtle:
> Also, not sure where to make this comment, but Cdash builds for Mac haven't been running.

They've been running but there is some issue actually submitting them to CDash.  I'm looking into it.

cryos:
Going to merge this as it would be good to get wider feedback, and we are not done tweaking but I would like to see a new build attempt.

PR #352, opened 2016-02-26
Add mode for moving objects in scene

mathturtle:
@Hovden @cryos 

This adds the mode for enabling moving objects.  When the new mode is disabled (default) no moving of objects is allowed and the widget doesn't show up.  When it is enabled, the widget shows up and works as in the previous patch.  It probably needs more work (and clearly shows why I don't make icons, although I could do better with an image editor that I understood better than gimp).  I've also thought of a modification that might get rid of the flicker of the widget when loading large data, but I'll need to refactor this a bit.  Pushing this since it works and you can give feedback.


mathturtle:
Updated like I described last night.  We still need a better icon, but other than that this is about ready.

PR #353, opened 2016-03-15
Itk python location

cryos:


cryos:
This updates #336 to address conflicts, and moves logic.

PR #355, opened 2016-03-23
Save state of axes grid

mathturtle:
The axes grid updates a few months ago broke the save/load state file.
We needed to save the state of the axes grid object so that ParaView
doesn't use a NULL one when loading a state.

All I can say is that apparently noone uses the save/load state feature.
@cryos

### DIFF
--- a/tomviz/ModuleManager.cxx
+++ b/tomviz/ModuleManager.cxx
@@ -29,6 +29,7 @@
 #include "vtkPVXMLParser.h"
 #include "vtkPVXMLElement.h"
 #include "vtkSmartPointer.h"
+#include "vtkSMPropertyHelper.h"
 #include "vtkSMProxyIterator.h"
 #include "vtkSMProxyLocator.h"
 #include "vtkSMProxyManager.h"
@@ -319,6 +320,17 @@ bool ModuleManager::serialize(pugi::xml_node& ns, const QDir& saveDir) const
         qWarning("Failed to serialize view.");
         ns.remove_child(vnode);
       }
+
+      vtkSMProxy* axesGrid = vtkSMPropertyHelper(view, "AxesGrid").GetAsProxy();
+      pugi::xml_node axesGridNode = vnode.append_child("AxesGrid");
+      axesGridNode.append_attribute("id").set_value(axesGrid->GetGlobalIDAsString());
+      axesGridNode.append_attribute("xmlgroup").set_value(axesGrid->GetXMLGroup());
+      axesGridNode.append_attribute("xmlname").set_value(axesGrid->GetXMLName());
+      if (!tomviz::serialize(axesGrid, axesGridNode))
+      {
+        qWarning("Failed to serialize axes grid");
+        ns.remove_child(axesGridNode);
+      }
     }
   }
   return true;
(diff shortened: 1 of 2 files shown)


mathturtle:
I'm surprised that noone found that loading a state file crashes tomviz between when #321 was merged and now.

cryos:
@mathturtle I am glad you spotted this regression before our next release, it is something as you point out was introduced a while back by one of your previous contributions. This functionality is extremely important, and when adding new code you should consider how you can ensure it doesn't crash even when things are not in the state you expect, and again it is great to see this patch take steps to address that in a follow up patch.

We need to find some time to add some automated testing to assure that functionality is verified in a more automated fashion, and this is something I have been aware of for some time. When you are adding functionality you need to consider the wider impacts of your changes, and write defensive code to avoid possible use of null pointers, etc. This is something I look out for when reviewing code, but missed on this occasion. We will miss things at times, such as this, thanks for following up on this issue and submitting a fix.

mathturtle:
I normally check for null pointers too, but I was expecting that if the view proxy was valid its subproxy for the Axes Grid would also be non-null.  Apparently that isn't the case when loading from a file if you didn't save out the Axes Grid...

cryos:
If you get a pointer from a function you should check it is valid before dereferencing it, APIs can change, and it is far more preferable to fail gracefully than crash. When we think something is guaranteed an assert can be used, but in cases like these it is far better to check and fail gracefully as an assert has the same behavior as far as an end user is concerned.

tomviz/ModuleManager.cxx
```diff
@@ -345,6 +357,9 @@ bool ModuleManager::deserialize(const pugi::xml_node& ns, const QDir& stateDir)
   int numViews = 0, numLayouts = 0;
   for (pugi::xml_node node = ns.child("View"); node; node = node.next_sibling("View"))
   {
+    pugi::xml_node agnode = node.child("AxesGrid");
```
cryos: These variables are not very symmetric with those above, and the capitalization is off. I would suggest using axesGridNode as above so the relation is clear, and the meaning easier to see at a glance.

tomviz/ModuleManager.cxx
```diff
@@ -345,6 +357,9 @@ bool ModuleManager::deserialize(const pugi::xml_node& ns, const QDir& stateDir)
...
+    pugi::xml_node agproxyNode = agnode.child("Proxy");
```
cryos: As above, and Proxy should be capitalized, i.e. axesGridProxyNode, axesGridNode, etc.

tomviz/ModuleManager.cxx
```diff
@@ -355,6 +370,16 @@ bool ModuleManager::deserialize(const pugi::xml_node& ns, const QDir& stateDir)
       viewSummary.append_attribute("name").set_value(
           QString("View%1").arg(++numViews).toStdString().c_str());
     }
+    if (agproxyNode)
```
cryos: As these are scoped, and local meaning is pretty obvious, if it isn't used simply proxyNode would work fine too (and be shorter). Either way, I think this should be cleaned up and the variable names between reading and writing kept consistent.

cryos:
Works as expected now, see comments on variable naming. Once they are consistent this can be merged.

mathturtle:
Updated variable names... merging.

PR #357, opened 2016-03-25
Update move objects icon

mathturtle:
Here is the new move objects button icon that Steve and I came up with.  This PR changes the icon used in the menu bar.

hand_box

@cryos


mathturtle:
The last part to fix #38.

cryos:
Looks good, although the dimensions are off. Can we get one matching the standard toolbar icon size please (that is still 64x64 as far as I remember - definitely square).

mathturtle:
Ok, new 64x64 icon in updated branch.

cryos:
Much better, and we can see how the builds do...

PR #360, opened 2016-04-01
Move objects: ignore whether a datasource is selected

mathturtle:
Now that this is a mode and not always on, use the mode enabled rather
than whether a datasource is selected to determine whether to show the
widget and enable moving objects.

Should be the last part of #38.


cryos:
Looks good, not tested it yet, but will take a better look when in master. Thanks @mathturtle

PR #363, opened 2016-04-04
Histogram margins

cquammen:
These are some minor changes to tweak the margins and size of editor widgets.

New
image

vs. previous

image


cquammen:
@Hovden How do you like the smaller margins and thinner color bar?

Hovden:
Nice!

cryos:
This looks good to me too, one thing I think we need to look at is the ticks and how they cause a shift in the start of the color bar and the x axis. I think there are some features in the charts I added to line up charts in different ways, and they may help here to ensure the color bar is parallel with the start of the x range.

cquammen:
@cryos Ha ha, was just requesting review when you merged...

PR #365, opened 2016-04-04
Style guide

cryos:


cryos:
I think this is good for now @cquammen, we can expand it out further in future, thanks for putting it together.

PR #366, opened 2016-04-04
Restore contour setting by double clicking in histogram chart

cquammen:
This commit restores the setting of a contour value by double
left-clicking in the histogram view. It does so by adding new opacity
function control points only when the mouse is clicked very near or on
the piecewise function. Prior to this commit, new points would be
added to the opacity function when the mouse was clicked anywhere in
the histogram.

Also, the behavior for middle- and right-clicking has been changed so
that nothing happens if these buttons are clicked. Without changing
their behavior, these buttons would select the control points, which
isn't useful and can cause some confusing things to happen when trying
to edit a single opacity function control point.


cquammen:
@cryos This should remedy the apparent disappearance of setting the contour value in the histogram view.

cryos:
Looks good to me.

PR #372, opened 2016-04-17
Update linux binary default python path

mathturtle:
When we switched to installing paraview without the version number in
the path to its libraries, it broke the logic for where to find the
paraview python packages.  This updates this logic for the new locations
as well as updating the logic to find ITK and numpy/scipy for the linux
binaries.

@cryos


cryos:
LGTM

PR #374, opened 2016-04-19
Add ITK python install dir to python path on osx

### LINKED ISSUE #373: Generating Sample Data Crashes Again
Tomviz is back to crashing when a user runs generated sample data.

Repeatable on OSX by running "Sample Data" > "Random Particles".

mathturtle:
This fixes #373.  It feels a bit hacky, really we should make ITK install its python libraries to tomviz.app/Contents/Python.  But this works and was quick to implement.


mathturtle:
@cryos I fixed this on Linux last week, but I'm not sure it ever really worked there (X errors prevented testing of python).  But now since it broke on OSX, I decided to investigate Windows too.  I've updated this to contain a commit that will fix the windows binaries.

There were two things that caused this:  The patch that was merged to update the paths to ITK for 4.9 didn't work and changing the superbuild's paraview not to use its version in the install directories seems to have broken the built-in logic to find the paraview python libraries.  I didn't test these branches thoroughly enough :(.

Issue #369, opened 2016-04-14
UI for reconstruction progress


yijiang1:
tomviz needs UI for tomography reconstruction, which often lasts hours for real data.
Here is simple mock-up.
<img width="879" alt="screen shot 2016-04-14 at 10 08 08 am" src="https://cloud.githubusercontent.com/assets/13088588/14530855/d32836fc-0228-11e6-9feb-5757e3586481.png">

The components of this UI is very similar to the ones in manual axis alignment UI (RotateAlignWidget.cxx). As a first step, the UI should loop though each x position in tilt series, extract the slice (y-z plane) using "TomographyTiltSeries::getSinogram" function, and reconstruct it using "TomographyReconstruction::unweightedBackProjection2".

PR #375, opened 2016-04-21
Reconstruction ui

mathturtle:
@yijiang1 I've made the UI you talked about in #369.  It still needs a bit of work, but you can try it out if you want.  (The Simple Back Projection (C++) menu item under the Tomography menu pulls it up).

### DIFF
--- a/tomviz/CMakeLists.txt
+++ b/tomviz/CMakeLists.txt
@@ -133,6 +133,8 @@ set(SOURCES
   PipelineWidget.h
   ProgressBehavior.cxx
   ProgressBehavior.h
+  ProgressDialogManager.cxx
+  ProgressDialogManager.h
   PythonGeneratedDatasetReaction.cxx
   PythonGeneratedDatasetReaction.h
   RecentFilesMenu.cxx
--- a/tomviz/CMakeLists.txt
+++ b/tomviz/CMakeLists.txt
@@ -139,6 +139,8 @@ set(SOURCES
   RecentFilesMenu.h
   ReconstructionReaction.h
   ReconstructionReaction.cxx
+  ReconstructionWidget.h
+  ReconstructionWidget.cxx
   ResetReaction.cxx
   ResetReaction.h
   RotateAlignWidget.cxx
@@ -249,6 +251,7 @@ qt5_wrap_ui(UI_SOURCES
   MainWindow.ui
   ModulePropertiesPanel.ui
   RotateAlignWidget.ui
+  ReconstructionWidget.ui
   ViewPropertiesPanel.ui
   )
 qt5_add_resources(RCC_SOURCES resources.qrc)
--- a/tomviz/CMakeLists.txt
+++ b/tomviz/CMakeLists.txt
@@ -139,8 +139,10 @@ set(SOURCES
   PythonGeneratedDatasetReaction.h
   RecentFilesMenu.cxx
   RecentFilesMenu.h
-  ReconstructionReaction.h
+  ReconstructionOperator.cxx
+  ReconstructionOperator.h
   ReconstructionReaction.cxx
+  ReconstructionReaction.h
   ReconstructionWidget.h
   ReconstructionWidget.cxx
   ResetReaction.cxx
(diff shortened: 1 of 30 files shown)


mathturtle:
Ok, this is rebased and (mostly) cleaned up now.  I had to change the UI a bit since it is just a progress dialog now.  Essentially progress dialogs don't need a start button, and since canceling an operator halfway could leave the DataSource in a weird state I removed the cancel button for now too.

New summary:  This branch adds progress dialogs for running operators.  By default this is just a dialog with a QProgressBar in it, but it can be overridden if an operator has something else it wants to use to display progress.  It also converts the C++ reconstruction code to an operator that uses this functionality.  I also fixed a few bugs I found along the way (a null guard and a memory leak).

@cryos

mathturtle:
Rebased on master so I don't have to keep rebuilding ParaView when I switch branches.

mathturtle:
Fixed another case of blank dialog that didn't paint by making operators emit a dummy "no progress" event before starting.

yijiang1:
Two dialogs boxes are appearing.
Images can still be rotated...
Clicking "OK" button create additional reconstructions.
 
screen shot 2016-05-04 at 3 27 42 pm

mathturtle:
Hmmm, you're right.  Since there are no options to set on the reconstruction, the first dialog is a bit redundant.  (It is the same dialog as editing the other operators, but there are no options on the reconstruction so that part of it is empty)  I'll make it skip that dialog.  Maybe a warning that the reconstuction can be long running?

Hovden:
A single dialogue would be good.

The simplest option would be to have only a cancel button in that dialogue. The a better option would be to also have a button that allows uses to pause / unpause.

mathturtle:
> The a better option would be to also have a button that allows uses to pause / unpause.

I'm curious why you would want to pause.  If you do anything in tomviz that causes the dataset to change, you have to restart the reconstruction anyway.  Is this so you can do something else on your computer that needs a lot of CPU power?

Hovden:
Yeah, reconstruction eats up the CPU resources. In practice most people will not be doing other things while reconstructing a dataset so the pause functionality is not critical. A cancel button is important since you often realize the reconstruction will turn out poorly once it's started.

mathturtle:
So you want to cancel based on the reconstructed slices that the progress dialog shows?

Canceling mid-operation is a more interesting use case.  We need to decide how to handle it.  Do you want it to remove the reconstruction data transform (i.e. not add it at all) or mark it as incomplete/failed somehow?  What about when the user edits something above it in the list of data transforms and it re-executes.  Is canceling it then different?

Hovden:
Cancel button will abort the reconstruction, discard the reconstructed data. It will be in the same state as before the reconstruction operation was clicked.

cryos:
Agreed, I think this is the way any cancel should work. Restore state, remove the operation that was in progress.

mathturtle:
Ok, I just got the Cancel button working as discussed.  It was more complex than I expected since no one part of the system has the information to listen to the Cancel button on the progress dialog and delete the in progress Operator from its DataSource.  Probably some refactoring could be done here, but I haven't come up with anything that works well yet.  So some of this feels a bit hacky.

And I added a commit to disable the interactions that Yi mentioned.

Hovden:
@yijiang1 Does this look OK to you?

yijiang1:
Overall looks good to me. 
Right now the reconstruction starts automatically. Can we have a start button? In general we should have some descriptions about the algorithm and perhaps some parameters for users to specify.

Hovden:
I'm ok with it starting right away if there are no parameters to set for this one. If we add user-specified parameters we can add a start button.

mathturtle:
> Can we have a start button?

The problem with having a start button is that the dialog is no longer purely a progress dialog.  You didn't like the two dialogs with the OK/Apply buttons to start the operation... I suppose we could combine the progress dialog and the settings dialog for all the operators.  That could get messy, but I think it is possible (I'd really rather not do this though).

yijiang1:
> I suppose we could combine the progress dialog and the settings dialog for all the operators.

I think this is best for user experience. We want users to specify parameters (for more advanced reconstruction methods) and view reconstruction progress in one dialog.

Hovden:
Here, we can merge this as a back-projection reconstruction that runs automatically.

Let's discuss the start / apply / pause etc as another issue. We can talk about the tradeoffs between user experience and implementation at the next meeting.

yijiang1:
Agreed

PR #379, opened 2016-04-28
Save non-python operators correctly in state file

### LINKED ISSUE #378: Crop transform not saved in state file
If you crop a data set and then save a state file, the crop transform is not reapplied when loading the state file. A `transform_scalars()` transform is instead placed in the Data transforms flow.

mathturtle:
I added an OperatorFactory similar to the ModuleFactory so that the
deserialize code can ask for an operator of a type from the file rather
than hardcoding a python operator.  Tested will all types of operators.

Fixes #378.

@cryos This was something I overlooked when I first implemented non-python operators.  Adding new state files to my list of 'make sure this loads right' states.

### DIFF
--- a/tomviz/CMakeLists.txt
+++ b/tomviz/CMakeLists.txt
@@ -123,6 +123,8 @@ set(SOURCES
   MoveActiveObject.h
   Operator.cxx
   Operator.h
+  OperatorFactory.cxx
+  OperatorFactory.h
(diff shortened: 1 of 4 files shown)


mathturtle:
@cryos ping?  This sort of conflicts with my reconstruction ui/operator branch.  Can it get merged soon?

PR #390, opened 2016-05-05
Rename some menu entries

yijiang1:
Rename some menu entries so they are easier to understand.


cryos:
Please give us the opportunity to review changes before they are merged

PR #395, opened 2016-05-06
Remove the cube axes visibility property

cquammen:
ParaView removed the cube axes, so this property no longer exists
and an error is thrown whenever the state is saved. Removing
this property where it is used solves the problem.


cquammen:
@cryos Axes grid appears to work just fine as a substitute, and it is saving to and being restored from a state file correctly.

PR #397, opened 2016-05-09
Delay the update of the data properties panel

cryos:
There is a strange interaction between Qt 5 and QVTKWidget-based classes
in docks, such as the color transfer function editor. If no one ever
looks at this panel it is nice to avoid doing all this work, and by
doing so we delay initialization until the widget is visible.

This is a pretty hackish solution to that bug, and it needs more testing
on other platforms to see if it helps everywhere. We also need to do
more work to ensure the VTK-based OpenGL widgets always initialize their
context successfully, but we can do that later.

### DIFF
--- a/tomviz/DataPropertiesPanel.cxx
+++ b/tomviz/DataPropertiesPanel.cxx
@@ -121,6 +121,12 @@ DataPropertiesPanel::~DataPropertiesPanel()
 {
 }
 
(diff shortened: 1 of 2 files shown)


cryos:
This is a relatively small patch, and it helps (on Linux at least) with the color transfer function editor showing up in the data properties panel. I am going to merge this so it triggers builds of new installers.

PR #398, opened 2016-05-11
Use QStackedWidget instead of manual code.

utkarshayachit:
QStackedWidget is designed exactly for such use-cases. Let's use it.
Addresses the duplicated widgets seen on OsX.

Scrolling still causes issues, I suspect, due to the QVTKWidget.

### DIFF
--- a/tomviz/MainWindow.ui
+++ b/tomviz/MainWindow.ui
@@ -27,7 +27,7 @@
      <x>0</x>
      <y>0</y>
      <width>1280</width>
-     <height>29</height>
+     <height>22</height>
     </rect>
    </property>
    <widget class="QMenu" name="menu_File">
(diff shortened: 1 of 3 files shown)


tomviz/MainWindow.ui
```diff
@@ -184,7 +184,7 @@
         <bool>true</bool>
        </property>
        <property name="columnCount">
-        <number>2</number>
+        <number>1</number>
```
mathturtle: Qt Designer updates this automatically to what it thinks is right, but this being 1 causes crashes when interacting with the OperatorsWidget.  Can you set it back to 2?

mathturtle:
Other than that this looks good.  Thanks for pointing out the reimplementing the library thing.

PR #406, opened 2016-05-23
Memory efficient downsample by 2

Hovden:
Current downsampling routine uses 5x memory during execution.
With this pull request it will use 2x memory during execution.

@yijiang1


yijiang1:
It works.

Issue #60, opened 2015-05-31
Manual alignment: difference map comparison mode


ElliotPadgett:
In addition to "flashing" comparison mode where the user minimizes the wobble as the display flashes back and forth between images, it would be useful to have a "difference map" comparison, where the user makes the display of (imageA-imageB) flat around the alignment point.

I'm not sure what the best way is to make this fit with the current color map scheme, since it will need to display positive and negative values with magnitude significantly less than the typical data values.  Maybe the difference map will need to be scaled to fit in the same range as the original data?

PR #409, opened 2016-05-25
Align widget display modes

mathturtle:
Taking a stab at #60.  @cryos How is this?

### DIFF
--- a/tomviz/AlignWidget.h
+++ b/tomviz/AlignWidget.h
@@ -23,6 +23,7 @@
 
 #include <QVector>
 #include <QPointer>
+#include <QScopedPointer>
 
 class QLabel;
 class QSpinBox;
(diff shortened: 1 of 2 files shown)


mathturtle:
For waffle.io: resolves #60.

Hovden:
Can we merge?

tomviz/AlignWidget.h
```diff
@@ -77,34 +78,10 @@ protected slots:
 
   void sliceOffsetEdited(int slice, int offsetComponent);
 
-protected:
```
cryos: I would rather we avoided moving everything to an internals class, we are not worried about inheritance/minimizing header includes. I would like to stick with normal members unless there is a good reason to switch.

cryos:
We can merge, I would like to stop moving everything to an internals class unless we have a compelling reason, but other than that this patch looks fine at first glance (I haven't really gone through the view modes, or tested out the UI improvements, but we can iterate on it once merged too).

mathturtle:
Updated.  I tested the new version and it still works for me.

cryos:
Thanks Shawn

PR #410, opened 2016-05-26
Shift crop volume selection vtk widget when data source moved

### LINKED ISSUE #407: Interactive Crop Box appear at incorrect position when viz object has been shifted
1) open a dataset, and shift the visualization using the "hand on box" icon.
2) crop dataset

The crop box will appear where the dataset originally was placed, not where it is has been shifted to.

Also applies to Clear Volume Box

mathturtle:
@Hovden Fixes #407.


yijiang1:
Tested, works well

Issue #414, opened 2016-05-27
UI Crashes when clicking between visualization modules.


Hovden:
Crashes when clicking between visualization modules (OSX):
1) open two datasets, such as sample data reconstruction.
2) click quickly between visualization modules and datasets in the pipeline

mathturtle:
This doesn't appear to have anything to do with the visualization modules... It is clicking on the dataset that is crashing tomviz for me.  This is another OSX + Qt5 issue (or so it appears).  When the data properties panel is loading, I get a lot of messages like this:
```
QWidget::repaint: Recursive repaint detected (:0, )
```
Each time I switch between the module properties panel being shown and the data properties panel being shown (especially if I'm selecting a dataset that is not the last one shown) I get a longer wait while it figures out the repaint.  Given that the trace in lldb is >25000 frames, I'm guessing this is a stack overflow.

mathturtle:
So I actually managed to track this down a bit more today (Well, I found the DataPropertiesPanel was slow to load and did some debugging to find out why, then remembered this issue).  Anyway, the problem is in DataPropertiesPanel where we override paintEvent to call our update() method before repainting.  But update() modifies the UI, which then triggers another round of repaints and another paintEvent call.  @cryos said he was going to fix this as part of his branch to remove the colormap from the data properties panel.

cryos:
Hopefully this is fixed, I didn't look at this particular issue directly, but I did rename the slot, made it a protected member, and removed the render slot that wasn't needed. I think update() was badly named, let me know if things look better the loop should hopefully be gone now.

mathturtle:
I just tried and couldn't reproduce this bug.  It seems fixed to me.  @Hovden?

...I did find a new bug, but I'll create a new issue for that.

mathturtle:
@Hovden ping?  Can we close this?

Hovden:
Looks ok. Can't break it.

PR #423, opened 2016-06-14
Color ui reorg

cryos:


cryos:
Hey @mathturtle can you take a quick look at this when you have a chance, and check I haven't missed anything obvious. I will add some of the other buttons, and need to emit a signal for re-renders when switching color maps. I think I also need to remove the color stuff for the discrete module color maps too.

mathturtle:
The only issues I found you appear to already know about:
- 'Use a separate colormap for this module' still uses the old ParaView UI we are trying to remove in the module properties panel.
- It doesn't re-render when I change the preset.

mathturtle:
Tested on OSX too, didn't see anything.

mathturtle:
You'll probably want to revert e89e3004cfba70b59e6e97cd9fded45e94473f5e while you are at it.  Since we no longer have the QVTKWidget-based class in the properties panel (which that commit attempted to fix) and that commit is the direct cause of the recursive repaint bug (#414).

cryos:
I would still rather delay initialization until the panel is actually shown.

cryos:
I think this is ready, merging and heading out. I will check the builds later.

PR #425, opened 2016-06-15
Check that the AxesGrid property isn't null

cryos:
This was causing crashes when the view was not the standard 3D view.


cryos:
@mathturtle this fixes the crash caused by the unguarded use of the proxy, but it now outputs "Failed to locate property: AxesGrid" every time you switch views. Is there an elegant way of doing this by asking if the property exists first? I will probably just merge this to fix the crash...

mathturtle:
> Is there an elegant way of doing this by asking if the property exists first?

I'm not sure.  It sounds like something you should be able to do.  I'll look into it.

PR #427, opened 2016-06-15
Check axes grid exists

mathturtle:
Follow up on #425, removes "You asked for a property that doesn't exist on this view" warnings.

@cryos.


cryos:
Looks good, thanks Shawn.

PR #429, opened 2016-06-16
Whitelist the views we want, show less buttons

cryos:
This change adds some of the custom application support from ParaView
(thanks to Utkarsh) in order to simplify the user interface. We should
add more views, and possibly more buttons, but this shows a simple
starting point and we can test it more widely.

### DIFF
--- a/tomviz/ViewFrameActions.cxx
+++ b/tomviz/ViewFrameActions.cxx
@@ -1,0 +1,54 @@
+/******************************************************************************
+
+  This source file is part of the tomviz project.
+
+  Copyright Kitware, Inc.
+
+  This source code is released under the New BSD License, (the "License").
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


tomviz/ViewFrameActions.cxx
```diff
@@ -0,0 +1,57 @@
...
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+
+******************************************************************************/
+
+#include "ViewFrameActions.h"
+
+#include <pqRenderView.h>
+#include <pqSpreadSheetView.h>
```
mathturtle: Are these used?

mathturtle:
Tested and it works.

tomviz/ViewFrameActions.cxx
cryos: They were in my testing, but I yanked that code. I can remove them.

cryos:
Thanks for testing, I will merge this and see if it looks good in the binaries later.

PR #438, opened 2016-06-21
Make sure we update when the panel is visible

cryos:
Thanks to Shawn for tracking down the issue, ensure we update when the
panel is actually visible/the data source is changed.


mathturtle:
I've tested it and it handles the cases I have found where this was buggy.

Issue #434, opened 2016-06-20
Manual image alignment double-counts shifts


ElliotPadgett:
If the user uses the manual alignment tool to align the data, then clicks "ok" when finished, tomviz applies the shifts to the data and ALSO keeps the shifts in the manual alignment box.  Then, if they open their alignment operator again to update it, tomviz applies the saved shifts on the already shifted data, making the alignment incorrect.

To work around this the user can build up a stack of manual alignment operations, which is fine, but this issue defeats the purpose of having the alignment operator be editable the way it is.

mathturtle:
You're right, it is broken.  The first set of shifts are the only ones applied if you go back and edit the shifts it doesn't work.  I'll look into it.

mathturtle:
I've pushed a pull request that should fix this (I merged another related change last night for a bug I found when testing this).  @ElliotPadgett can you and/or @yijiang1 test this and see if it fixes what you were seeing please?

ElliotPadgett:
I can test after it comes out in the nightly build.

mathturtle:
Ok, so I tried a different dataset and now I see what you are talking about.  The fix in #437 addresses a related issue, but doesn't address this.  Working on a fix for this now.

mathturtle:
@ElliotPadgett #440 should fix this issue.  Sorry about the confusion earlier.

yijiang1:
@ElliotPadgett I've tested and the bug is fixed now.

ElliotPadgett:
Tested again.  This is now fixed

PR #440, opened 2016-06-21
Operators need middle-of-pipeline data before they show their editors

mathturtle:
Summary from first commit message (other changes are minor cleanups):

When an operator is edited we occasionally need information about the
dataset to construct its editor dialog.  For instance on the Crop
operator we need the origin, spacing and extent of the image to position
the widget and determine the range of valid entries in the input boxes.
But these values can be changed by subsequent transforms in the
pipeline, so the "current" values for that DataSource may not be valid.
This gets even stranger when the editor widget displays the
"intermediate" data like the translation align widget (Reported in #434)

This commit adds a method which gets a copy of the data in a DataSource
at a particular point in its data transform pipeline.  This new data
object is then passed to the Operator in getEditorWidget() so that the
editor widget will have accurate data for the input of the Operator.

### DIFF
--- a/tomviz/AddAlignReaction.cxx
+++ b/tomviz/AddAlignReaction.cxx
@@ -56,7 +56,7 @@ void AddAlignReaction::align(DataSource* source)
   }
 
   QSharedPointer<Operator> Op(new TranslateAlignOperator(source));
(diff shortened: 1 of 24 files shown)


mathturtle:
@cryos This is the change we discussed before lunch where the edit dialog needs a copy of the data from the middle of the pipeline.

Issue #439, opened 2016-06-21
Image should not freeze when "stop" button is clicked in manual alignment widget


yijiang1:
In manual alignment widget, the image cannot be changed when the "stop" button is clicked.
We should allow this to happen. The "stop" button should only stop flashing two images.

mathturtle:
Good catch.  Try out the change in #441 and see how that works.

ElliotPadgett:
This looks good now in today's nightly build.

PR #441, opened 2016-06-21
Align widget fixes

mathturtle:


mathturtle:
@cryos for #439.

yijiang1:
Tested well 👍

tomviz/AlignWidget.cxx
```diff
@@ -709,10 +721,20 @@ void AlignWidget::applySliceOffset(int sliceNumber)
...
+    this->modes[this->currentMode]->update();
+  }
+  this->widget->update();
 }
 
 void AlignWidget::startAlign()
 {
+  if (!(this->frameRate > 0))
+  {
+    qDebug() << "Frame rate must be positive to start timer";
```
cryos: I think we should favor opening up a warning rather than qDebug, but setting the minimum to 0, and counting a start with a frame rate of 0 as stop would seem reasonable and intuitive without a warning wouldn't it?

cryos:
I just checked and the frame rate minimum is 0, so making 0 an implicit stop if started and removing the qDebug would work great. To be defensive we would use <= 0 to stop.

mathturtle:
Updated.

PR #444, opened 2016-06-23
Add scale factor and unit to datasource with display in data properti…

mathturtle:
…es panel

@cryos @Hovden @yijiang1 This is a first pass at showing units and length data.  It defaults to assuming the spacing is in real-world units measured in nm, but allows you to change this on the data properties panel.  I know some transforms may modify the spacing... but I haven't looked for those cases yet, this just takes care of displaying the information.


yijiang1:
LGTM

Hovden:
Cool, let's merge and get feedback.

Issue #442, opened 2016-06-22
utils.set_tilt_angles does not work properly


yijiang1:
Here is a strange bug:
1. Open tomviz and load a dataset.
2. Click "Generate Tilt Series" in Tomography menu. Use default parameters.
3. When finished, the dataset is marked as a tilt series. However, its tilt angles are still zero.
<img width="282" alt="screen shot 2016-06-22 at 5 11 48 pm" src="https://cloud.githubusercontent.com/assets/13088588/16283828/78e539a6-389c-11e6-8548-2f686fcd4330.png">

mathturtle:
I've seen this myself and tracked it down to the same root issue as #426.  But it should have been fixed by #438.  Have you tried this with a build from today?  (I know the windows build machine is a bit behind, so only the OSX builds are up to date)

yijiang1:
> Have you tried this with a build from today?

Yes I'm running on the latest build.

mathturtle:
I just tried the latest build and it worked fine.  Maybe the build for that merge hadn't finished when you downloaded yours?

yijiang1:
I just tried the latest build (on Mac) and still have the problem.

mathturtle:
So I tried it again.... re-reading the issue and doing exactly what you said.  And I was able to reproduce it.  Actually @ElliotPadgett clued me into what was going on in his point 1 on #445.  Since I thought it was a display update issue, I was re-running the transform.  And the second time, since the dataset was *already* a tilt series, it worked fine.  Then on the newer builds I was only testing the display update issue (which IS fixed).

It looks like I introduced this in #322, a case we missed when testing it.  I had wondered why reconstructions acted strange the first time I ran them on a dataset...  It turns out this is why.  I just hope my solution doesn't break anything else.

For a consistent view of the data and tilt series info, I really think we need to adopt my proposed solution to #257 where any modifications to the data source type or tilt angles are saved as operators in the pipeline.  The current code feels kludgy and has strange corner cases like this.

mathturtle:
@yijiang1 The branch for this is merged now.

Issue #445, opened 2016-06-23
"Mark as tilt series" must be run twice for some datasets to unlock tomography menu


ElliotPadgett:
After loading a tilt series in tiff format, when I click "Mark as tilt series" and fill in the angle selection popup the angles get populated correctly but the tomography menu options are all greyed out except "Mark as volume" and "Generate tilt series".  I can work around the bug by clicking "Mark as volume" and then "Mark as tilt series" again, which makes the menu options available.

I have reproduced this with two different tiff stacks, but it does not seem to happen with the sample dataset .

ElliotPadgett:
@yijiang1 pointed out that this does not always reproduce, so I went and tried to see what conditions cause it and I noticed a few things:
1) Pushing the cancel button on "Mark as tilt series" does not stop it from marking it as tilt series (I can open a separate issue for this.
2) Deleting a dataset or pushing File>Reset with something open may be necessary to trigger the bug.
3) When it works correctly, the dataset gets selected in the "Visualization Modules" list to the left as soon as the "Mark as tilt series" popup appears.  When it does not work, this does not happen and the current visualization module stays selected instead.

mathturtle:
I think I have actually seen this once, but I couldn't reproduce it at all so I didn't mention it.  If you've seen it enough to report (and track down conditions to make it happen) then I can start debugging.  Thank you for taking the time to figure out some of what causes this.  Reading your list triggered an idea of what was going on in another issue too.

I can explain 1 from your list above... and probably fix it without too much trouble.  That menu item is doing two things.  First it is marking the data source as a tilt series.  Then it is calling the 'Set Tilt Angles' menu item's callback that displays the dialog.  When the dialog is canceled, it only cancels the 'Set Tilt Angles' part.

ercius:
Not sure what the status of this issue is, but the newest binary (0.8.0-151) requires the user to run "mark as tilt series" multiple times to unlock the Tomography menu. This was working correctly in a previous nightly (0.8.0-135) but is now broken again in 0.8.0-151.

mathturtle:
I was looking into it earlier and could only reproduce it when I do File->Reset.  I'm still trying to figure out why this happens.  @ercius are you on OSX?  Maybe it happens more there.

mathturtle:
[...]

PR #446, opened 2016-06-23
Preserve tilt angles from utils.set_tilt_angles

mathturtle:
@yijiang1 For #442, a (hopefully) minor fix.  Please test everything you can think of to do with tilt angles to be sure we don't miss anything else.  I will be doing the same.

### DIFF
--- a/tomviz/DataSource.cxx
+++ b/tomviz/DataSource.cxx
@@ -72,7 +72,10 @@ public:
       array->SetName("tilt_angles");
       array->SetNumberOfTuples(num_tilt_angles);
       array->FillComponent(0,0.0);
-      fd->AddArray(array.GetPointer());
+      if (!fd->HasArray("tilt_angles"))
+      {
+        fd->AddArray(array.GetPointer());
+      }
       this->TiltAngles = array.Get();
     }
     else


yijiang1:
This fixes #442 but reveals another bug probably related to #445 .
One consistent way to reproduce the bug:
1. Load a dataset.
2. Click "Generate Tilt Series" under "Tomography" menu.
When finished, the dataset is marked as a tilt series and tilt angles are updated like this
screen shot 2016-06-23 at 3 22 23 pm

4. Now, click "Reset" under "File" menu.
5. Repeat step 1 and 2.
When finished, the tilt angles are updated but the dataset is NOT marked as tilt series.
screen shot 2016-06-23 at 3 23 13 pm

mathturtle:
Updated branch.  Here is a slightly cleaner way to fix it.  It doesn't help the #445-related bug you found though.  That appears to be an update issue where the menu items don't get the dataset-was-updated signal correctly.  The data source is marked as a tilt series (otherwise the tilt angles wouldn't show up on the properties panel), but the menu wasn't notified.

tomviz/DataSource.cxx
```diff
@@ -66,13 +66,19 @@ class DataSource::DSInternals
     vtkFieldData* fd = data->GetFieldData();
     if (!this->TiltAngles)
     {
+      if (!fd->HasArray("tilt_angles"))
```
cryos: This does not seem to do anything.

mathturtle: I've deleted it, and updated the branch.  I think this was left over from the first implementation of this branch.  I copied the contents out into the containing block and I guess I forgot to delete this.

PR #453, opened 2016-06-28
Allow modules to use custom UI to modify proxies

mathturtle:
Only used by the contour module.

@cryos Since we talked I changed everything but contour to use the old code.  I haven't really gotten around to tweaking the UI to make it look like we wanted, but this is the proof-of-concept-it-works version.


mathturtle:
Ok, I've done some UI tweaking for the contour module.  Let me know what you think and what else we need to modify.  (Other modules still use the old code for now).

Issue #450, opened 2016-06-24
Save as PNG does nothing


ElliotPadgett:
When I try to save a tilt series or a reconstruction as a PNG the program closes the save dialog as if it were successful but does not produce a file and does not show any error.  I don't know if you can even have image stacks in PNG format, but there should at least be an error message if no saved file is produced.

Hovden:
Yep. This happens on my OSX machine.

ercius:
PNG does not allow for stacks. I suggest removing PNG as an output type for
data.

mathturtle:
@utkarshayachit @cquammen We are directly using ParaView here, is this a problem with ParaView's implementation of the PNGWriter or is it incorrectly reporting which readers support the data type?  Or something else?  I can fairly easily search/replace the filetype containing PNG from the valid file types passed to the dialog, but is that the right way to go here?

cryos:
I agree with @ercius, we should just remove PNG as it does not support the type of output required.

Hovden:
Also, the JPG does not save a stack, but appears to save the first slice. If there is not stack support we should also remove jpg. Not to mention jpg is a questionable format for science.

cryos:
Agreed, JPEG has no stack support I am aware of.

PR #455, opened 2016-06-29
Disable saving data as PNG or JPEG

mathturtle:
@Hovden For #450.


cryos:
Looks good to me, merging so it will be built by morning.

PR #456, opened 2016-06-30
Switch to fftw in direct Fourier reconstruction

yijiang1:
Switch to FFTW instead of numpy's FFT library in direct Fourier reconstruction.


Hovden:
Great. pyFFTW should be comparably fast with C++ FFTW.

yijiang1:
@Hovden so we can merge this now?

PR #457, opened 2016-06-30
Add example of using vtkImageFFT

mathturtle:
@yijiang1 Here is an example of using VTK's FFT.  I've tested that it compiles but given my knowledge off FFTs I wouldn't know if it was working right if I did look at the output...  Anyway, why don't you try to use it and we can work out what is going wrong if something doesn't look right.

There are two functions here, one shows how to use the FFT if you have a vtkImageData, the other shows how to create an image data from the raw array and dimensions.  Ideally you should probably use FFTW in that case, but we'll get to that later I guess.

### DIFF
--- a/tomviz/FFTExample.cxx
+++ b/tomviz/FFTExample.cxx
@@ -1,0 +1,38 @@
+#include <vtkPointData.h>
+#include <vtkDataArray.h>
+#include <vtkImageData.h>
+#include <vtkNew.h>
+#include <vtkImageFFT.h>
+
+
+// So I think this is the basic version of what you want to do.  I'm not sure though.
+// If you don't have a vtkImageData you would need to make one (see other function
+//
+// I don't know what the other options on VTK's FFT do... And you may want to set the dimensionality to 2
+// Here are the other options:
+// http://www.vtk.org/doc/nightly/html/classvtkImageFFT.html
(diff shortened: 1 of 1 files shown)


tomviz/FFTExample.cxx
```diff
@@ -0,0 +1,38 @@
...
+#include <vtkImageFFT.h>
+
+
+// So I think this is the basic version of what you want to do.  I'm not sure though.
+// If you don't have a vtkImageData you would need to make one (see other function
+//
+// I don't know what the other options on VTK's FFT do... And you may want to set the dimensionality to 2
+// Here are the other options:
+// http://www.vtk.org/doc/nightly/html/classvtkImageFFT.html
+void fftOfImage(vtkImageData *image)
```
yijiang1: Is vtkImageData always 3D? I do want to do 2D and even 1D FFT.

mathturtle: Well technically it is always 3D.  But if you want a 2D image, you just set the third dimension to be one slice thick.  And you may be able to do 2D FFTs across all the slices.  I'm guessing that is what vtkImageFFT does if you call `SetDimensionality(2)`, you'll probably want to test this to confirm.  The docs on vtkImageFFT are pretty sparse.

PR #460, opened 2016-07-07
WIP: Fixed a bug that Direct Fourier method doesn't take negative angles

yijiang1:
Fixed a bug that Direct Fourier method using fftw doesn't take tilt images with negative tilt angles.


yijiang1:
@Hovden I actually discovered more bugs and put "WIP" (work in progress) in the title. I'll fix the new bugs in another pull request.

Issue #461, opened 2016-07-07
More ways to rescale Fourier magnitude


yijiang1:
@Hovden 
When visualizing the Fourier magnitude of a dataset, tomviz should provide different ways to rescale the magnitude.
Right now, the "FFT (abs log)" transform takes the logarithm of the magnitude, which is not always the best option. For instance, if the dataset is low-pass filtered (often done by some constraint-based reconstruction algorithms), then taking log of 0 leads to -Inf.

PR #467, opened 2016-07-08
Added a small offset to FFT_AbsLog

yijiang1:
Added a small offset to the "FFT_AbsLog" transform. 
This avoids the infinity problem caused by taking log of 0 (as mentioned in #461).


Hovden:
Looks good. Added the smallest machine precision.

PR #471, opened 2016-07-11
Pipeline model unification

cryos:


cryos:
Merging as I would like to see the binaries on other platforms, please excuse some of the rough edges.

tomviz/PipelineModel.cxx
```diff
@@ -0,0 +1,596 @@
...
+bool PipelineModel::TreeItem::appendChild(const PipelineModel::Item &item)
+{
+  TreeItem *treeItem = new TreeItem(item, this);
+  m_children.append(treeItem);
+  return true;
+}
+
+bool PipelineModel::TreeItem::appendAndMoveChildren(const PipelineModel::Item &item)
+{
+  auto treeItem = new TreeItem(item, this);
```
mathturtle: We're allowed to use auto now?  (Not that I mind, I've been wanting to for a while).

cryos: Like I said at the time I wanted to ease into C++11 without going crazy. This branch seemed like a good time to see how things looked, and it let to a number of simplifications. I don't think we should spend time converting old code over to use auto, but please feel free to use it in new code.

tomviz/DataSource.cxx
```diff
@@ -51,7 +51,7 @@ class DataSource::DSInternals
 public:
   vtkSmartPointer<vtkSMSourceProxy> OriginalDataSource;
   vtkWeakPointer<vtkSMSourceProxy> Producer;
-  QList<QSharedPointer<Operator> > Operators;
+  QList<Operator*> Operators;
```
mathturtle: Do we need a destructor now that deletes these?

mathturtle: Ah, never mind.  I see how it works now.

Issue #468, opened 2016-07-08
Orthogonal slice does not update when the slice No. is changed in property panel


yijiang1:
This may be related to #453 .
When I change the slice number in property panel
<img width="299" alt="screen shot 2016-07-08 at 4 18 19 pm" src="https://cloud.githubusercontent.com/assets/13088588/16700366/a441edf0-4527-11e6-9c8e-67fb61557559.png">

The the slice in the "RenderView" window no longer update automatically.

Hovden:
Yep. Orthoslice is not moveable on recent release (OSX).

ercius:
Same on Windows 10 in 0.8.0-110. I also get this error message:
`QAbstractItemModel::endInsertRows:  Invalid index ( 2 , 0 ) in model tomviz::PipelineModel(0x219d2b5f830) (:0, )`

cquammen:
The error message is not exclusive to the orthogonal slice: see #477.

mathturtle:
This should be fixed (not the error message, the bug).

ercius:
Orthoslice updates properly using the latest binary on Windows 10.

Hovden:
Yep. Good on OSX too.

PR #483, opened 2016-07-12
Make sure modules call render() when something changes

mathturtle:
@Hovden Can you test this?  My fix for #468.

### DIFF
--- a/tomviz/ModuleOrthogonalSlice.cxx
+++ b/tomviz/ModuleOrthogonalSlice.cxx
@@ -140,6 +140,9 @@ void ModuleOrthogonalSlice::addToPanel(QWidget* panel)
     << "Slice";
   proxiesWidget->addProxy(this->Representation, "Slice", reprProperties, true);
   proxiesWidget->updateLayout();
+
+  this->connect(proxiesWidget, SIGNAL(changeFinished(vtkSMProxy*)),
+                SIGNAL(renderNeeded()));
 }
 
 bool ModuleOrthogonalSlice::serialize(pugi::xml_node& ns) const
--- a/tomviz/Module.h
+++ b/tomviz/Module.h
@@ -143,6 +143,10 @@ signals:
   /// map is modified
   void colorMapChanged();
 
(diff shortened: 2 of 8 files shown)


Hovden:
Orthoslice is working for my build (OSX)

Hovden:
@cquammen Can you give this code a review and merge now? It behaves correctly in the application.

cryos:
This works, but there is a fair bit of duplicated/boilerplate code. What is the mechanism it is replacing that went away? Can't the code to render be centralized a little more, rather than duplicated in all modules?

cryos:
Taking a better look I think that ModulePropertiesPanel::setModule would be a better place to connect to the signal, and add a signal to the base Module class that can be emitted. That class already has a render slot that can be reused, and it is looking for module related changed that require renders.

mathturtle:
@cryos that makes sense.  Here is a newer version that uses that approach.  I've tested it and it works fine locally.

tomviz/ModuleOrthogonalSlice.cxx
```diff
@@ -140,6 +140,9 @@ void ModuleOrthogonalSlice::addToPanel(QWidget* panel)
     << "Slice";
   proxiesWidget->addProxy(this->Representation, "Slice", reprProperties, true);
   proxiesWidget->updateLayout();
+
+  this->connect(proxiesWidget, SIGNAL(changeFinished(vtkSMProxy*)),
+                SLOT(requestRender()));
```
cryos: You know you can just connect a SIGNAL to another SIGNAL to forward a signal. There is no need to make a SLOT just to connect and emit a signal, and it should work fine when dropping arguments (like vtkSMProxy)

mathturtle: I didn't know that.  Neat.  I'm glad they thought of this when designing it since I think I've done very similar things before (maybe not on tomviz, but I don't remember).  I'll redo the branch to do it that way.

mathturtle:
Updated again and tested locally.

cryos:
Looks great, thanks!

PR #484, opened 2016-07-12
Update tilt angles text

mathturtle:
@Hovden


Hovden:
View > Toolbars
Is behaving oddly on my build version. Not sure if it is related to my local build, or if it is a result of change in code.

Hovden:
I think this is a local build thing. It is doing it on my other branches. I will merge and validate on the full build.

Issue #480, opened 2016-07-12
Strange behavior for Crop input boxes


ercius:
tomviz 0.8.0-110-... on Windows 10

The input boxes in the Crop GUI have a strange behavior.

1) Load a data set and then select Crop from the Transforms menu.
![image](https://cloud.githubusercontent.com/assets/15657108/16777818/ff784c78-481f-11e6-92b1-0e9241861e8f.png)

2) Type a number (for example 10) in the X start box
![image](https://cloud.githubusercontent.com/assets/15657108/16777829/0d699a1c-4820-11e6-8732-f71575fed5ac.png)

3) Now type a number (for example 500) in the X stop box:
![image](https://cloud.githubusercontent.com/assets/15657108/16777856/276480ee-4820-11e6-8902-4a606971b948.png)

Notice that the first number typed in the Stop box ends up in the Start box.

Hovden:
Yep. This is because we don't allow the start value to be higher than the end value. If you type the end value first, you won't have this problem.

How should we handle this?

Hovden:
@mathturtle might have a fix.

ercius:
Maybe its better to check these values when the user presses OK rather than on the fly.

mathturtle:
I've changed it to wait until the editor loses focus (you click outside it) or Enter is pressed to do the validation.  The branch is up for review now.

cryos:
Good stuff, should hopefully be more intuitive now.

PR #486, opened 2016-07-12
Wait for editing finished to validate text

mathturtle:
@Hovden Please test.  For #480.  Please test that the up/down arrow keys still update the shown volume box correctly.


Hovden:
Works well, OSX. Solves #480

PR #487, opened 2016-07-12
Change default outline color

mathturtle:
@Hovden please test.


Hovden:
This solves the white background problem.

PR #496, opened 2016-07-18
Compute some label object shape attributes

cquammen:
Introduces the use of ITK to compute label object shape attributes and
store them in the field data of the output data set. A label object
consists of all the voxels that have a particular label
value. Typically, these voxels are connected in some way, such as in a
connected component.

The attributes computed are object surface area (perimeter in ITK
parlance), object volume (number of voxels multiplied by the voxel
volume), and surface area to volume ratio.

Added to the Connected Components module.

### DIFF
--- a/tomviz/python/ConnectedComponents.py
+++ b/tomviz/python/ConnectedComponents.py
@@ -69,6 +69,47 @@ def transform_scalars(dataset):
 
         utils.add_vtk_array_from_itk_image(relabel_filter.GetOutput(), dataset, 'LabelMap')
 
+        # Now take the connected components results and compute things like volume
+        # and surface area.
+        shape_filter = itk.LabelImageToShapeLabelMapFilter.IUS3LM3.New()
+        shape_filter.SetInput(relabel_filter.GetOutput())
+        shape_filter.Update()
+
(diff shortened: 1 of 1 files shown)


cquammen:
To test:

- open one of the smaller data sets
- run the Connected Components data transform
- split Layout #1 by clicking the tool button 
image to the right of the "RenderView1" label.
- in the new view, choose a SpreadSheet View
- change the "Showing" combo box from "None" to "PassThrough1". Change "Attribute" to "Field Data".
- note the columns "SurfaceArea", "SurfaceAreaToVolumeRatio", and "Volume"

I would love for someone to look this early iteration over and suggest improvements.

cryos:
Thanks Cory, I will take a look this morning hopefully, and maybe we can talk in the afternoon at some point.

cryos:
I guess my data set is too large (using SMALL_PtCu_NanoParticle) after selecting PassThrough1.

cquammen:
Hmm, should take on the order of seconds to view in the SpreadSheet for that data set.

I would hold off trying this further until after we chat.

cquammen:
Subsumed by better way to report these results in #567.

PR #499, opened 2016-07-18
Ensure operators are restored from state files

cryos:
The data source is not added until after the operators have been
restored, so added some code to ensure they are added to the model and
displayed as expected.


cryos:
@mathturtle see any issues before I merge this? It addresses the fact that operators are deserialized before the data source is added.

Issue #459, opened 2016-06-30
Crash when loading a transform after tomographic reconstruction


yijiang1:
This happens on latest binary.
1. Open tomviz, load a dataset.
2. Mark it as tilt series. Do a reconstruction (e.g. weighted back projection).
3. Now, do a python-based data transform (e.g. rotate).
4. Double-click the transform in pipeline, and tomviz crashes.
<img width="263" alt="screen shot 2016-06-30 at 4 04 03 pm" src="https://cloud.githubusercontent.com/assets/13088588/16502428/5df439b2-3edc-11e6-972f-93611c837daa.png">

Interestingly, tomviz works fine if the tilt series is created by the "Generate Tilt Series" function.

mathturtle:
This should be fixed by #501.

Hovden:
When double clicking, it reruns all the transforms above it before displaying the python code.

mathturtle:
It should do that now... part of the fix in #440, we have to recompute the data just before the operation before showing the operation dialog.  At least in some cases like the alignment widget.... I think I might be able to skip it if the dialog doesn't need it.  Let me try that.

mathturtle:
So that wasn't very hard and it seems to work.  It will help with perceived responsiveness of the application too. I've made it a little bit better by not recomputing if it isn't used.  I added a new commit to #501 to add this.

Hovden:
This issue is fixed in latest OSX build.

PR #501, opened 2016-07-18
Tilt angles operator

mathturtle:
@cryos This should help with #459 and #445.  This branch makes the 'Mark as Tilt Series', 'Mark as Volume' and 'Set Tilt Angles' operations into Operators in the pipeline.

### DIFF
--- a/tomviz/DataSource.cxx
+++ b/tomviz/DataSource.cxx
@@ -92,21 +92,6 @@ public:
 
 namespace {
 
-// Converts the data type to a string for writing to the save state
-const char* dataSourceTypeToString(DataSource::DataSourceType type)
-{
-  switch (type)
-  {
-  case DataSource::Volume:
-    return "volume";
-  case DataSource::TiltSeries:
-    return "tilt-series";
-  default:
-    assert("Unhandled data source type" && false);
-    return "";
-  }
-}
-
 // Converts the save state string back to a DataSource::DataSourceType
 // Returns true if the type was successfully converted, false otherwise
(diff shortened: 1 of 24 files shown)


cryos:
Great, I will take a look this morning.

mathturtle:
New commit added in response to @Hovden's comment on #459.  An improvement in the responsiveness of the application in many cases, but doesn't have anything else to do with this branch (put here to avoid conflicts later).

tomviz/DataSource.cxx
```diff
@@ -92,6 +92,7 @@ class DataSource::DSInternals
 
 namespace {
 
+/*
```
cryos: Shouldn't we just delete this, and the one below if it is no longer needed?

cryos:
Looks good to me, I think you should just delete the commented out code if it is no longer needed - it will be in the history if we ever need to resurrect it.

mathturtle:
Updated to remove the code rather than commenting it out.  One thing I'm not sure of... if you load an old state file into this version of tomviz, what will happen.  I tested a little bit and it loads correctly, but I think some of the bugs that motivated this change will still happen in that case.

cryos:
Commented out will not help, but if we need a legacy code path then we should consider it. We need to build up a set of state files to guarantee behavior, and it is something that would be worthwhile after M&M.

Hovden:
Don't worry about legacy state files. Users can package older tomviz versions with datasets for reproducibility.  tomviz is too young to worry about backwards compatibility.

Issue #479, opened 2016-07-12
Typing in boxes in tilt axis alignment GUI


ercius:
The tilt axis alignment GUI  immediately updates the reconstructions when typing in any input box. This becomes tedious when for example you want to set the axis shift value to 20. It first shifts to 2, recalculates, then shifts to 20 and recalculates.

Further, the <Enter> keyboard button is connected to the the "Transform" button. I often accidentally press enter when trying to update a value therefore quitting the GUI.

I suggest the enter button is used by the user to update the input box values and clicking the "Transform..." button will exit the GUI. Pressing the arrow buttons next to the input boxes should immediately update the GUI images though.

cryos:
@mathturtle didn't you address this issue in a recent pull request?

Hovden:
No, it was only addressed for the cropping in  #486

The fix used there was:
-  this->connect(ui.startX, SIGNAL(valueChanged(int)),
 +  this->connect(ui.startX, SIGNAL(editingFinished()),

cryos:
Ah, OK - then @mathturtle can you look at adding a similar change there, and see if you can spot anywhere else the same pattern might have been inadvertently used.

cryos:
This should hopefully be resolved, please reopen if not.

PR #506, opened 2016-07-19
Delay valueChanged events to improve UI responsiveness when typing

mathturtle:
valueChanged fires after every character typed and for slower operations
this can be annoying.  But editingFinished doesn't fire when the up/down
arrows are clicked on spin boxes, which is also annoying.

For #479.  @cryos @Hovden I don't know if I like this any better.  It fixes the typing issue but breaks the live update when clicking the arrows on the spin box (at least on Linux).

### DIFF
--- a/tomviz/AddPythonTransformReaction.cxx
+++ b/tomviz/AddPythonTransformReaction.cxx
@@ -845,10 +845,6 @@ OperatorPython* AddPythonTransformReaction::addExpression(DataSource* source)
   }
   else if (scriptLabel == "Reconstruct (Constraint-based Direct Fourier)")
   {
-    vtkTrivialProducer *t = vtkTrivialProducer::SafeDownCast(source->producer()->GetClientSideObject());
-    vtkImageData *data = vtkImageData::SafeDownCast(t->GetOutputDataObject(0));
-    int *extent = data->GetExtent();
-    
     QDialog dialog(pqCoreUtilities::mainWidget());
     dialog.setWindowTitle("Constraint-based Direct Fourier Reconstruction");
     
--- a/tomviz/AddPythonTransformReaction.cxx
+++ b/tomviz/AddPythonTransformReaction.cxx
@@ -20,6 +20,7 @@
 #include "OperatorPython.h"
 #include "SelectVolumeWidget.h"
 #include "EditOperatorDialog.h"
+#include "SpinBox.h"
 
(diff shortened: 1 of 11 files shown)


cryos:
Isn't there another signal for when the arrows are clicked?

cryos:
Yeah I guess not, we could get a little clever and fire a one shot timer, only update after 1 second or something. I would have hoped editingFinished would have been emitted after the arrows have been clicked.

mathturtle:
Here is the update with the one-shot timer.  1 second seems too long in some cases, but I think we should err on the long side for now.

cquammen:
Boy, this is a lot of code. I worry about maintaining it. Something more encapsulated, like a custom Qt spin box that fires the signals you want, would probably be preferable.

mathturtle:
You have a point Cory.  And I'm not really happy with the solution...  I'm working on a new version that detects clicks to the buttons in the spin box (based on [this thread](https://forum.qt.io/topic/61529/about-qspinbox-how-to-make-it-emit-editingfinished-signal-when-click-the-up-down-arrow-button/7)).  I have no idea why Qt doesn't provide a signal for this.

mathturtle:
Here is a new version that uses a custom subclass of QSpinBox/QDoubleSpinBox that makes the editing finished signal behave the way we want it to.

Hovden:
I suggest 250 milliseconds. According to Google, this is the magic number for people's patience regarding delays in UI (http://www.nytimes.com/2012/03/01/technology/impatient-web-users-flee-slow-loading-sites.html)

cryos:
So my main suggestion would be to shorten the class names, they are already in the tomviz namespace, and the Qt ones are called QSpinBox. Why not go with tomviz::SpinBox and tomviz::DoubleSpinBox? Otherwise looks great, and glad we can forego the arbitrary delay. Good to know the magic number should we need to use arbitrary delays in other places though, I will try to remember that.

mathturtle:
Updated with shortened class names.  My one concern is will `#include "SpinBox.h"` be ambiguous?  Hopefully CMake puts that include directory first.

cquammen:
I don't think there is another header file named `SpinBox.h` included in tomviz. Qt's header is QSpinBox, QDoubleSpinBox, etc.

Tested locally, works great.

LGTM

cryos:
I am not worried about name collision, and "" around includes means take the local file first anyway.

PR #507, opened 2016-07-19
Module properties panel

cryos:


cquammen:
Tested after rebasing on master for Shawn's fix in #508. Works nicely and opens up more visualization real estate on the screen.

LGTM

Hovden:
Moving properties panel to the left is good, but worried that scrollbars will appear. Also, how does it look with animation panel in place? I agree with this design decision, just want to make sure it is ready.

I need get VTK setup to build now, so unable to test before merge.

cquammen:
image

This is what it looks like when you start tomviz with the -dr command line option, which prevents any saved window settings (including panel locations) from being loaded.

Hovden:
A scroll bar shouldnt appear when there is no content in the properties panel. Or when all of the content fits in the properties panel. I think there is some invisible frame being placed.

It looks like the properties panel will be too small for the data properties contents and some of the visualization properties.

cquammen:
Agreed about the scroll bar in principle. However, this kind of thing can be surprisingly fiddly to fix, so I would put a low priority on it.

The properties panel might start out small for new users, but after a one time enlargement it will stay bigger.

Hovden:
Should we hold off on moving the properties panel over if it introduces fiddly scroll bars and information that is not visible without scrolling? It sounds like this will take some extra work to implement.

cquammen:
Maybe @cryos or @mathturtle know the secret sauce to do these UI adjustments quickly. If so, might as well do that now. I'm just advising from my own experience that it seems to be a black art.

If the change are not quick, I think the increased visualization space is a huge win compared to a few fairly minor (IMHO) UI issues. But I'm happy to defer to y'all.

cryos:
I will take a look at the scroll bar, but our properties panels are small and I think letting the user resize the panel once (and likely closing the animation dock as most seem to) will still be a big win. Also the locality of the pipeline and properties make it easier to switch and edit. I would prefer to merge, but can hold off if you feel very strongly about the scroll bar @Hovden

mathturtle:
I played with it for a few minutes to see if I could get it to work, but nothing I tried helped.  Too advanced a black art for me I guess.

The one thing I find strange is that when you shrink the window and then make it bigger, the module properties stays small while the pipeline widget expands.  It seems like the priority there should be equal (or maybe even give the module properties panel more space).

Hovden:
If we can handle the scroll bars, let's do it, otherwise I'd like to optimize layout in the next release. If we cannot remove the whitespace below all the properties can we have the properties panel take up ~1/2 of the vertical space?

cryos:
So the scroll bars are there because the data properties panel is that long. I think that the data properties panel should be tweaked such that the length in x, y and z is set on one line with the simple labels X, Y, and Z rather than X Length (units), ...

That will help reduce the length of the panel, but the panel always reserves the length it is expected to require.

mathturtle:
> So the scroll bars are there because the data properties panel is that long.

That could be fixed by putting each panel in the QStackedWidget inside its own scroll area rather than having the scroll area around the QStackedWidget.  Then you would only get a scroll bar if the active panel in the stacked widget was too long.

cryos:
There it is with the separate scroll areas, looks pretty good to me, quite a bit of tweaking but likely worth it in the long run. If you guys are happy then please merge this. I tweaked a few other layout pieces to make the default take up a little more space too.

cquammen:
This is what it looks like now at first startup.

image

Looks a lot better.

However, I can't seem to adjust the distribution of vertical space between the Pipelines and Properties panels.

A possibly unrelated problem: once a dock widget has been dislodged from the main window, I can't figure get it to re-dock.

mathturtle:
> However, I can't seem to adjust the distribution of vertical space between the Pipelines and Properties panels.

Probably because the are both at their minimum size.  I can only adjust to a point because of this.  Try making the window bigger and you should be able to adjust a bit.

> A possibly unrelated problem: once a dock widget has been dislodged from the main window, I can't figure get it to re-dock.

@Hovden has reported this on OSX.  Apparently it wasn't fixed by Qt 5.7.  I don't have any trouble here on Ubuntu.

mathturtle:
The length UI changes don't seem to have affected anything...  but it didn't do very much with that to begin with.

cryos:
I am going to merge, unfortunately macOS has some odd issues with dock panels at the Qt level that we cannot do much about. I think this branch is an improvement, and saves space in a number of ways that can be effectively used by the visualization with minimal loss to the pipeline/properties panels.

Hovden:
Ok, I think you're right.  I will check out the build and put feedback in an issue.

cryos:
The scroll bars are gone, and only appear when you switch to the data properties if needed. Several other improvements were made in response to feedback, and we can still tweak this more of course.

Hovden:
That's great. It is certainly a big improvement to the design.

PR #508, opened 2016-07-19
Use unsigned int rather unsigned long long

mathturtle:
Apparently pugixml doesn't support unsigned long long on all platforms
(or in all versions or something).  This was causing build failures, so
switch to unsigned int


mathturtle:
@cryos This compiles locally, but so did the last one...

PR #509, opened 2016-07-19
Separate data transform menu

cquammen:
This branch lays some groundwork for conditionally enabling Data Transforms based on attributes of the Data Source. The idea is that some Data Transforms will be available only if a segmentation exists in a Data Source, but this is by no means the only attribute that may be queried to determine whether a Data Transform may be applied.

This branch also lays the ground work for dynamically loading Python script Data Transforms from users. Every time the Data Transforms menu is clicked, the menu items it contains are rebuilt, so it is plausible to look for new Python scripts in certain locations, add menu items, and create new reactions for them.

### DIFF
--- a/tomviz/CMakeLists.txt
+++ b/tomviz/CMakeLists.txt
@@ -84,6 +84,8 @@ set(SOURCES
   DataPropertiesPanel.h
   DataSource.cxx
   DataSource.h
+  DataTransformMenu.cxx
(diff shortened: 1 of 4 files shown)


cquammen:
Note: this neither adds nor subtracts functionality from tomviz. It is only an infrastructure change.

@cryos

Issue #478, opened 2016-07-12
Incorrect tilt angles attribute when saving as XMF/HDF5


ercius:
tomviz 0.8.0-110- crashing (Windows 10) when setting data set as Tilt series for data loaded from XMF/HDF5 files.

1) Load a data set and set as tilt series with tilt angles specified
2) Save a data set as XMF (also generates .h5 HDF5 data file)
3) Load the XMF file.
4) Press set data set as tilt series (Notice that a single tilt angle is specified already)
5) Press OK to set the tilt angles and tomviz crashes

The XMF file has an attribute with only a single tilt angle:
`<Attribute Name="tilt_angles" AttributeType="Scalar" Center="Grid">
    <DataItem Dimensions="1" NumberType="Float" Precision="8" Format="XML">
        -70 
     </DataItem>
</Attribute>`

Deleting `<attribute>....</attribute>` in the XMF file fixes the problem.

This attribute should have all the tilt angles written into the XMF file or specify no tilt angles as a work around.

Hovden:
Can you send a sample of the file?

ercius:
The system originally rejected my XMF data set. I generated a 100x100x10 data set From the Sample Data menu and reproduced the error. See the data in the attached ZIP file.
[testZeros.zip](https://github.com/OpenChemistry/tomviz/files/359955/testZeros.zip)

mathturtle:
So I've fixed the crash, but the data is still questionable.  It looks like a limitation or bug in the XMF/HDF5 writer where it isn't handling field data arrays the way we would expect.  It is only saving the first tilt angle in the array and setting the array size to 1.

@utkarshayachit Is this correct behavior of the XMF writer/reader?

Hovden:
@ercius Can you test now that #510 is merged?

ercius:
The crash is fixed now and it operates exactly as @mathturtle describes.

It would be nice to save the tilt angles and automatically set the data as a tilt series when loaded. But this is a minor point which possibly can be answered by @utkarshayachit

PR #510, opened 2016-07-19
Ensure tilt angles array is long enough before writing to it

mathturtle:
@Hovden This should address #478.  I've tested it and the example dataset no longer crashes for me when you load it in and set the tilt angles.  It was a memory corruption bug.... so I'm glad it crashed rather than doing something even stranger.


cquammen:
LGTM

mathturtle:
Thanks @cquammen for the quick review.  Merging and we can get them to test it tomorrow.

Issue #511, opened 2016-07-19
Display Problem when opening on Mac


VeraBeer:
When opening the newst Tomviz version on Mac it looks "glitshy", see screenshot.

![screen shot 2016-07-13 at 4 06 19 pm](https://cloud.githubusercontent.com/assets/20047722/16967766/be7fe672-4dd8-11e6-9624-ee44e147e4fe.png)

![screen shot 2016-07-13 at 4 11 43 pm](https://cloud.githubusercontent.com/assets/20047722/16967770/c24407f2-4dd8-11e6-97ac-baf1648cc4f8.png)

Hovden:
Woah. Haven't seen that. My version works. Does it always look like that?

VeraBeer:
Yes, always since I downloaded the latest version, all the times slightly different. I know that others had the same.
But forgot to mention, as soon as I scroll or move the mouse it disappears

mathturtle:
It isn't nearly as bad here, but I do see some small glitches like this when I open it on my mac.  For me it is just a black bar on the right of the view.  Then again I don't have other applications using the GPU, and it looks like it is dumping out uninitialized GPU memory to the screen.

cquammen:
I first noticed this when tomviz moved to Qt 5.6. @VeraBeer  Is this from a nightly binary downloaded from http://www.tomviz.org/downloads/ or built on your machine locally?

VeraBeer:
The nightly one

ElliotPadgett:
I see this too on the nightly binary I downloaded today.  Here's an example:
![tomvizglitch](https://cloud.githubusercontent.com/assets/12242066/16968556/9bde8268-4ddd-11e6-8a45-75e074e823ee.png)

cquammen:
FWIW, I am building tomviz against a very recent ParaView master, and I no longer see these rendering issues. I think @utkarshayachit did some recent Qt5-related work to fix this in ParaView and/or VTK.

cquammen:
Update: I still don't usually see this when starting tomviz, but I did just see it one time this morning.

cryos:
Hopefully this is improved now, it seemed to alleviate the obvious issues on some test systems.

PR #522, opened 2016-07-21
Attempt to remove artifacts in initial render windows

cquammen:
Try rendering all views one second after the MainWindow class is instantiated.

### DIFF
--- a/tomviz/MainWindow.cxx
+++ b/tomviz/MainWindow.cxx
@@ -445,6 +445,14 @@ void MainWindow::showEvent(QShowEvent *e)
     this->Internals->isFirstShow = false;
     QTimer::singleShot(1, this, SLOT(checkForAutosaveFile()));
   }
+
+  // Add a delayed render after the main window is shown.
+  // This is an attempt to work around a problem with Mac render
+  // windows not repainting appropriately.
+  //
+  // NOTE: remove this once painting events in QVTKWidget or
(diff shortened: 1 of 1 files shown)


cquammen:
This is a stab at fixing the artifacts seen at startup as described in #511. Because I don't consistently see the artifacts in my Mac development build, I have not been able to verify that this fixes the problem.

We might try merging this to see if it fixes the superbuild for Mac

cquammen:
Mind reviewing, @mathturtle ?

mathturtle:
I might move it to the showEvent() function (near the bottom of the file).  And possibly use a shorter timeout.  I did see the glitch for a second and then the render went through and it dissappeared.

cryos:
This may alleviate the symptom a little but it is clear that the QVTKWidget has issues and is not repainting as often as it should. This even happens on Linux at times, and a render call may make it a little less apparent. The Mac builds are pretty far behind right now, if we want to put this in then add a comment reminding us to remove it once QVTKWidget/its successor is working better.

mathturtle:
Yes, QVTKWidget has issues with Qt5.  And there really isn't a plan to fix it.  So it will be once tomviz migrates to its successor.

cquammen:
I incorporated all suggestions in this branch.

mathturtle:
@cquammen That is MUCH better.  I can barely notice it now and I know what to look for.

PR #523, opened 2016-07-21
Change default colormap

cquammen:
Changed to "Plasma_17"

### DIFF
--- a/tomviz/Behaviors.cxx
+++ b/tomviz/Behaviors.cxx
@@ -31,6 +31,7 @@
 #include <pqStandardViewFrameActionsImplementation.h>
 #include <pqViewStreamingBehavior.h>
 #include <vtkSMReaderFactory.h>
+#include <vtkSmartPointer.h>
 #include <vtkSMTransferFunctionPresets.h>
 #include <vtkNew.h>
 #include <vtkSMSettings.h>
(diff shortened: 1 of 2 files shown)


cquammen:
Mind reviewing, @mathturtle @cryos ?

mathturtle:
So the code looked ok, but I found that I needed a few more changes to make it work when running with `-dr` to clear the settings first.  See https://github.com/mathturtle/tomviz/commit/a7f7777d1ecddb561f0b18c88f19a40cf9e75e83  for the changes I made to get it to work.

cquammen:
Thanks for checking with the "-dr" option. I mistakenly thought `vtkSMTransferFunctionPresets` was a singleton, but that's not the case. Hmm, it shouldn't work even without the `-dr` option, I think. Wonder why it does. I'll push a version that works in both modes.

mathturtle:
I think it works without -dr if your tomviz settings already have those presets (the presets object serializes itself to the settings in its destructor).  I don't know how it found the presets though, since the path for OSX was wrong.  Did you import them manually at some point?

Anyway, the new version looks good and it works in my testing.

cquammen:
Ah, explanation makes a lot of sense.

I wound up copying the matplotlib presets to the path specified in the code prior to this change after tomviz started complaining that it couldn't find the file. I wasn't sure if that path should have been modified, so left it as was.

PR #524, opened 2016-07-21
Changed organization from 'Kitware' to 'tomviz'

cquammen:


cquammen:
Okay with you, @Hovden ?

cryos:
It is fine, merging, will bump the version later too

PR #526, opened 2016-07-21
Remove the decoration from the pipeline view

cryos:
This gets rid of the arrows, lines, etc from the pipeline view.


cryos:
image

Hovden:
Looks good.

cryos:
Great, I will get this merged in - as you can see it just enables a custom stylesheet I had experimented with earlier.

Hovden:
bigmac is behind on tomviz builds. If we can get one soon, we will squeeze the new UI screenshot will into Yi's poster.

mathturtle:
bigmac is behind on everything... so it is unlikely to get to tomviz before sometime tonight.  You may just want to create a local build, it will be much faster than waiting on bigmac.

PR #527, opened 2016-07-21
Update to display new minimum required versions

mathturtle:
@cryos Found this when explaining to someone how to build ParaView... and realized we were out of date here.


BUILDING.md
```diff
@@ -16,12 +16,12 @@ be used to point to them if they are not found automatically.
 Dependencies
 ------------
 
- * Qt 5.5.1
- * CMake 3.1 (2.8.8 minimum)
+ * Qt 5.5.1 (known bugs make 5.7.0 the preferred version)
```
cryos: Just say 5.7.0 recommended

BUILDING.md
```diff
@@ -16,12 +16,12 @@ be used to point to them if they are not found automatically.
...
+ * CMake 3.5
```
cryos: This was relaxed to CMake 3.3

cryos:
Please simplify to 5.7.0 recommended in brackets, and correct to CMake 3.3 as I pointed out inline. This reminds me that I think there is a minor commit I made updating some CMake code, I will look on my laptop when I get home.

mathturtle:
Done.

I also added a commit that adds some helpful flags for building ParaView to the build instructions.

You may also want to update our minimum required CMake (in both tomviz and the superbuild).  It is pointless for tomviz to require an older version than paraview since you have to build paraview too.  I think you had wanted to hold off so that we could use the default CMake on Ubuntu LTSs, but that can't work anymore because of ParaView (except for 16.04 LTS which ships 3.5).

cryos:
Like I said above - "think there is a minor commit I made updating some CMake code, I will look on my laptop when I get home", I remember writing it at the VTK hackathon to update the CMake minimum and use the new C++11 variables in CMake 3.3+. This looks good now, merging.

Issue #277, opened 2015-11-23
Default Load State


Hovden:
I found new users that try out the software are lost as to where to start. They just want to see something in 3D and start exploring from there.

I suggest we have a default load state, with a simple isosurface visualization of a small dataset. This might be the star particle and a colorful isosurface.

This should be simple, but one issue to worry about, as tomviz progresses, the state files may become more complex and this could cause crashes in future builds.

cryos:
I think we can do that on first load, but we should be careful to make it easy to disable too. You don't want regular users delete the default state every time the application loads, and likely want to offer the option on first load not to show this next time. We already have the sample data to help guide new users, and another option is also popping up a dialog on first load asking if they want to visit a tutorial page.

ercius:
Maybe have a splash screen with options to load a sample dataset and
visualization? Then have a check box that will never show the splash screen
again?

cryos:
Yes, I think something along those lines would be ideal. Not sure if we can realistically help with shared systems, might need some thought there if we want to help.

PR #528, opened 2016-07-21
Show a welcome dialog that asks to load sample data on startup

mathturtle:
@cryos Taking a stab at #277, based loosely on the ParaView welcome dialog.  The UI could certainly be improved, I just threw something together.  I wasn't sure if this was still something we wanted for 0.9 or not.

### DIFF
--- a/tomviz/CMakeLists.txt
+++ b/tomviz/CMakeLists.txt
@@ -192,6 +192,8 @@ set(SOURCES
   vtkCustomPiecewiseControlPointsItem.h
   vtkNonOrthoImagePlaneWidget.cxx
   vtkNonOrthoImagePlaneWidget.h
+  WelcomeDialog.cxx
+  WelcomeDialog.h
   )
 
 set(python_files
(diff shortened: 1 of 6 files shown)


cryos:
I think this needs more work, visually I would like to improve it a little, it doesn't detect that sample data is not there, and silently does nothing if it is not. We should use the tomviz logo from the about dialog too. It does nicely show up on first run, and respect the request not to see it again.

mathturtle:
Here is a new version (screenshot below).  It should now detect that the sample data is not present too.  tomviz_welcome_dialog_v2

Hovden:
Looks good. Minor comments:

How does it behave if the dataset is not there? If it isn't, it shouldn't give you the option to load.

I would like it to say "Would you like to load an example dataset and visualization?"

Upon clicking "Yes" it should load a volume render and bounding box. (Orthoslice is optional, whatever you want).

cryos:
Sounds good, once the wording change is integrated we can merge this.

mathturtle:
The wording change is pushed.  @Hovden right now it throws up a warning dialog if the sample data file is not found.  If it is there then it loads it like you would load any other dataset... but changing it to a volume rendering shouldn't be that hard.  @cryos Do we want the other changes that @Hovden asked for in this branch?

Hovden:
We can add a volume render in a new branch.

mathturtle:
@Hovden Your additional requests didn't take very long and I was looking for something quick... See #551.

Hovden:
Sounds good.

PR #529, opened 2016-07-21
Solve some invalid model index errors in PipelineModel

cquammen:
Fixes some QModelIndex errors reported by
PipelineModel::endInsertRows() when a data source is first loaded. The
problem was that the last row was not the same as the first when only
one new row is being inserted, but it should be.

Some invalid model index errors persist when an operator is first added.

### DIFF
--- a/tomviz/PipelineModel.cxx
+++ b/tomviz/PipelineModel.cxx
@@ -454,7 +454,7 @@ QModelIndex PipelineModel::moduleIndex(Module *module)
 
 void PipelineModel::dataSourceAdded(DataSource *dataSource)
 {
-  beginInsertRows(QModelIndex(), 0, 1);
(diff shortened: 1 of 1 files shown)


cquammen:
@cryos This branch appears to fix the initial model index errors that occur when a data set is first loaded. However, I have been unable to fix the model index error that occurs when you first add an operator. Subsequent operators can be added without generating additional errors - it seems only to occur when the module children are moved upon the first addition of an operator. However, I have been unable to find exactly where the invalid index is being generated, or how to fix it.

cquammen:
@cryos For your consideration. The invalid index errors remain when adding only the first module to a data set.

cryos:
Let's merge this, and I will see if I can spot the remaining issue.

PR #531, opened 2016-07-22
Bump to CMake 3.3, use C++11 support added in 3.1

cryos:
ParaView requires 3.3 now, require at least that.


cryos:
@mathturtle this bumps CMake minimum, and switches to use the C++11 support from CMake 3.1.

mathturtle:
LGTM.

Issue #530, opened 2016-07-22
Layout Tweaks


Hovden:
The new Pipeline with the Properties Panel is a great improvement.

Can we reduce the width of the right panels and the height of the animation panel. The current build:
<img width="1285" alt="screen shot 2016-07-21 at 10 08 25 pm" src="https://cloud.githubusercontent.com/assets/6197997/17045150/1b7aa642-4f95-11e6-9454-a75b633f9758.png">

A preferred appearance:
<img width="1276" alt="screen shot 2016-07-21 at 10 48 11 pm" src="https://cloud.githubusercontent.com/assets/6197997/17045165/3a65578c-4f95-11e6-8e7a-21bad7742695.png">

cryos:
This was merged, closing.

PR #532, opened 2016-07-22
Tweak the initial size/layout of the docks

cryos:
This addresses issue #530 by adjusting the initial size of the dock
widgets, something that has only recently been made possible. In my
testing this results in the initial layout being changed, but when
loading with a config file the settings from there are respected.

### DIFF
--- a/tomviz/MainWindow.cxx
+++ b/tomviz/MainWindow.cxx
@@ -132,6 +132,18 @@ MainWindow::MainWindow(QWidget* _parent, Qt::WindowFlags _flags)
   QIcon icon(":/icons/tomviz.png");
   setWindowIcon(icon);
 
+  // Tweak the initial sizes of the dock widgets.
+  QList<QDockWidget*> docks;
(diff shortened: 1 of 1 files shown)


cryos:
@mathturtle @cquammen can you confirm the behavior I observe with this branch? In my testing removing the config file or passing -dr shows the layout with smaller docks as we want, if the config file is there it respects whatever the user did (closed, resized, etc).

cryos:
The function was not introduced until Qt 5.6, so I can bump the minimum version requirement too (or guard if we really want to support older Qt, don't think we do).

cquammen:
I'm on it.

cquammen:
LGTM running with the `-dr` option.

cryos:
@cquammen did the opposite case look good where you say closed the animation dock, made the other two wider, closed and reopened as normal. Was the restored geometry kept as you would expect?

cquammen:
Yes.

cryos:
Thanks for testing, @mathturtle just independently verified the behavior too. I will merge this in now.

PR #535, opened 2016-07-25
Fix rpath-related packaging on OSX

mathturtle:
@cryos This fixes the tomvizExtensions library's packaging on OSX.

### DIFF
--- a/tomviz/pvextensions/CMakeLists.txt
+++ b/tomviz/pvextensions/CMakeLists.txt
@@ -20,8 +20,17 @@ add_paraview_plugin(tomvizExtensions "1.0"
 )
 target_link_libraries(tomvizExtensions PUBLIC pqApplicationComponents)
 
-install(TARGETS tomvizExtensions
-        RUNTIME DESTINATION "bin"
-        LIBRARY DESTINATION "lib"
-        ARCHIVE DESTINATION "lib"
-	COMPONENT "runtime")
+if (NOT APPLE)
+  install(TARGETS tomvizExtensions
+          RUNTIME DESTINATION "bin"
+          LIBRARY DESTINATION "lib"
+          ARCHIVE DESTINATION "lib"
+          COMPONENT "runtime")
+else()
+  install(TARGETS tomvizExtensions
+          RUNTIME DESTINATION "bin"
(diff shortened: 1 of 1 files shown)


tomviz/pvextensions/CMakeLists.txt
```diff
@@ -20,8 +20,17 @@ add_paraview_plugin(tomvizExtensions "1.0"
...
+if (NOT APPLE)
+  install(TARGETS tomvizExtensions
+          RUNTIME DESTINATION "bin"
+          LIBRARY DESTINATION "lib"
+          ARCHIVE DESTINATION "lib"
+          COMPONENT "runtime")
+else()
+  install(TARGETS tomvizExtensions
+          RUNTIME DESTINATION "bin"
+          LIBRARY DESTINATION "Applications/tomviz.app/Contents/Libraries"
```
cryos: You can just use BUNDLE DESTINATION in the main call, rather than using if/else I think, see 'cmake help-command install'. Do you know why this changed when bumping the minimum CMake version? I am guessing that is what happened. I am also pretty sure the BUNDLE location means it can be a relative path within the bundle, keeping this a little simpler.

mathturtle: This didn't change.  The install tree was using @rpath in the install name of the tomvizExtensions  library and the fixup_bundle script doesn't handle that very well.  So rather than debug the script for hours, I changed the install code so that the fixup_bundle script doesn't need to find this file since it is already installed correctly.

I didn't know you could do that BUNDLE part of the install command on non-executables...  I'll try it and see if it works.

mathturtle: So I tried it... it doesn't work.  Apparently only executables can have a `BUNDLE DESTINATION`, it just ignores it for libraries.  (It is also possible that if I could access the `add_library` command and insert the arguments `BUNDLE 1`, it would work... but that isn't possible from `add_paraview_plugin`).  There is also a [cmake bug report](https://gitlab.kitware.com/cmake/cmake/issues/14671) about this saying that it only works for executables and should be extended to all other targets.

cryos:
Ah, OK, thanks for checking into this. Merging this, and hopefully they will extend CMake in the future.

Issue #538, opened 2016-07-28
Reconstructions in tilt axis alignment UI do not scale properly with data size


yijiang1:
Images in reconstruction windows in the manual tilt axis alignment UI do not scale properly with data size.
<img width="828" alt="screen shot 2016-07-27 at 9 57 46 pm" src="https://cloud.githubusercontent.com/assets/13088588/17198742/b0bfbe22-5445-11e6-8bd4-633910ea1c3d.png">

bdalevin:
To add to Yi's comment. The brightness/contrast in each of these reconstruction slices needs to be individually autoscaled. Otherwise it is very difficult to spot artifacts in the reconstruction. Fixing this is very important for making the software easy for general users to use.

It would also be very helpful to be able to change the colour maps assigned to these reconstruction slices to help look for artifacts. Right now the only way to do this is to change the colour map for the whole dataset, which is OK, but not ideal.

cryos:
Can you look into this bug @mathturtle please.

mathturtle:
So it turns out this is a side effect of #536.  I have a branch that will fix the issue that @yijiang1 reported.  My question is: how many of the things that @bdalevin asked for do we want done immediately?  @cryos @Hovden

Hovden:
Only need to fix the bug. We can make a separate issue for the issue raised by @bdalevin and discuss / prioritize it later.

mathturtle:
Ok, then the branch I have in that MR should be sufficient.

mathturtle:
So are we making a new issue for the additional requests?  I have a local branch that takes a first pass at the separate colormaps scaled to the range of the slice thing.  It looked useful.

mathturtle:
See #552.

Hovden:
Great, thanks for handling @bdalevin suggestion.

yijiang1:
Right now, only the third reconstruction window is zoom-able. We should allow zooming for all three windows.

mathturtle:
@yijiang1 Strange... I see that behavior too.  I'll look at it.  So you want zoom and pan and probably a reset camera button for each reconstruction window?  This should be a new issue.

yijiang1:
> So you want zoom and pan and probably a reset camera button for each reconstruction window?

@Hovden and @ElliotPadgett may have stronger opinions on this than I do. 
I just think the three recon windows should behave similarly.

Issue #536, opened 2016-07-26
Manual tilt axis alignment UI doesn't initialize properly


yijiang1:
On v0.9, when I first open the manual tilt axis alignment UI, the three reconstruction windows show the tilt image instead of reconstructions.

PR #541, opened 2016-08-02
Compute reconstruction slices when UI opens

mathturtle:
For speed of popping up the dialog, we used to use slices from the
original data, make this use slices from the reconstruction instead even
though it will take longer.

Addresses #538, since that was a side effect of #536.


mathturtle:
The UI may be slower to appear since it has to compute those slices.... I didn't do anything fancy in the first pass like firing a single shot timer to compute the slices.  We may want to do that.

cryos:
Merging this, and we can look at fancier approaches if necessary.

Issue #542, opened 2016-08-03
Modifying the Crop Transform runs all transforms


ercius:
Tomviz 0.9.0 on Windows 10. Modifying the crop transform causes all other transforms to run twice:
1) Add a rotate transform (for example)
2) Add a crop transform and crop the data
3) Double click on the crop transform to modify it
4) All previous transforms (the rotate transform in this case) are run again before the Crop GUI shows up
5) Crop the data to new values and click OK
6) All previous transforms are run a second time.
This gets very time consuming and I do not think that the process needs to be run twice.

mathturtle:
I know why it does this right now and I'll see if I can restructure it so that it doesn't.

PR #548, opened 2016-08-04
Cache a weak pointer to state before operator so to save on re-executing

mathturtle:
@cryos This is a first pass at #542.  Let me know what you think.  I don't think it will increase the memory footprint by too much since I'm using vtkWeakPointer and it should significantly reduce the re-execution time.


mathturtle:
ping?

cryos:
Looks good to me

PR #550, opened 2016-08-05
Fix index errors

cquammen:


cquammen:
@cryos On my system, this fixes the model index errors when adding the first operator to a data source.

cryos:
LGTM

PR #551, opened 2016-08-08
Welcome dialog

mathturtle:
Addresses the additional requests made by @Hovden in #528. i.e. it doesn't show the welcome dialog if the sample data is not found and it replaces the orthoslice with a volume render when loading the sample data from the welcome dialog.


cquammen:
Tested with and without data in place.

Without data in place, I did not see the welcome screen.

With data in place, I did see the welcome screen. Clicked Yes and sample data was loaded with volume rendering and outline. Clicked No and no sample data was loaded. Clicked option to not show again and I did not see it again on next startup.

LGTM

cryos:
Thanks for the testing, what you have looks good to me too.

PR #552, opened 2016-08-08
RotateAlignWidget: add set colormap for each reconstruction slice button

mathturtle:
Allow the colormap for each reconstructed sample slice to be set
individually and scaled to the range for that slice rather than the
whole data range for better detection of artifacts.

This adds a button to each reconstructed slice beside the spinbox that selects which slice number.  This new button allows the user to set the colormap for that reconstructed slice.  These colormaps are now separately rescaled to the range of data in the reconstructed slice that they correspond to each time the slice is updated.  See @bdalevin's comment on #538 for the motivation for this change... It looked very useful for the amount of effort I had to put in to implement it.

@Hovden @cryos

### DIFF
--- a/tomviz/RotateAlignWidget.cxx
+++ b/tomviz/RotateAlignWidget.cxx
@@ -26,6 +26,8 @@
 #include "AddPythonTransformReaction.h"
 #include "Utilities.h"
 
+#include "pqCoreUtilities.h"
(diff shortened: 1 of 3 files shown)


Hovden:
Great.

cryos:
Looks good to me

PR #554, opened 2016-08-09
Unindent modules

cquammen:
In preparation for adding derived data sets, unindent the Modules so
that users do not confuse Modules with derived data sets. Operators
are still on top and Modules are on the bottom. They red X's on the
right of the Operators should distinguish them sufficiently from
Modules with eyeball visibility icons.

### DIFF
--- a/tomviz/PipelineModel.cxx
+++ b/tomviz/PipelineModel.cxx
@@ -161,14 +161,13 @@ bool PipelineModel::TreeItem::remove(DataSource *source)
   if (source != dataSource()) {
     return false;
   }
-  for (auto i = 0; i < childCount(); ++i) {
-    if (child(i)->module()) {
-      remove(child(i)->module());
-    }
-    else if (child(i)->op() && child(i)->childCount()) {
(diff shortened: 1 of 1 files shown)


cquammen:
Here's a screen grab of the unindented arrangement.

image

cquammen:
@cryos @Hovden Ready for review.

cryos:
@cjh1 just to make sure you are aware, new changes from @cquammen to the tree model stuff we were talking about.

cjh1:
Thanks, that was pretty much the cleanup I was going to do!

cryos:
Looks good, as @cjh1 has been looking through some of this it would be great to get his opinion on the changes. I don't have time to look at them right now, but visually I think this is a nice change (still on the fence about not indenting visualization modules but I think it looks pretty good, keeps the visual impact we wanted, and will not end up being overly indented as pipelines get more advanced).

cjh1:
This PR LGTM, I have based another PR off it that does some clean up.

cquammen:
Closing because this has been subsumed by #558.

PR #555, opened 2016-08-11
Proposed clang-format file for Tomviz

cryos:
This is based quite closely on the CMake style, which is based on the
Mozilla style. It makes it very easy to conform to a consistent style,
and after discussions with Brad it looks like it is pretty easy to add
into our development workflow.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,7 @@
+---
+# This configuration requires clang-format 3.8 or higher.
+BasedOnStyle: Mozilla
+AlwaysBreakAfterReturnType: None
+AlwaysBreakAfterDefinitionReturnType: None
+BreakConstructorInitializersBeforeComma: false
+...
--- a/tomviz/ActiveObjects.cxx
+++ b/tomviz/ActiveObjects.cxx
@@ -25,17 +25,12 @@
 #include <vtkSMSourceProxy.h>
 #include <vtkSMViewProxy.h>
 
-namespace tomviz
-{
+namespace tomviz {
 
 ActiveObjects::ActiveObjects()
-  : Superclass(),
-    ActiveDataSource(nullptr),
-    VoidActiveDataSource(nullptr),
-    ActiveDataSourceType(DataSource::Volume),
-    ActiveModule(nullptr),
-    VoidActiveModule(nullptr),
-    MoveObjectsEnabled(false)
+  : Superclass(), ActiveDataSource(nullptr), VoidActiveDataSource(nullptr),
+    ActiveDataSourceType(DataSource::Volume), ActiveModule(nullptr),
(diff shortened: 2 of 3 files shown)


cquammen:
Looks cool. Sadly, homebrew on Mac has only up to version 3.7.1, so it complains with

    YAML:4:29: error: unknown key 'AlwaysBreakAfterReturnType'

I'm sure there is a way to get a more up-to-date version. 3.7.1 also does something funny with comments. Trying it on `tomviz/Operator.h`, for example, turns some comment blocks from

    -  /// Returns the number of steps the operator has finished.  The total number of
    -  /// steps is returned by totalProgressSteps and should be emitted as a finished

to

    +  /// Returns the number of steps the operator has finished.  The total number
    +  /// of
    +  /// steps is returned by totalProgressSteps and should be emitted as a
    +  /// finished

cryos:
Shame about the version, the comments thing looks like it is enforcing a line length limit, I will take a look and see if that is off but I would understand it doing that if they hit the line length

mathturtle:
Looks good to me.

@cquammen Are you sure your brew is updated?  I just tried it and got clang-format 3.9.0.

cquammen:
Ah, maybe I need to update brew's database. I'll check.

cquammen:
Confirmed that 3.9.0 is available on updated homebrew.

cquammen:
I suggest adding

`ReflowComments: false`

to prevent `clang-format` from goofing up the comments.

cryos:
What do you mean by goofing up comments? If they are over the line length limit at least it becomes obvious and we can fix the cases that irritate us by hand. I would like to apply this soon, and then start using it to format our code.

cquammen:
This

    /// There are two versions of this function, this one and getEditorContentsWithData.
    /// Subclasses should override this one if their editors do not need the previous
    /// state of the data.  Subclasses should override the other if they need the data
    /// just prior to this Operator for the widget to display correctly.  If this data
    /// is needed, the default implementation to return nullptr should be left for this

gets transformed to this

    /// There are two versions of this function, this one and
    /// getEditorContentsWithData.
    /// Subclasses should override this one if their editors do not need the
    /// previous
    /// state of the data.  Subclasses should override the other if they need the
    /// data
    /// just prior to this Operator for the widget to display correctly.  If this
    /// data
    /// is needed, the default implementation to return nullptr should be left for
    /// this

True, these could be manually modified to fit within the column width limit, which appears to be 80 columns. Doing so and rerunning `clang-format` does not goof up the comments.

LGTM

cquammen:
Shall we merge?

cryos:
Let's merge, and then we can run this once most branches have been merged to avoid conflicts.

PR #557, opened 2016-08-12
Travis clang format

cjh1:
This PR adds support for running clang-format on PRs using Travis. Included in the PR is a change that touches a file to cause clang-format to be run on the change. The [check fails](https://travis-ci.org/OpenChemistry/tomviz/builds/151804402) and suggests the appropriate formatting change. In the future we can add other tests to Travis and of cause we could build tomviz ( against pre build ParaView etc. ).

### DIFF
--- a/STYLE.md
+++ b/STYLE.md
@@ -1,3 +1,22 @@
+clang-format
+------------
+
+We use [clang-format](http://llvm.org/releases/3.8.0/tools/clang/docs/ClangFormatStyleOptions.html)
+to keep formatting in the code base consistently. Please run clang-format
+on your patches before submitting.
+
+clang-format ships with a python script ```clang/tools/clang-format-diff.py``
+that can be using to reformat patchs. For example the following command will
+reformat all the lines in the latest commit
+
+```shell
+git diff -U0 HEAD^ | clang-format-diff.py -i -p1
+
+```
+
+Code style
+----------
+
 This project is developed primarily in C++ and Python. Please follow these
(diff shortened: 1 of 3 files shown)


cjh1:
@cryos PTAL

STYLE.md
```diff
@@ -1,3 +1,22 @@
+clang-format
+------------
+
+We use [clang-format](http://llvm.org/releases/3.8.0/tools/clang/docs/ClangFormatStyleOptions.html)
+to keep formatting in the code base consistently. Please run clang-format
```
cryos: I think you mean consistent here.

STYLE.md
```diff
@@ -1,3 +1,22 @@
...
+on your patches before submitting.
+
+clang-format ships with a python script ```clang/tools/clang-format-diff.py``
+that can be using to reformat patchs. For example the following command will
```
cryos: 'that can be used to reformat patches'

STYLE.md
cjh1: I do

cryos:
Pointed out a few minor things inline, looks great otherwise. Arch puts it at /usr/share/clang/clang-format-diff.py - is that where other distros place that script? It would be good to make it easier to use when committing.

cjh1:
Ubuntu actually provides an executable: clang-format-diff-3.8, Should we add these location some how?

cryos:
Merge this, and maybe we could add a section on commonly used distros, :+1: thanks for looking into this.

PR #558, opened 2016-08-12
Pipeline model cleanup

cjh1:


cjh1:
@cryos @cquammen PTAL

tomviz/PipelineModel.cxx
```diff
@@ -182,6 +182,9 @@ bool PipelineModel::TreeItem::remove(Module *module)
   foreach(auto childItem, m_children) {
     if (childItem->module() == module) {
       removeChild(childItem->childIndex());
+      // Not sure I like this, an alternative is to make TreeItem a
+      // QObject and use a signal?
+      ModuleManager::instance().removeModule(module);
       return true;
```
cjh1: Any suggests here? I think its better than the double traversal we where doing before.

cquammen:
Unless there is a compelling reason to abbreviate names or parameters or variables, I would rather keep the full names rather than shortening them, e.g., maintain "module" instead of shortening to "mod". "Op" is obviously a special case to get around the C++ keyword "operator".

cjh1:
It was to get rid of the compiler warnings ( shadow warnings ) ...

cquammen:
Sorry, didn't see that. That is an annoying warning. Could we silence it in the interests of code readability?

cjh1:
I would be happy with that, I agree its a backwards step to abbreviate things to keep the compiler happy.

cjh1:
@cryos What do you think?

cryos:
So I think this is much improved in recent GCC versions, so maybe we can disable the shadow warning for older versions of GCC. I think there is value to warning on shadowed variables, and it has helped spot subtle issues in the past on stuff I have worked on. I am on GCC 6.1.1 here, and it does not warn about any shadowed variables in the Pipeline* files. I actually only see two warnings and I will submit a patch to fix them, I agree shadow used to be a little over the top but still saw value in using it.

cjh1:
Ok, I will update my compiler and revert my last commit.

cjh1:
Should we add a minimum/recommended compiler version?

cryos:
We should probably add a recommended compiler version, we need decent C++11 support, but it looks like outside of warnings GCC 4.8 supports everything we are using right now.

Looking at the changes you and Cory made it looks good to me. Once I have my tagged build I think this should get merged - still waiting on the Windows builds there though.

cjh1:
FYI I was using 4.8.4, the default on Ubuntu 14.04

cjh1:
Compiler upgraded and removed my last commit ...

cquammen:
LGTM

Tested locally, works as expected.

PR #559, opened 2016-08-15
Add missing header

mathturtle:
@cquammen Can you try this branch?  This seems to have fixed it on bigmac.  I don't know why gcc let it compile, maybe something else was pulling in <array> on Ubuntu?

@cryos When testing on bigmac, this fixed the OSX compile issue.


cryos:
LGTM, I am going to merge as it is always good practice to include the header of the class you want to use. Hopefully we will see a build result soon.

cquammen:
My Mac is happy with master now after this merge.

Issue #539, opened 2016-08-01
Downsample x2 Needs to be Different for Tilt Series Data


bdalevin:
When data is marked as Tilt Series Data, Downsample x2 should only scale the data along the X and Y spatial dimensions, and should leave the tilt dimension alone. 
Currently, Downsample x2 acts the same as it does on volume data, scaling along all 3 dimensions, including tilt, which is bad. 

Alternatively this Transform could be disabled for tilt series data, forcing users to use Resample instead.

mathturtle:
@Hovden which approach do you think fits here?  I'm more in favor of changing how it works to only scale in X and Y, but I'm not a scientist... What would make the most sense to you?

ercius:
I would leave this transform as is and add a similar one to the tilt series
menu which does what was requested (xy binning only). That would allow for
both transforms and keep the confusion to a minimum.

yijiang1:
I prefer @ercius 's approach. I can add this to tomviz if no one is working on it.

mathturtle:
AFAIK, noone is working on it.  Go ahead.

ercius:
@yijiang1 I suggest renaming the current transform to indicate it operates on the full volume. "Bin volume x2" for example.

bdalevin:
I've tested this out on the latest build. It works well. I'm happy to close the issue.

PR #562, opened 2016-08-17
Added a function that bin tilt series by 2

yijiang1:
Added a python function that bin tilt series by 2 to tomography menu. This addresses #539 

I also renamed to the old "Downsample x2" function in data transform menu to "Bin volume x2" and added a feature to downsample tilt angles in case the dataset is marked as a tilt series.


mathturtle:
The changes make sense.  My main concern is the text for the menu items.  Will people in your field know what 'Bin Volume x2' means?  I know they understand 'Downsample x2'.

bdalevin:
The phrase "Bin" to mean Downsample is used in Gatan's Digital Micrograph, and in ImageJ, so people should be familiar with this terminology.

Hovden:
Yeah. It's a tough call. Downsample is more widely used and I prefer it slightly. However Bin has the advantage that it tells the user that the values are being averaged (binned) and not sampled by every other pixel (tossing out information; something people actually do).

Hovden:
We can revisit "downsample" v. "bin" in a future pull request. This looks good.

Issue #565, opened 2016-08-18
State file does not save colormap range or opacity settings


ercius:
Windows 10, tomviz 0.9.0 (release)

State files do not reload colormap range and opacity settings. 

I want to recall these settings in the Histogram using a state file to produce a nice visualization:
![clipboard02](https://cloud.githubusercontent.com/assets/15657108/17787261/458fe736-653d-11e6-9a74-0afc007b6d9d.png)

However, saving a state file and reloading it sets the opacity and range to the full range:
![clipboard01](https://cloud.githubusercontent.com/assets/15657108/17787278/5518830c-653d-11e6-8ee1-6f851c562157.png)

Hovden:
This is important for reproducibility

mathturtle:
I'm going to look at this today.

cquammen:
Fix was merged.

ElliotPadgett:
This only looks half fixed to me.  The state file now saves the color map but fails to save the opacity settings.

ercius:
I've also noticed this in recent Windows builds but was holding off
reporting it until I could get the newest nightly from Cdash.

ElliotPadgett:
@cquammen @mathturtle Any comment on this?  Altering the opacity function for a volume rendering is very common, so this issue is really a nuisance for making or sharing volume renderings.

cquammen:
Sorry, not ignoring. I'll take a look later today.

ElliotPadgett:
Thanks!

cquammen:
Update: it seems that only the default scalar opacity function is being saved to the state file, which means the `DataSource` has a reference to the wrong one. Looking into why.

cquammen:
Actually, I was wrong in my previous comment. The `ScalarOpacityFunction` proxy is the same in the `DataSource` (when a separate color map is not used).

The actual problem is that the `HistogramWidget` modifies the client-side `vtkPiecewiseFunction`. However, the opacity map values never make it back from the VTK object to the proxy.

A good question is: how does it work with the color map, which works essentially the same way? I'm not sure yet. I have a quick hack in mind that should at least solve this problem, which I will push up sometime today, hopefully.

cquammen:
Hi folks, this should be fixed in #685, which has been merged to master. Please reopen if you have any problems.

ElliotPadgett:
[...]

PR #566, opened 2016-08-19
Change deserialize in DataSource to not clobber colormap range

mathturtle:
We were saving the range correctly... and we were loading the colormap.  But then when it loaded the data, it was clobbering the colormap range.  So I changed the order of operations.

See #565.  @cquammen @Hovden @ercius

A more general question is if we do in fact want to reset the colormap range every time the data changes (the current behavior) or if we want to have some logic to test if it should be reset (and what that behavior should be).


ercius:
I noticed that clobbering problem whenever a transform is run. I'm for
adding a feature to lock the color map if desired.

cquammen:
I tested it locally, it works as expected. LGTM

PR #575, opened 2016-08-30
Remove unused member functions

cquammen:
Remove some unused functions.


cquammen:
Minor change found while wrestling with some PipelineModel stuff to support segmentations in the UI. Hopefully a small win for me. Mind reviewing, @cjh1.

cjh1:
LGTM, are we sure these weren't added with some future functionality in mind? Looks like there where added by @cryos

cquammen:
These were needed prior to our recent PipelineModel changes. My preference would be to remove unused code and revive from the history if it proves useful in the future.

PR #576, opened 2016-08-31
Child data sets

cquammen:
Adds basic support for single child data sets produced by an operator. Modified the BinaryThreshold operator to produce its label map as an output datset rather than a second array in the main dataset.

### DIFF
--- a/tomviz/DataTransformMenu.cxx
+++ b/tomviz/DataTransformMenu.cxx
@@ -104,7 +104,9 @@ void DataTransformMenu::buildMenu()
   new CropReaction(cropDataAction, mainWindow);
   new ConvertToFloatReaction(convertDataAction);
   new AddPythonTransformReaction(binaryThresholdAction, "Binary Threshold",
-                                 readInPythonScript("BinaryThreshold"));
+                                 readInPythonScript("BinaryThreshold"), false,
+                                 false,
+                                 readInJSONDescription("BinaryThreshold"));
   new AddPythonTransformReaction(
     connectedComponentsAction, "Connected Components",
     readInPythonScript("ConnectedComponents"), false, false,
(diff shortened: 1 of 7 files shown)


cquammen:
Please take a look when you get a chance, @cjh1 @cryos. Note that the child dataset will not show up in the UI in this PR, but if you run the Binary Threshold operator on some test data, you should see in a SpreadSheet view a new trivial producer proxy. If you delete the Binary Threshold operator, the new trivial producer proxy should also be removed.

cryos:
This is still freezing for me when I attempt to view the trivial producer (I see a 1 and 2). Chris just tried this and it worked for him, so I am not sure what the local issue is. I need to head home soon, but am going to try a fresh ParaView build, and see what this looks like on my laptop. I remember it was freezing for me before with an earlier branch too, so it may be something machine specific - the backtrace doesn't give any clues, it looks to be frozen in Qt's event loop.

cquammen:
Hmmm. Maybe try with the smallest image data you have? It takes a couple seconds to load up the SMALL data set image I have available, so my guess was it was a function of trying to show too much data. In any case, one shouldn't normally need to look at the child data set in the SpreadSheet view, so hopefully it's not a big deal.

cryos:
It seems to take forever on even the smallest things, I see the error Error: Could not find script file:  "BinaryThreshold.json" on startup too, giving up for now, it may be a bug in the Qt style being used too.

cquammen:
Ah, CMake copies over the JSON files at configuration time, so please try running cmake manually next time you can try it. That should resolve the JSON file not found error.

cjh1:
@cquammen I still see the error @cryos is reporting after rerunning cmake.

cquammen:
Thanks for trying out this branch. I'm taking a look at the JSON file loading problem.

cquammen:
Closing as all the commits in this PR are contained in #577.

Issue #572, opened 2016-08-29
Crashes when deleting data / All datasets deleted when deleting top dataset


Hovden:
Crashes:
1) open two datasets
2) delete the second dataset from pipeline
3) crashes
Deletes all data
1) open two datasets
2) delete the top dataset from pipeline
3) both datasets deleted

Tested on OSX

cquammen:
Confirmed crashing when deleting data sets, but with a different testing sequence:

1) Open two datasets
2) Delete first
3) Delete second. Crash.

Confirmed deletion of all data sets using the "Delete" key. Cannot reproduce by deleting with "Delete" option in context menu with right-clicking the data source in the pipeline.

cquammen:
By the way, catching crash in debugger doesn't reveal anything obvious. Whoever modifies the `PipelineModel` class to support child data sets should look at this, too.

cquammen:
This *seems* to be resolved in the latest master. Please report if you continue to have problems.

Hovden:
Only partially fixed. No longer crashes, but still deletes all data:
1) open two datasets
2) delete the top dataset from pipeline
3) both datasets deleted

cryos:
I do not see this with master, just following up on old bugs. Is this behavior still an issue for you @Hovden? I tried deleting it both from the top, and from the bottom, using the delete key and the context menu.

Hovden:
Yeah. Open two sample datasets. Delete a viz module in the top dataset.

I'm finding several bugs in the pipeline (with two datasets try scrolling down by pressing the down arrow key. You'll find it repeats.) I will create a bug list.

yijiang1:
On dashboard, the latest binary's version number is 
<img width="313" alt="screen shot 2016-10-18 at 2 52 26 pm" src="https://cloud.githubusercontent.com/assets/13088588/19491899/a42e40a0-9542-11e6-80bb-03d5e81d4763.png">

My local build's version number is
<img width="209" alt="screen shot 2016-10-18 at 2 53 52 pm" src="https://cloud.githubusercontent.com/assets/13088588/19491914/b5d3816c-9542-11e6-9262-0ee2f1572ded.png">

Are they the same version?

cryos:
I see the repeat, I can take a look at it. Not able to reproduce the deletion of all data sets here though. I was going to take a look through the code and see if I can spot anything.

cryos:
[...]

Issue #392, opened 2016-05-05
UI that incorporates new data generated by data transforms


cquammen:
Some data transforms may generate new data sets, such as label maps that describe segmented objects. Linking these new data objects to their source in the UI so that the user knows what is being visualized is very important, as is maintaining the link between original and derived data for data provenance.

Here is a mockup of a potential UI to show original data, the transforms applied to it, and visualizations of all original and derived data sets.

![image](https://cloud.githubusercontent.com/assets/360056/15051920/12043c9e-12c9-11e6-9dbb-f9362aac2951.png)

The idea is to convey the linear/hierarchical relationship among original data set and derived data sets. Each data set may be represented with a number of visualization modules.

There are obvious similarities with the Pipeline Browser in ParaView. The key difference is that tomviz would define visualizations for each data set original, and derived.

Hovden:
It looks good from a user design perspective.

Would the data transforms then only apply to the parent dataset, before segmentation?

cquammen:
In my view, a segmentation is just another form of data transform, but it is one that produces new data. With that view, you can just chain the data transforms together as you would do now. After segmentation, you might want a transform that yields a volume measurement, for example, so you could add a volume measurement transform after segmentation.

cryos:
It breaks the model we chose for Tomviz, where we generally keep no mid points, so it is not possible to visualize things in the middle of a pipeline. Does this become an exception, how do we best express that, perhaps it is reasonable.

cquammen:
I'm aware that I may be biased from my work on ParaView, so I'm trying hard to do what is best for tomviz users rather than just revert to what is familiar.

In my image analysis work, I have found it is useful - necessary, even - to visualize mid points, particularly if you have a set of processing steps that you apply to different tomographic reconstructions and something goes wrong in the output. Do we need to provide this capability in tomviz? Not necessarily, but instinctively I want to.

For another perspective, Slicer's model let's you view up to two grayscale images at the same time (orthogonal slices) with a user-settable fade between the grayscale images for comparison purposes, and it lets you view at most one label map with adjustable opacity to show grayscale images (orthogonal slices). We could follow this model by listing label maps associated with a data set and letting you select just one to visualize, rather than having visualizations of the label maps themselves. LabelMap-aware visualization modules could then use this label map as needed.

cryos:
[...]

PR #577, opened 2016-09-01
Child data sets UI

cquammen:
Adds support to the `PipelineModel` for showing child datasets and visualization modules created to visualize them.

It is not yet possible to apply Operators to child datasets, but that's the next item to implement.

### DIFF
--- a/tomviz/OperatorPython.cxx
+++ b/tomviz/OperatorPython.cxx
@@ -300,11 +300,12 @@ bool OperatorPython::applyTransform(vtkDataObject* data)
         m_childDataSourceNamesAndLabels[i];
       QString name(nameLabelPair.first);
       QString label(nameLabelPair.second);
-      PyObject* child = PyDict_GetItemString(outputDict, name.toLatin1().data());
+      PyObject* child =
+        PyDict_GetItemString(outputDict, name.toLatin1().data());
       if (!child) {
         errorEncountered = true;
-        qCritical() << "No child data source named '"
-                    << name << "' defined in output dictionary.\n";
+        qCritical() << "No child data source named '" << name
+                    << "' defined in output dictionary.\n";
         continue;
       }
 
@@ -343,7 +344,6 @@ bool OperatorPython::applyTransform(vtkDataObject* data)
       const char* objectReprString = PyString_AsString(objectRepr);
       qCritical() << "Dictionary return from Python script is:\n"
                   << objectReprString;
-
     }
   }
 
--- a/tomviz/OperatorPython.cxx
+++ b/tomviz/OperatorPython.cxx
@@ -184,8 +184,8 @@ void OperatorPython::setJSONDescription(const QString& str)
       Json::Value nameValue = childDatasetNode[0]["name"];
       Json::Value labelValue = childDatasetNode[0]["label"];
       if (!nameValue.isNull() && !labelValue.isNull()) {
-        QPair<QString, QString> nameLabelPair(nameValue.asCString(),
-                                              labelValue.asCString());
+        QPair<QString, QString> nameLabelPair(QString(nameValue.asCString()),
+                                              QString(labelValue.asCString()));
         m_childDataSourceNamesAndLabels.append(nameLabelPair);
       } else if (nameValue.isNull()) {
         qCritical() << "No name given for child DataSet";
@@ -298,13 +298,13 @@ bool OperatorPython::applyTransform(vtkDataObject* data)
     for (int i = 0; i < m_childDataSourceNamesAndLabels.size(); ++i) {
       QPair<QString, QString> nameLabelPair =
         m_childDataSourceNamesAndLabels[i];
-      const char* name = nameLabelPair.first.toLatin1().data();
-      const char* label = nameLabelPair.second.toLatin1().data();
-      PyObject* child = PyDict_GetItemString(outputDict, name);
+      QString name(nameLabelPair.first);
+      QString label(nameLabelPair.second);
+      PyObject* child = PyDict_GetItemString(outputDict, name.toLatin1().data());
       if (!child) {
         errorEncountered = true;
         qCritical() << "No child data source named '"
-                    << nameLabelPair.first << "' defined in output dictionary.\n";
+                    << name << "' defined in output dictionary.\n";
         continue;
       }
 
@@ -333,7 +333,7 @@ bool OperatorPython::applyTransform(vtkDataObject* data)
         DataSource* childDS =
           new DataSource(vtkSMSourceProxy::SafeDownCast(producerProxy),
                          DataSource::Volume, this);
-        childDS->setFilename(label);
+        childDS->setFilename(label.toLatin1().data());
         setChildDataSource(childDS);
       }
     }
--- a/tomviz/OperatorPython.cxx
+++ b/tomviz/OperatorPython.cxx
@@ -19,13 +19,22 @@
 #include <QPointer>
 #include <QtDebug>
 
+#include "DataSource.h"
 #include "EditOperatorWidget.h"
 #include "OperatorResult.h"
 #include "pqPythonSyntaxHighlighter.h"
+
 #include "vtkDataObject.h"
+#include "vtkNew.h"
 #include "vtkPythonInterpreter.h"
 #include "vtkPythonUtil.h"
+#include "vtkSMParaViewPipelineController.h"
+#include "vtkSMProxy.h"
+#include "vtkSMProxyManager.h"
+#include "vtkSMSessionProxyManager.h"
+#include "vtkSMSourceProxy.h"
 #include "vtkSmartPyObject.h"
+#include "vtkTrivialProducer.h"
 #include <sstream>
 
 #include "vtk_jsoncpp.h"
@@ -233,6 +242,8 @@ bool OperatorPython::applyTransform(vtkDataObject* data)
   // Look for additional outputs from the filter returned in a dictionary
   PyObject* outputDict = result.GetPointer();
   if (PyDict_Check(outputDict)) {
+
+    // Results (tables, etc.)
     for (int i = 0; i < numberOfResults(); ++i) {
       OperatorResult* operatorResult = resultAt(i);
       std::string resultName = operatorResult->name().toStdString();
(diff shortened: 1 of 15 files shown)


cjh1:
@cquammen I ran the Binary Threshold, should I expect to see a child dataset? This is was I see in my UI:

tomviz

cquammen:
Indeed, you should see the child hanging off the Orthogonal Slice module. It's probably not there because of the JSON file not found issue you are having, which causes the Binary Threshold operator to think it doesn't produce any children.

cquammen:
@cjh1 Please try running CMake after updating your local branch. I forgot to add the `BinaryThreshold.json` file to the list of files to copy to the build directory.

cjh1:
@cquammen That resolved the JSON file issue and the child dataset now appears. I have seeing the following signal related message:

```
QObject::connect: invalid null parameter (:0, )

```

I also get a seg fault when I tried to to remove the Binary Threshold, this stack is below:

```
Program received signal SIGSEGV, Segmentation fault.
0x00007ffff359cbe4 in vtkSMCoreUtilities::GetFileNameProperty (proxy=0x3c47340)
    at /home/cjh/work/source/paraview/ParaViewCore/ServerManager/Core/vtkSMCoreUtilities.cxx:50
50	  if (proxy->GetHints())
(gdb) bt
#0  0x00007ffff359cbe4 in vtkSMCoreUtilities::GetFileNameProperty (proxy=0x3c47340)
    at /home/cjh/work/source/paraview/ParaViewCore/ServerManager/Core/vtkSMCoreUtilities.cxx:50
#1  0x00000000004d9dd1 in tomviz::DataSource::filename (this=0x5a5a020)
    at /home/cjh/work/source/tomviz/tomviz/DataSource.cxx:199
#2  0x000000000051fc02 in tomviz::PipelineModel::data (this=0xd41960, index=..., role=0)
    at /home/cjh/work/source/tomviz/tomviz/PipelineModel.cxx:276
#3  0x00007ffff5abc080 in QStyledItemDelegate::initStyleOption(QStyleOptionViewItem*, QModelIndex const&) const () from /home/cjh/Qt5.6.1/5.6/gcc_64/lib/libQt5Widgets.so.5
#4  0x00007ffff5abab79 in QStyledItemDelegate::paint(QPainter*, QStyleOptionViewItem const&, QModelIndex const&) const () from /home/cjh/Qt5.6.1/5.6/gcc_64/lib/libQt5Widgets.so.5
#5  0x00007ffff5a73ea6 in QTreeView::drawRow(QPainter*, QStyleOptionViewItem const&, QModelIndex const&) const () from /home/cjh/Qt5.6.1/5.6/gcc_64/lib/libQt5Widgets.so.5
#6  0x00007ffff5a7875e in QTreeView::drawTree(QPainter*, QRegion const&) const ()
   from /home/cjh/Qt5.6.1/5.6/gcc_64/lib/libQt5Widgets.so.5
#7  0x00007ffff5a7db56 in QTreeView::paintEvent(QPaintEvent*) ()
   from /home/cjh/Qt5.6.1/5.6/gcc_64/lib/libQt5Widgets.so.5
#8  0x00007ffff5827bd8 in QWidget::event(QEvent*) ()
   from /home/cjh/Qt5.6.1/5.6/gcc_64/lib/libQt5Widgets.so.5
#9  0x00007ffff5924b9e in QFrame::event(QEvent*) ()
   from /home/cjh/Qt5.6.1/5.6/gcc_64/lib/libQt5Widgets.so.5
#10 0x00007ffff5a4106b in QAbstractItemView::viewportEvent(QEvent*) ()
   from /home/cjh/Qt5.6.1/5.6/gcc_64/lib/libQt5Widgets.so.5
#11 0x00007ffff5a7ea10 in QTreeView::viewportEvent(QEvent*) ()
   from /home/cjh/Qt5.6.1/5.6/gcc_64/lib/libQt5Widgets.so.5
#12 0x00007ffff49ef8f3 in QCoreApplicationPrivate::sendThroughObjectEventFilters(QObject*, QEvent*) ()
   from /home/cjh/Qt5.6.1/5.6/gcc_64/lib/libQt5Core.so.5
#13 0x00007ffff57e4b75 in QApplicationPrivate::notify_helper(QObject*, QEvent*) ()

```

cquammen:
Oh, you want to remove the operator ? :smiley: I haven't added proper support for removing operators with children yet, but obviously that's needed. Thanks for the stack trace.
There are additional issues related to removing DataSources reported in #572 that also need to be addressed.

The chances are 100% that you will find issues all over the place. I hate leaving technical debt, but in this case I plan to pay it back quickly. The goal is to get something that Marcus can show next week.

cjh1:
I understand, sorry to trying unsupported operation :smile:

cquammen:
I'm glad you did.

So what is supported?

- Loading data
- Adding modules to loaded data sources
- Adding modules to child data sources and changing visualization parameters. Note that the child data source must be selected before adding a visualization module, otherwise it will be added to whichever data source was previously selected.
- Deleting visualization modules from top-level or child data sets

There is also the additional support for returning new data objects from Python scripts via a dictionary, but that is all under the covers.

cryos:
I would like to defer supporting operators on child data for now, and get this stuff working/displaying in the UI correctly. Just taking a look at the updated branch, but this is already looking quite large as a pull request so we should get it to the stage where it can be merged before doing much much so that we can see builds on different platforms, etc.

cryos:
I am rebuilding ParaView from a clean branch to see if I can get past the freeze I keep seeing locally. The main visual issue I see is that the child data is not a child of the operator, and I don't know how/don't see the vtkTable data showing up in the pipeline at all right now - is that currently manually selecting the TrivialProducer items in the spread sheet view?

Some of these could be carved off for me or Chris to work on, and don't all need to happen in this branch. My clean build of ParaView is not going well, trying a second time as it seems enabling C++11 features results in compile failures right now.

cquammen:
> The main visual issue I see is that the child data is not a child of the operator, and I don't know how/don't see the vtkTable data showing up in the pipeline at all right now

Child data as produced by BinaryThreshold should be a child of the last visualization module that is applied to the parent data set.

image

Operator results such as the table produced by ConnectedComponents should be a child of the operator.

image

> I don't know how/don't see the vtkTable data showing up in the pipeline at all right now - is that currently manually selecting the TrivialProducer items in the spread sheet view?

Yes, you have to view it manually with this branch. It would be great if you or Chris wanted to implement clicking on the table output in the pipeline viewer and bringing up a spreadsheet view to show it.

cryos:
That doesn't look right to me, and I would say that the child data should be a child of the operator too. It then becomes much clearer to me visually that the binary threshold operator produced the data, and logically seems to group the output of that operator such that if I delete the operator I fully expect the outputs of it to go away too.

cryos:
What should the connected components statistics end up being called in the spreadsheet view? Not sure if I am missing something, I see an error,

No result named "component_statistics" defined in output dictionary. (:0, )

cquammen:
Let's bring in @Hovden who suggested grouping the child data sets after the operators and modules in #392 (look toward the bottom 1/4 of the discussion page).

cryos:
Sure, I would offer up as a prime example that we want to work with several operators that have child data, which I see as being vtkTable, vtkImageData, and other derived types. They should all be treated in a fairly common fashion. Having more than one with child vtkImageData results in:

image

This would be much clearer to me if they were able to have data as a direct child of the operator that produced them. I don't think I see the utility you are aiming for in putting them below the visualization modules.

cquammen:
In that use case, I agree with you. However, @Hovden's idea as I understood it was that you could apply only one child data-producing operator, then no more operators could be applied after that. I'm happy to implement whichever way you and Rob decide.

cryos:
I think even if we wanted to limit it to that, I would want to pull that "special" final operator down and make the data a child, such that if you were to delete the operator I would fully expect the child data to also disappear, along with all its child modules, etc.

cryos:
I am not sure if this branch breaks it, but I am still seeing this error that looks like it might be related to changes made in this branch:

No result named "component_statistics" defined in output dictionary. (:0, )

cquammen:
> No result named "component_statistics" defined in output dictionary. (:0, )

Are there other warnings/errors before or after this message? Make sure CMake was run again to copy over the Python files if you haven't.

cryos:
CMake was run, but on Unix systems it is symlinking not copying anyway for the very reason that I didn't want to have to deal with ensuring CMake was run each time I changed a Python file. Full output below, but the other things don't look related.

Warning, array does not have Fortran order, making deep copy and fixing...
...done.
No result named "component_statistics" defined in output dictionary. (:0, )
Warning: In /home/marcus/src/ParaView/ParaViewCore/ClientServerCore/Rendering/vtkPVImageSliceMapper.cxx, line 207
vtkPVImageSliceMapper (0x3de6800): Failed to locate selected scalars. Will use image scalars by default.

cquammen:
Hmmm, not sure. If you could, can you see what the value of `spreadsheet` is  around line 100 of `ConnectedComponents.py`? It should be a vtkTable object.

cryos:
It matches what is in the repo, is there a simple way to check what the keys the C++ side sees are on the OperatorPython code. Right now you just error out, if failing to find the expected key I think it would be useful to dump the keys that are there.

cquammen:
Can you print out the value of `spreadsheet` in the Python script when you run it? I'll add some reporting of the keys received to help debug.

cryos:
vtkTable (0x8e32550)
  Debug: Off
  Modified Time: 1209709
  Reference Count: 1
  Registered Events: (none)
  Information: 0x4644aa0
  Data Released: False
  Global Release Data: Off
  UpdateTime: 0
  Field Data:
    Debug: Off
    Modified Time: 1209708
    Reference Count: 1
    Registered Events: (none)
    Number Of Arrays: 0
    Number Of Components: 0
    Number Of Tuples: 0
  RowData: 
    Debug: Off
    Modified Time: 1209730
    Reference Count: 1
    Registered Events: (none)
    Number Of Arrays: 3
    Array 0 name = SurfaceArea
    Array 1 name = Volume
    Array 2 name = SurfaceAreaToVolumeRatio
    Number Of Components: 3
    Number Of Tuples: 242
    Copy Tuple Flags: ( 1 1 1 1 1 0 1 1 )
    Interpolate Flags: ( 1 1 1 1 1 0 0 1 )
    Pass Through Flags: ( 1 1 1 1 1 1 1 1 )
    Scalars: (none)
    Vectors: (none)
    Normals: (none)
    TCoords: (none)
    Tensors: (none)
    GlobalIds: (none)
    PedigreeIds: (none)
    EdgeFlag: (none)

cquammen:
Thank you. That is what I would expect.

cryos:
I think I spotted it - your use of toStdString().c_str() was producing an empty string/NULL here, and so never returned an object. Everything works great when using toLatin1().data() to get a const char* name for the GetItemString call. I can finally view the table too, I usually work with RelWithDebInfo and a very recent GCC, I wonder if either of those was causing the issue. I can make a small pull request with the change.

tomviz/OperatorPython.cxx
```diff
@@ -227,27 +264,75 @@ bool OperatorPython::applyTransform(vtkDataObject* data)
...
-        PyDict_GetItemString(outputDict, resultName.c_str());
-      if (pyDataObject) {
-        vtkObjectBase* vtkobject =
-          vtkPythonUtil::GetPointerFromObject(pyDataObject, "vtkDataObject");
-        if (vtkobject) {
-          setResult(i, vtkDataObject::SafeDownCast(vtkobject));
+
+    // Results (tables, etc.)
+    for (int i = 0; i < m_resultNames.size(); ++i) {
+      const char* name = m_resultNames[i].toStdString().c_str();
```
cryos: I think you mean toLatin1 here anyway, and using

  const char* name = m_resultNames[i].toLatin1().data(); 

Gives me the results I would expect locally.

cquammen:
Ah, thanks. That would have taken a while for me to find here. I'll make the change as I am making another change in that part of the code to print out the dictionary keys if there is an error.

cryos:
Sounds good, looks like this was the root cause of all my issues not being able to see the vtkTable output - that call always gives an empty string/null here.

cquammen:
Okay. I'll get out of the habit of calling toStdString() and use toLatin1() instead. Thanks for digging into it with a debugger.

cquammen:
I see the errors with the data source in the data after ConnectedComponents is run, know what the problem is, and will address it.

Hovden:
I'm available tomorrow if you want to discuss the pipeline over video chat. I think it is a challenging ux problem.

mathturtle:
> Okay. I'll get out of the habit of calling toStdString() and use toLatin1() instead.

I'll need to break that habit too.  We may want to look through the existing code for places the wrong one is used.

cryos:
I would fix up the clang-format issues, and we can get this merged as-is and see some build results. It is getting close, and we can afford some work in progress in master right now. The more I look at it, and test it out the more convinced I am that the child vtkImageData should be a child of the operation producing it. To have something to demo next week it can appear either way, and I would like to check we are still building binaries correctly once this merges.

cquammen:
Implemented, additional commits coming. Here's what it looks like.

image

I modified ConnectedComponents to produce a child data set for the label map in addition to the table of object properties.

cquammen:
Should be set for review and merging

cryos:
Looks good to me, merging this is, we can look at other enhancements as we test this out on other platforms. The pipeline can obviously still be changed once merged, but I think this is something that I can talk about next week.

Hovden:
OK. I will test things out this evening.

PR #579, opened 2016-09-01
Add operator threading

cjh1:
Requires ParaView master built with VTK_PYTHON_FULL_THREADSAFE=ON and VTK_NO_PYTHON_THREADS=OFF


cjh1:
Ouch the conflict is horrific!

cryos:
I warned you it would be, and travis is stuck again, Brad had a way of using filter-branch or something that could help I think.

cjh1:
:tada:

cjh1:
Note the required VTK fix has now been merged into [master](https://github.com/Kitware/VTK/commit/279ff7cded9d88ba2f84a93f0c444f89a1b58584)

cjh1:
@cryos I have fixed up the issues with the reconstruction work, so this is rebased on master.

cjh1:
The required VTK change has now been merged in to [Paraview](https://gitlab.kitware.com/paraview/paraview/commit/0633490f656ef0a9c7bfa2ec4db68992348fa22f)

cjh1:
I have update the [super build as well](https://github.com/OpenChemistry/tomviz-superbuild/pull/121)

cjh1:
@cquammen @mathturtle If either of you fancy recompiling ParaView master and trying this out please go right ahead :smile: I am sure there are plenty of corner cases it would be good to catch!

cquammen:
Hmm, I'm getting this error when closing `tomviz`:

Exception KeyError: KeyError(140735305629696,) in <module 'threading' from '/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/threading.pyc'> ignored

cjh1:
@cquammen Thanks, looks like I will have to find myself a Mac to reproduce this on. Does it seem to function apart from this or is it a total bust?

cjh1:
[This stackoverflow post might be relevant](http://stackoverflow.com/questions/8774958/keyerror-in-module-threading-after-a-successful-py-test-run)

cquammen:
Sorry, should have mentioned it seems to work great. I ran a long-running FFT and was able to interact with the UI and add some visualization modules.

I'm also getting some VTK leaks reported when using the Binary Threshold transform - might be related to the exception above.

cjh1:
@cquammen Thanks, how are you running the leak detection?  ( I was pretty sure it would be leaking something! )

cquammen:
Enabling `VTK_DEBUG_LEAKS` in your ParaView build will report VTK objects that are not cleaned up properly.

tomviz/DataSource.cxx
```diff
@@ -123,6 +127,26 @@ void deserializeDataArray(const pugi::xml_node& ns, vtkDataArray* array)
...
+
+  if (m_future != nullptr) {
+    connect(m_future, SIGNAL(finished()), this, SIGNAL(finished()));
+    connect(m_future, SIGNAL(canceled()), this, SIGNAL(canceled()));
+  }
+}
+
+DataSource::ImageFuture::~ImageFuture()
+{
+  if (this->m_future != nullptr) {
```
cquammen: Minor style point that I'm not sure about: I don't think we are using `this->` in front of `m_var`. In any case, it's usage is inconsistent in this file.

cjh1: Yep, I use this-> everywhere, I can change these if its not part of the style.

tomviz/DataSource.cxx
```diff
@@ -471,35 +514,54 @@ void DataSource::operate(Operator* op)
...
-  if (op->transform(tp->GetOutputDataObject(0))) {
-    this->dataModified();
+  // If we are currently executing the pipeline, just add the operator
+  if (this->Internals->Future != nullptr &&
+      this->Internals->Future->isRunning()) {
+    this->Internals->Future->addOperator(op);
+  }
+  // We need to initiate a new run
+  else {
+    vtkDataObject* copy = this->copyData();
```
cquammen: Is copying the data here a necessary evil?

cjh1: Yes, we need to copy it so we can fall back to the previous copy if the user cancels an operation ( which is coming ...)

tomviz/DataSource.cxx
cquammen: Dunno for sure. `m_` should take the place of `this->`. I think @cryos wants to move to `m_var` instead of `this->Var`, but I defer to him.

cquammen: Note that we are not consistent in the rest of tomviz.

tomviz/DataSource.cxx
cquammen: Makes sense.

tomviz/DataSource.h
```diff
@@ -149,8 +155,21 @@ public slots:
 
 protected:
   void operate(Operator* op);
+
+  /// Reset the data output of the trivial producer to original data object.
   void resetData();
 
+  /// Set data output of trivial producer to new data object, the trivial
+  /// producer
```
cquammen: Funny comment wrapping.

tomviz/OperatorPython.cxx
```diff
@@ -377,4 +426,15 @@ EditOperatorWidget* OperatorPython::getEditorContents(QWidget* p)
 {
   return new EditPythonOperatorWidget(p, this);
 }
+
+void OperatorPython::createNewChildDataSource(
+  const QString& label, vtkSmartPointer<vtkSMProxy> producerProxy)
+{
+
+  DataSource* childDS = new DataSource(
```
cquammen: I think the VTK object leaks may be coming from here. The child `DataSource` parent should be `this`, the operator instance. Doing so will make the child data source get deleted with the `OperatorPython` is deleted.

cjh1: Your, right, just looking at it, thanks!

cjh1: Yep, I had removed 'this' as the parent as I was running into thread affinity problems, but never put it back.

cquammen:
Nice, setting the parent solves the `VTK_DEBUG_LEAKS` issue.

cquammen:
I was able to queue up a few longer-running operators while the first was running. Worked like a charm!

The only remaining issue I see is the Python exception at exit. Thoughts?

cjh1:
@cquammen I think the Python exception at exit is innocuous, but trying to recreate on bigmac to test a fix to remove it.

tomviz/PipelineWorker.cxx
```diff
@@ -0,0 +1,282 @@
...
+};
+
+class PipelineWorker::Run : public QObject
+{
+  Q_OBJECT
+
+public:
+  Run(vtkDataObject* data, QList<Operator*> operators);
+
+  // Start the next operator is in the queue
```
cquammen: Can remove "is"

tomviz/PipelineWorker.cxx
```diff
@@ -0,0 +1,282 @@
...
+  void startNextOperator();
+  /// Clear all Operators from the queue and attempts to cancel the
+  /// running Operator.
+  void cancel();
+  /// Returns true if the operator was successfully removed from the queue
+  /// before
```
cquammen: Funny wrapping. `clang-format` doesn't do a great job wrapping single-line comments.

cquammen:
> @cquammen I think the Python exception at exit is innocuous, but trying to recreate on bigmac to test a fix to remove it.

Seems to be, but I figure someone will ask about it in the future.

As far as I am concerned, this LGTM, so feel free to merge and fix the Python exception later. Nice work!

cjh1:
@cquammen Thanks for taking a look at this, I will let you know when I fix up the Python exception.

cquammen:
Actually, hang on. There is some funkiness with OperatorResults. Let me take a closer look and get you a description.

cjh1:
@cquammen Sure, no rush, not planning on merging today ...

cjh1:
@cquammen Regarding Python exception, I don't see it on bigmac, was there a particular set of steps need to recreate?

cjh1:
One difference I did notice is on bigmac we seem the be using python installed in /usr/lib/python2.7, rather than /System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7, not sure if that is significant.

cquammen:
`/usr/bin/python2.7` is a symbolic link to `System/Library/Frameworks/Python.framework/Versions/2.7/bin/python2.7`, so I'm not sure that the exception referring to the frameworks binary is a big deal.

This is what I tried:

- Launch `tomviz`
- Open small example data set
- Run Custom ITK Transform
- Close `tomviz`

Can you reproduce with these steps?

cjh1:
Nope, that works for me. Is your OS version different to bigmac?

cjh1:
Python version on bigmac is 2.7.6.

cquammen:
I have 10.11.6, Python 2.7.12. Maybe let's not worry about it for now, and I can test a nightly build at some point to see if it is just a quirk in my development environment. It doesn't seem to be a show-stopper.

cjh1:
Ok, sounds like a plan.

cquammen:
I am more concerned about the problem with setting `OperatorResults`. This problem is more non-deterministic, but if you try the following sequence a few times, you should get a crash.

- launch `tomviz`
- Open small data set
- Run Connected Components transform
- Create a new SpreadSheet view
- There should be a table data object in the "Showing" combo box.

The table data object showed up a couple times when I tried it or sometimes not at all. Once it crashed.

cjh1:
I will take a look

cjh1:
I am consistently seeing the following in the "Showing" drop down:

None
Outline1
Passthrough1
TrivialProducer3

cquammen:
Hmmm. There should be at least two more TrivialProducers after the Connected Components transform is run - one for the child data set, and one for the table result.

cjh1:
@cquammen Pardon my ignorance, but where do those values come from? I am seeing a vtkTable return from the transformation. That has columns, "SurfaceArea", "Volume" and "SurfaceAreaToVolumeRatio".

cquammen:
The Python script `ConnectedComponents.py` generates the `vtkTable`, and `OperatorPython` connects the `vtkTable` to the `vtkTrivialProducer` object whose proxy you are seeing. The high-level idea is that operators can create additional data objects, not just child `DataSources`. In the `OperatorPython` case, a Python script optionally generates one or more data object.

cjh1:
@cquammen After some quality time with gdb this may fixed the problem. It seems that calling the proxy creation/registration  on another thread ( not the UI thread ) was the problem. A signal gets lost in the conversion of a vtkCommand to a Qt signal, not sure why ( I may did a little more in the future if I have time ). Anyway, give it a try, no guarantees :smile:

cquammen:
Closer, but not quite. Now, when running the Connected Components transform, a second Trivial producer for the child data label map is produced. However, the Trivial producer for the table data operator result is not there.

cjh1:
@cquammen Is that the trivial producer create by [this call] (https://github.com/OpenChemistry/tomviz/blob/master/tomviz/OperatorPython.cxx#L286)? That is the one I thought I had fix :frowning:

cjh1:
See the following:

TrivialProducer1
Outline1
PassThrough1
TrvialProducer2
TrivialProducer3

cquammen:
Looks like there may be a logic error in setting the operator result. I'm digging in with a debugger now.

cquammen:
Okay, the `const char* name` argument is getting garbled between the signal emission and slot invocation. Making this a `QString` instead seems to fix it.

If you want to try opening up this PR permission for me to push a commit, I can get the changes I tried in this PR. See https://github.com/blog/2247-improving-collaboration-with-forks

cjh1:
@cquammen You should have permission now, thanks alot for looking at this.

cquammen:
Cool, looks like pushing my commit worked.

cquammen:
In any case, this is looking good to me now. Maybe @cryos, @mathturtle, or both should have a look at this point?

cjh1:
I have open #602 for the toLatin1(..).data() issues.

cjh1:
@cquammen Thanks for the QA and fix on this PR.

cjh1:
@cquammen Was thinking about this more and pushed another commit. Even with your change (which we definitely need) we need to ensure that the QByteArray is not destroyed, basically we have been lucky!

mathturtle:
Why not just pass the QString by value through the signal/slot connection?  The performance hit will be tiny and Qt knows how to do that without needing any workarounds to avoid objects being deleted.

cjh1:
Pass by reference is fine, the issue was that we where passing a pointer to garbage ( a freed buffer ).

tomviz/DataSource.cxx
cryos: Yeah, clang-style wouldn't do this for me, I would rather move to m_ and that should replace this->, let's not use both, I could probably do a quick grep for this->m_ to m_, at some point soon.

cryos:
Great to have this merged, there may still be some lingering issues but Chris and I spent some quality time debugging corner cases today. Let's see what the builds look like, this seems pretty solid on my machine with the last few patches applied.

PR #583, opened 2016-09-02
Remove context items for child data

### LINKED ISSUE #581: Prevent deletion of child data sets
Child data sets should not be deletable - they should only be deleted when the producing operator is deleted.

cquammen:
Fixes #581


cryos:
LGTM

PR #585, opened 2016-09-02
Add behavior to pipeline view to show table on double click

mathturtle:
This will only happen if the active view is a spreadsheet view.  We could later change it to search the current views for a spreadsheet view.

@cryos 

@cquammen Question about the `OperatorResult` class: why is `m_producerProxy` a `vtkWeakPointer` rather than a `vtkSmartPointer`?  It looks like you are doing the memory management manually.  I had to add a getter for it because I needed the proxy rather than the dataset.

### DIFF
--- a/tomviz/OperatorResult.cxx
+++ b/tomviz/OperatorResult.cxx
@@ -97,6 +97,12 @@ void OperatorResult::setDataObject(vtkDataObject* object)
   producer->SetOutput(object);
 }
 
+vtkSMSourceProxy* OperatorResult::producerProxy()
+{
+  createProxyIfNeeded();
+  return m_producerProxy;
+}
+
 void OperatorResult::createProxyIfNeeded()
 {
   if (!m_producerProxy.Get()) {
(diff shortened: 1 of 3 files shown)


cquammen:
@mathturtle I suppose `m_producerProxy` could be a `vtkSmartPointer` as the proxy is created and deleted within the `OperatorResult`. I followed the model of `DataSource::Internals->Producer`, which is a `vtkWeakPointer`. I don't have a strong opinion about it either way.

cquammen:
LGTM. Thanks for adding.

I wonder if in a follow in PR we want to open a spreadsheet view if there isn't one?

cquammen:
Feel free to merge. Looks like a rebase on master or, less to my liking, a merge from master into this branch is needed.

mathturtle:
> I wonder if in a follow in PR we want to open a spreadsheet view if there isn't one?

Probably.  I'll have to look at how ParaView does its automatic opening of views.  Given how long this has been sitting here, I may just append it to this branch.

cryos:
Looks good to me, you can add more in a follow up.

PR #588, opened 2016-09-03
Reverse connected components labeling

cquammen:
Flip the labels so that the largest component has the largest label
value and the smallest component has the smallest value. The label
value is really arbitrary, so changing the labeling in this way
shouldn't be a big deal. It does, however, make display of the largest
components via color map nicer because the largest component now has a
value most different from the background label.

### DIFF
--- a/tomviz/python/ConnectedComponents.py
+++ b/tomviz/python/ConnectedComponents.py
@@ -72,9 +72,21 @@ def transform_scalars(dataset):
 
         itk_image_data = relabel_filter.GetOutput()
         label_buffer = itk.PyBuffer[itk_output_image_type].GetArrayFromImage(itk_image_data)
+
+        # Flip the labels so that the largest component has the highest label
+        # value, e.g., the labeling ordering by size goes from [1, 2, ... N] to
+        # [N, N-1, N-2, ..., 1]. Note that zero is the background value, so we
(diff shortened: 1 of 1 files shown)


cquammen:
Mind reviewing @cjh1 @mathturtle @cryos ? Thanks.

cquammen:
First implementation was terribly memory inefficient. Now uses in-place operations instead.

tomviz/python/ConnectedComponents.py
```diff
@@ -72,9 +72,23 @@ def transform_scalars(dataset):
...
+        maximum = np.max(label_buffer)
+
+        # Try more memory-efficient approach
+        gt_zero = label_buffer > 0
+        label_buffer[gt_zero] = minimum - label_buffer[gt_zero] + maximum
+
         label_map_data_set = vtk.vtkImageData()
         label_map_data_set.CopyStructure(dataset)
 
+        #utils.set_label_map(label_map_data_set, flipped_label_buffer)
```
cjh1: Should we remove this?

cquammen: Indeed. Thanks for catching.

cjh1:
Had to break out ipython to understand the numpy array operations but LGTM :smile:

PR #590, opened 2016-09-06
Fix multi-clients warning at startup

cjh1:
Fixes the following warning:
```
This code may not work in multi-clients mode (:0, )
```


cjh1:
@cryos Looks like some code got commented out in [the clang formatting PR](https://github.com/OpenChemistry/tomviz/blame/master/tomviz/Behaviors.cxx#L88)

cquammen:
LGTM. Tested locally, warning goes away.

PR #604, opened 2016-09-16
When double clicking a table, open a spreadsheet if not available

mathturtle:
This is a follow up to #585 that extends it to open a new spreadsheet view if the current view is not a spreadsheet view.


mathturtle:
@cryos

cryos:
This works great here, and works when merged into master. I think a simple rebase is all you need (I didn't get any merge conflicts).

PR #605, opened 2016-09-16
Charts

cryos:


cryos:
To be more useful this needs some additional features, but it shows some of the things I would like to see. Main to do is the addition of column visibility selection to the interface, and probably some of the axis scaling stuff too.

PR #610, opened 2016-09-22
Add support for canceling python operators

cjh1:
This PR provides the infrastructure to allow Python operators to be cancelled. I have modified the implementation of [InvertData.py](https://github.com/cjh1/tomviz/blob/python-cancel/tomviz/python/InvertData.py) and [Square_Root_Data.py](https://github.com/cjh1/tomviz/blob/python-cancel/tomviz/python/Square_Root_Data.py) to demonstrate how a cancelable operator can be implemented. I have also added [Operators.md](https://github.com/cjh1/tomviz/blob/python-cancel/tomviz/python/Operators.md) which provides some details of how to implement operators using the new interface. 

Note that this branch introduced a submodule ( for pybind11 ) so you will need run the following:

```
git submodule init && git submodule update
```

### DIFF
--- a/tests/cxx/CMakeLists.txt
+++ b/tests/cxx/CMakeLists.txt
@@ -1,13 +1,12 @@
 include(CxxTests.cmake)
 
 find_package(GTest REQUIRED)
-include_directories(SYSTEM
-  ${QtCore_INCLUDE_DIRS})
 
-include_directories(SYSTEM ${PARAVIEW_INCLUDE_DIRS})
-include_directories(${PROJECT_SOURCE_DIR}/tomviz)
-include_directories(SYSTEM
+include_directories(SYSTEM $
+  {QtCore_INCLUDE_DIRS}
+  ${PARAVIEW_INCLUDE_DIRS}
   ${GTEST_INCLUDE_DIRS})
+include_directories(${PROJECT_SOURCE_DIR}/tomviz)
 
 include(CheckIncludeFileCXX)
 include(CheckCXXSymbolExists)
--- a/tests/cxx/CMakeLists.txt
+++ b/tests/cxx/CMakeLists.txt
@@ -1,0 +1,55 @@
+include(CxxTests.cmake)
+
+find_package(GTest REQUIRED)
+include_directories(SYSTEM
+  ${QtCore_INCLUDE_DIRS})
+
+include_directories(SYSTEM ${PARAVIEW_INCLUDE_DIRS})
+include_directories(${PROJECT_SOURCE_DIR}/tomviz)
+include_directories(SYSTEM
+  ${GTEST_INCLUDE_DIRS})
+
+include(CheckIncludeFileCXX)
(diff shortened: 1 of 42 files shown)


cjh1:
@cryos @mathturtle @cquammen This branch is now ready for review. @cquammen this may be of particular interest to you as you implement new ITK operators, ping me if you would like to discuss further.

cquammen:
Compiled and ran locally, good user experience. Is it possible to tell a cancelled transform to execute again?

cjh1:
@cquammen Currently no, you need to trigger the pipeline in some other way (remove/add operator). This is in plan for a future PR.

cquammen:
LGTM. Just a branch conflict to deal with before merging.

Nice work!

cjh1:
Fixed merge conflict ( my own fault this time :frowning: ) and failing test. Thanks @cquammen for the QA

mathturtle:
> The dashboards may need some adjustment in order to deal with Tomviz submodules too...

Not really.  Buildbot already does a `git submodule update --init` as part of its checkout step.  If we ever change where the submodule points, we will have to make some buildbot changes to force a `git submodule sync`.  I can make those changes now if we want, we will just get a few useless steps on the buildbot interface since we don't do testing for PRs (the buildbot scripts have some code to handle searching a user's fork of a submodule for the commit pointed at by their branch).

cryos:
Build issues in testing - make the Python version a cache variable, tell it to use C++11, and a clean build gives me

/home/marcus/src/tomviz/tomviz/WelcomeDialog.cxx:18:30: fatal error: ui_WelcomeDialog.h: No such file or directory
 #include "ui_WelcomeDialog.h"

So it looks like an ordering issue with generating that file.

tests/cxx/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
+include(CxxTests.cmake)
+
+find_package(GTest REQUIRED)
+include_directories(SYSTEM
+  ${QtCore_INCLUDE_DIRS})
+
+include_directories(SYSTEM ${PARAVIEW_INCLUDE_DIRS})
+include_directories(${PROJECT_SOURCE_DIR}/tomviz)
+include_directories(SYSTEM
```
cjh1: I have combined these

cjh1:
@cryos I have fixed the ui_WelcomeDialog issue, seems calling qt5_wrap_ui twice doesn't seem to work, anyway I combined the two calls. I wonder is we should try using [CMAKE_AUTOUIC](https://cmake.org/cmake/help/v3.0/variable/CMAKE_AUTOUIC.html)?

cryos:
> @cryos I have fixed the ui_WelcomeDialog issue, seems calling qt5_wrap_ui twice doesn't seem to work, anyway I combined the two calls. I wonder if we should try using CMAKE_AUTOUIC?

Might be an idea, but let's put that in a follow up if this works for now. I didn't know multiple calls would fail, seems kind of lame.

PR #613, opened 2016-09-23
Custom ui other modules

mathturtle:
@cryos This is my branch to convert the other modules away from a pqProxiesWidget-based UI.  I figured out what I was missing with the outline's color button, so here is the branch.

### DIFF
--- a/tomviz/ModuleThreshold.cxx
+++ b/tomviz/ModuleThreshold.cxx
@@ -16,18 +16,25 @@
 #include "ModuleThreshold.h"
 
 #include "DataSource.h"
+#include "DoubleSliderWidget.h"
 #include "Utilities.h"
+#include "pqDoubleRangeSliderPropertyWidget.h"
 #include "pqProxiesWidget.h"
+#include "pqSignalAdaptors.h"
+#include "pqStringVectorPropertyWidget.h"
 #include "vtkNew.h"
 #include "vtkSMPVRepresentationProxy.h"
 #include "vtkSMParaViewPipelineControllerWithRendering.h"
 #include "vtkSMPropertyHelper.h"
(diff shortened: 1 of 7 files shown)


cryos:
Just to confirm this does not crash in master, although even on the small data set it does force my machine to swap, do we want that button - what is its main purpose? Everything else looks good, may want to tweak the layouts a little in the future but I think we can do that in follow up commits.

mathturtle:
I tracked down the crash and I've pushed a new version that fixes it.

tomviz/ModuleThreshold.cxx
```diff
@@ -140,26 +147,79 @@ void ModuleThreshold::addToPanel(QWidget* panel)
...
+      this->ThresholdFilter->GetProperty("SelectInputScalars"),
+      this->ThresholdFilter.Get());
+  layout->addWidget(arraySelection);
 
-  proxiesWidget->addProxy(this->ThresholdFilter, "Threshold", fprops, true);
+  pqDoubleRangeSliderPropertyWidget* range =
+    new pqDoubleRangeSliderPropertyWidget(
+      this->ThresholdFilter.Get(),
+      this->ThresholdFilter->GetProperty("ThresholdBetween"));
+  range->setProperty(this->ThresholdFilter->GetProperty("ThresholdBetween"));
```
mathturtle: This is the line that fixes the crash.  After discussion with Utkarsh, I'm going to move this line into the constructor of pqDoubleRangeSliderPropertyWidget since it is strange to have to pass the property to both.  So this will be fixed by a ParaView update... and I'll take this out.  We can merge now or hold off for the ParaView change.

cryos:
I think we should merge now if you have the issue fixed.

Issue #473, opened 2016-07-12
Add single color selection for isosurface contour


ElliotPadgett:
It would be very convenient to have an interface for selecting a single color for an isosurface rendering.  I think this is most often the preferred method for setting up an isosurface. @mathturtle

ElliotPadgett:
This looks good.  My only comment before closing is that it would be nice if the "Use Solid Color" check box was next to the control to pick the color.  Right now one is at the top and the other at the bottom of the properties panel.

ElliotPadgett:
Also, this may be connected to #565 but the color selected for the contour doesn't get saved and restored in state files.  It remembers that you checked the solid color box (and the opacity) but it reverts the color to white.

ElliotPadgett:
Solid color for contour is now saved, although the icon reverts to showing white instead of the actual color.

cryos:
We should fix that before closing this issue.

mathturtle:
@ElliotPadgett Can you see if #731 fixes this?

yijiang1:
Yes this is fixed by #731 .

PR #614, opened 2016-09-23
Contour support solid color

mathturtle:
This is based on #613 right now since I was modifying the UI of modules anyway.

This is for #473, which came from a discussion with @ElliotPadgett while visiting Cornell.  For contours it is more useful to let the user set a solid color for the contour.  Because most data in tomviz only has one data array and the contour is by definition an isosurface on that array, the contour will be a solid color anyway.  Depending on the colormap, the contour can be very difficult to see against the background and it is more helpful to just let the user select the color they want it to be.

@cryos

### DIFF
--- a/tomviz/ModuleContour.cxx
+++ b/tomviz/ModuleContour.cxx
@@ -19,6 +19,7 @@
 #include "DoubleSliderWidget.h"
 #include "Utilities.h"
 
+#include "pqColorChooserButton.h"
 #include "pqPropertyLinks.h"
 #include "pqSignalAdaptors.h"
 #include "pqWidgetRangeDomain.h"
@@ -119,6 +120,9 @@ bool ModuleContour::initialize(DataSource* data, vtkSMViewProxy* vtkView)
   this->Internals->NonLabelMapArrayName =
     std::string(colorArrayHelper.GetInputArrayNameToProcess());
 
+  vtkSMPropertyHelper colorHelper(this->ContourRepresentation, "DiffuseColor");
+  double white[3] = { 1.0, 1.0, 1.0 };
+  colorHelper.Set(white, 3);
   // use proper color map.
   this->updateColorMap();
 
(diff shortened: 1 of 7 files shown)


Hovden:
Elliot was just asking about this! Thanks.

cryos:
This looks good, it depends on the other pull request that introduces a new crash - at least here for me. It would be good to get that figured out or cherry pick this commit.

mathturtle:
If I cherry-pick it, it conflicts with the other branch... and with the new merge rules I'd have to rebase one of them anyway.  So I just rebased on the fixed version of the other branch.

Hovden:
Shouldn't this feature have a checkbox to enable it or something? Right now only uses a solid color.

Also, it has the "Separate Color Map" checkbox that does nothing.

cryos:
Good points, we should make it a check box to enable the feature, and then the separate color map check box would be useful again.

mathturtle:
> Shouldn't this feature have a checkbox to enable it or something? Right now only uses a solid color.

I wondered about that.  This was my general thought process about it:

>Isocontours are by definition solid color anyway, even if you use a color map.
>
>We don't support multiple contours from the same contour module, so there will be at most one color used from the color map.
>
>So why support the whole color map?  I thought about removing the separate color map for this module, but @cquammen's color by label map stuff was still in there and I wasn't sure about removing that.

I can add it back in easily enough.  @cquammen is your color by label map stuff still used at all with the changes to segmentation?

cryos:
I think the visual correspondence with the histogram can be important, and should be preserved. We do support several contour modules in a single view, and being able to see approximate isovalue from the color map has value.

Hovden:
Also, the wireframe surface color does not change color in this after pull request.

mathturtle:
@Hovden #626 makes this optional.  I only just saw your comment about the wireframe, and I'm looking at it now.

mathturtle:
Added a commit to #626 to fix the wireframe.

PR #615, opened 2016-09-26
Use the QAction data member to store module name

cryos:
I was seeing an instance where & was being inserted into the text,
and the data member is the preferred place to store data such as
a uniquely identifiable module name string anyway. This fixes the
issue I was seeing with the & not matching.


cryos:
Thanks @mathturtle, I didn't need you all to review but the novelty of assigning more than one person got the better of me. I will get this merged, it is a pretty simple patch, but is a pattern we should adopt where the data member stores programmatic identifiers as UI strings are subject to manipulation, translation, etc.

PR #617, opened 2016-09-27
Actually set the JSON description string

cquammen:
Embarassingly, this was not being done in the setJSONDescription()
member function.


cquammen:
@mathturtle Mind reviewing? This should be an easy one.

mathturtle:
LGTM

PR #626, opened 2016-09-27
Contour make solid color optional... and solid color fixes

mathturtle:
@Hovden @cryos 

I also remembered to add the code to save the solid color in the state file.


Hovden:
LGTM, would like to merge—will be using these features today.

PR #627, opened 2016-09-27
Save JSON description of python operators to state file

cquammen:
Doing so is required to properly restore child data sets and operator results.

### DIFF
--- a/tomviz/OperatorPython.cxx
+++ b/tomviz/OperatorPython.cxx
@@ -445,13 +445,17 @@ Operator* OperatorPython::clone() const
 
 bool OperatorPython::serialize(pugi::xml_node& ns) const
 {
+  ns.append_attribute("json_description")
+    .set_value(this->JSONDescription().toLatin1().data());
   ns.append_attribute("label").set_value(this->label().toLatin1().data());
   ns.append_attribute("script").set_value(this->script().toLatin1().data());
+
   return true;
 }
 
 bool OperatorPython::deserialize(const pugi::xml_node& ns)
 {
+  this->setJSONDescription(ns.attribute("json_description").as_string());
   this->setLabel(ns.attribute("label").as_string());
   this->setScript(ns.attribute("script").as_string());
   return true;


cquammen:
Python operators store their python code in the state file. This branch also stores the JSON description of the script's inputs and outputs in the state file.

This feels wrong, but because the python code can be modified from the default code (and indeed it is for any python script that defines input parameters), shipping the JSON description around in the state file makes some sense.

cjh1:
@cquammen Isn't the python code being store in the state file for the very reason that the python code in the installation could be modified?

cquammen:
That would make sense, but I haven't heard positively that it is the case. @cryos or @mathturtle ?

mathturtle:
If the user is modifying the python code in their installation then all bets are off for the normal operation of tomviz... We read those files and do a search/replace to put in the parameters of the script.  Then we save the result of the search/replace as the script for that operator.

That is just the initial version though.  If the user opens the operator editor dialog for the operator, they are presented with the python code that they are free to edit.  This is the main reason to save the code, the user can edit it however they want and expect that to be loaded.  We have talked about using the JSON info to re-create the initial dialog rather than showing the script by default, but I don't know if this is implemented yet (I haven't followed the JSON stuff very closely).  And the ability to arbitrarily edit the script would not be removed.

I'm not sure if saving the JSON makes sense or not, I'll have to think about it more.

cquammen:
> We have talked about using the JSON info to re-create the initial dialog rather than showing the script by default, but I don't know if this is implemented yet (I haven't followed the JSON stuff very closely).

This has not yet been implemented.

> I'm not sure if saving the JSON makes sense or not, I'll have to think about it more.

One alternative implementation I can think of is saving the name of the initial python source file to the state file and then using that to look up which JSON description to set in the operator. However, one may modify the script script substantially, so linking it to the original python source doesn't necessarily work. We would really need a "frozen" python operator option that disallows modification of the python source and uses the JSON description to set parameters.

I think saving the JSON in the state file, as this branch does, is the safest way to proceed for now.

tomviz/OperatorPython.cxx
```diff
@@ -399,13 +399,21 @@ Operator* OperatorPython::clone() const
...
+    pugi::xml_node childNode = ns.append_child("DataSource");
+    this->childDataSource()->serialize(childNode);
+  }
+
   return true;
 }
 
 bool OperatorPython::deserialize(const pugi::xml_node& ns)
 {
+  this->setJSONDescription(ns.attribute("json_description").as_string());
```
mathturtle: I was worried this would break all older state files, but after testing it, it seems that it doesn't.  I guess it just defaults to empty string and that is handled fine elsewhere.

tomviz/OperatorPython.cxx
```diff
@@ -399,13 +399,21 @@ Operator* OperatorPython::clone() const
 
 bool OperatorPython::serialize(pugi::xml_node& ns) const
 {
+  ns.append_attribute("json_description")
+    .set_value(this->JSONDescription().toLatin1().data());
   ns.append_attribute("label").set_value(this->label().toLatin1().data());
   ns.append_attribute("script").set_value(this->script().toLatin1().data());
+  if (this->hasChildDataSource() && this->childDataSource()) {
+    pugi::xml_node childNode = ns.append_child("DataSource");
+    this->childDataSource()->serialize(childNode);
```
mathturtle: Are these added to the ModuleManager's list of DataSources?  If so, they will be serialized twice.  Does that cause problems?  Will the serialized modules be deserialized in the right place?

mathturtle: Answering myself:  The module manager doesn't write out the child datasets (I think on purpose).  But any visualization modules on the child dataset are still lost (I just tested this).

mathturtle:
> One alternative implementation I can think of is saving the name of the initial python source file

No, that would be far more fragile.  Anyone updating the python source files/ json files would have to think about how their changes would affect all state files written in the past.  The current implementation is much more future-proof.  ...So after thinking about it some more, I think the current approach is fine.  I did find a bug when testing your branch, but once that is fixed it should be good to merge.

cryos:
We want to support people modifying the Python scripts, and those modified scripts should be restored. We could introduce a concept of a pristine script, and simply use those, but what if they were modified in a future release? I think we can look at getting rid of the search and replace by supplying an object with user-supplied arguments, and then an MD5 could be used to verify equality (or SHA-256 if we want greater assurance).

tomviz/OperatorPython.cxx
cquammen: Hmm, I'm going to need to figure out an elegant way to do this.

cquammen:
I suggest we table the "pristine" or "frozen" Python concept for now. If it is a use case that comes up, we can look into supporting it then.

cquammen:
@mathturtle Restoring child data sets and modules properly is harder than I thought. This branch does not solve it, but it does include code that is required. Are we in agreement that the JSON, which specifies inputs and outputs in the Python scripts, should be stored in the state file? If so, I'd like to merge this, then tackle the state file work in another branch.

mathturtle:
@cquammen That makes sense.  This branch is at least an improvement on how it worked before.  Yes, we are in agreement about the JSON being included in the state file.  Is my review to request changes blocking the merge?

cquammen:
I don't think it is. I'll try merging.

PR #629, opened 2016-09-28
FIx generate random particles

### LINKED ISSUE #618: Crashes when running "Sample Data > Generate Random Particles"
Click "Sample Data > Generate Random Particles"
Then click OK.
Crashes

cjh1:
Ensure GIL before calling Python API. Fixes #618

### DIFF
--- a/tomviz/Utilities.cxx
+++ b/tomviz/Utilities.cxx
@@ -14,6 +14,7 @@
 
 ******************************************************************************/
 #include "Utilities.h"
+#include "vtkPython.h" // must be first
 
 #include "DataSource.h"
 #include "pqAnimationCue.h"
@@ -32,7 +33,6 @@
 #include "vtkPVXMLElement.h"
 #include "vtkPVXMLParser.h"
 #include "vtkPoints.h"
-#include "vtkPython.h"
 #include "vtkRenderer.h"
 #include "vtkSMNamedPropertyIterator.h"
 #include "vtkSMPropertyHelper.h"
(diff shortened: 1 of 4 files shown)


tomviz/Utilities.cxx
```diff
@@ -32,6 +32,7 @@
 #include "vtkPVXMLElement.h"
 #include "vtkPVXMLParser.h"
 #include "vtkPoints.h"
+#include "vtkPython.h"
```
cquammen: `vtkPython.h` should always be included first. From the comments in `vtkPython.h`:

```
/* We used to try to #undef feature macros that Python.h defines
to avoid re-definition warnings.  However, such warnings usually
indicate a violation of Python's documented inclusion policy:

 "Since Python may define some pre-processor definitions which
  affect the standard headers on some systems, you must include
  Python.h before any standard headers are included."
 (http://docs.python.org/c-api/intro.html#include-files)

To avoid re-definitions warnings, ensure "vtkPython.h" is included
before _any_ headers that define feature macros, whether or not
they are system headers.  Do NOT add any #undef lines here.  */
```

cjh1: Thanks for catching this, good to know!

cryos:
I would suggest leaving a blank line after the include of vtkPython.h, looks good to me.

cquammen:
Ditto what @cryos said, otherwise `clang-format` will reorder it against your wishes.

cjh1:
@cquammen I have added the space, I think we are ready for merge?

cquammen:
LGTM

Issue #631, opened 2016-09-29
Missing toolbar buttons


cquammen:
As of 8bcac91386ea1f3e1b9f6db7c7521b9111e9515f, there are some missing toolbar button icons on Mac OS X:

![image](https://cloud.githubusercontent.com/assets/360056/18960347/683c6a96-8637-11e6-9fa6-e450a7350617.png)

Reverting to 513d551d6b210fd68faddd79af646189a953299f, the icons are there.

![image](https://cloud.githubusercontent.com/assets/360056/18960540/1a079476-8638-11e6-9b9d-05708819e4f3.png)

@cjh1 I think this occurred in the changes in #610.

cjh1:
@cquammen I am seeing the same on linux, I will look into it.

cquammen:
All fixed now.

PR #632, opened 2016-09-29
Add resources to executable rather than static lib

cjh1:
They get stripped from the library so add them to the execuatable.


cjh1:
@cquammen This should fix #631, thanks @cryos for helping me track this down.

cquammen:
Nice, that fixes it @cjh1.

Issue #635, opened 2016-09-29
ITK segmentation doesn't work on sample datasets


yijiang1:
I get this error when doing segmentation on sample datasets:

Exception encountered while running BinaryThreshold
"itkTemplate : No template [<class 'itkImagePython.itkImageD3'>] for the itk::PyBuffer class"

cquammen:
Is this a local build or nightly binary?

yijiang1:
My local build.

cquammen:
The nightly binary downloaded yesterday appears to work fine and @cjh1 successfully tested the new Otsu thresholding transform on his local build, so I think this is a build issue.

Did you compile according to the directions posted at https://github.com/OpenChemistry/tomviz/blob/master/BUILDING.md? For convenience, I've reposted them here:

```
cd ..
git clone git://itk.org/ITK.git
cd ITK
git checkout v4.9.0
cd ..
mkdir itk-build
cd itk-build
cmake -DCMAKE_BUILD_TYPE:STRING=RelWithDebInfo \
  -DITK_LEGACY_REMOVE:BOOL=ON \
  -DITK_LEGACY_SILENT:BOOL=ON \
  -DITK_USE_FFTWD:BOOL=ON \
  -DITK_USE_FFTWF:BOOL=ON \
  -DModule_BridgeNumPy:BOOL=ON \
  -DBUILD_TESTING:BOOL=OFF \
  -DITK_WRAP_PYTHON:BOOL=ON \
  -DBUILD_EXAMPLES:BOOL=OFF \
  -DBUILD_SHARED_LIBS:BOOL=ON \
  ../ITK
cmake --build .
```

Note that I have only found ITK git tag v4.9.0 to work reliably.

yijiang1:
Yes I built my ITK following the instruction above.
Sorry I just realize that my description is not clear enough. My segmentation actually works well on the data imported from files. It only fails on the data generated from the "Sample Data" menu.

cquammen:
Ah, thanks, that's a big help 😸

On my system, I actually get a crash with all the data sets in the "Sample Data" menu. The crash is at `OperatorPython.cxx:111-112`.

```
this->Internals->InternalModule.TakeReference(
      PyImport_ImportModule("tomviz._internal"));
```

Interestingly, the sample tif files load work just fine. The main difference between the tif files and generated sample data is that the sample data has a `double` voxel type whereas the loaded data sets are `unsigned char` or `unsigned int`.

[...]

PR #638, opened 2016-09-30
Ensure the reference count is incremented before "steal"

cjh1:
The reference count need to be incremented before PyTuple_SET_ITEM "steals" a reference.


cjh1:
My fix #635 :smile:

cquammen:
Wow, nice find. Tested locally, solves the crash.

LGTM

mathturtle:
Nice catch.  I must have forgotten that PyTuple_set_item steals the reference when I wrote that...

LGTM.

PR #640, opened 2016-10-01
Interface builder

cquammen:
Infrastructure for building Operator UI for settings parameters in preparation for adding a number of new ITK-based filters.

PR #641, opened 2016-10-01
Morphology operators

cquammen:
Added a number of binary morphology operators. Note that this is a follow-on to PR #640.

### DIFF
--- a/tomviz/python/BinaryDilate.py
+++ b/tomviz/python/BinaryDilate.py
@@ -45,7 +45,7 @@ def transform_scalars(dataset):
 
         label_buffer = itk.PyBuffer[itk_input_image_type].GetArrayFromImage(itk_image_data)
 
-        utils.set_label_map(dataset, label_buffer)
+        utils.set_array(dataset, label_buffer)
     except Exception as exc:
         print("Exception encountered while running BinaryDilate")
         print(exc)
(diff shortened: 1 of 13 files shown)


cquammen:
@cryos @cjh1 @ElliotPadgett This should be ready for review.

To test:
* Launch Tomviz
* Open small data set
* Crop data set to something like a 50x50x50 cube for quicker testing
* Run Binary Threshold
* Select the child data set produced by Binary Threshold
* Create a Slice plane to visualize the binary label map
* Run Binary Dilate, default parameters. Zoom in to the data (the Slice gets translated; separate bug). The segmented object will be enlarged.
* Delete Binary Dilate. Try the other morphology filters (Binary Erode, Binary Open, Binary Close)

cjh1:
@cquammen I am happy to review this, but not sure I am using it right, an chance of doing a quick hangout to walk me through it?

tomviz/python/BinaryDilate.py
```diff
@@ -0,0 +1,54 @@
...
+        dilate_filter.Update()
+        itk_image_data = dilate_filter.GetOutput()
+
+        PyBufferType = itk.PyBuffer[itk_input_image_type]
+        label_buffer = PyBufferType.GetArrayFromImage(itk_image_data)
+
+        utils.set_array(dataset, label_buffer)
+    except Exception as exc:
+        print("Exception encountered while running BinaryDilate")
+        print(exc)
```
cjh1: Show we re-raise these exceptions so we get a traceback ( once its fixed ). print(exc) with just print the exception type and message.

cquammen:
Thanks for the review. I'll rebase and merge.

Issue #608, opened 2016-09-20
Set Units and Size (Data Properties Menu)


Hovden:
It would be great to complete the Set Units and Size feature in tomviz.

<img width="247" alt="screen shot 2016-09-20 at 5 07 25 pm" src="https://cloud.githubusercontent.com/assets/6197997/18688982/c78fa146-7f54-11e6-89f5-f434f9714618.png">

Under "Units and Size" it should display something like:
(100 x 200 x 235) nm

The change units button can be an editable box like the x,y,z values.

We will also want to show the units and size in the Axis Grid (which I think should be part of the bounding box visualization). I will make this part a new issue.

yijiang1:
> We will also want to show the units and size in the Axis Grid (which I think should be part of the bounding box visualization). I will make this part a new issue.

@Hovden Remember to make a new issue if you haven't.

yijiang1:
@Hovden When someone edits the values in X/Y/Z box, should we only change the label in the Axis Grid or resize the data set?

Hovden:
Either way could work.

Scenario 1:
User specifies aspect ratio of voxels (x,y,z) for the dataset
and also specifies the dimensions (e.g. nm) for x,y,z

Scenario 2:
The aspect ratio is pulled automatically from the x,y,z dimensions set.

We should pick the scenario that is easiest to program.

yijiang1:
Right now editing values in X/Y/Z box will "resize" the dataset:
<img width="1051" alt="screen shot 2017-01-20 at 10 21 05 am" src="https://cloud.githubusercontent.com/assets/13088588/22154899/5b37ad7a-defb-11e6-8bae-02bdab3430f3.png">
<img width="1047" alt="screen shot 2017-01-20 at 10 21 54 am" src="https://cloud.githubusercontent.com/assets/13088588/22154903/5ca0f950-defb-11e6-952e-9ea92adce29b.png">

Is there a way to edit the labels of the axis grid?

yijiang1:
Never mind, I realize my confusion is due to a bug. I'll make a new issue for that.

PR #643, opened 2016-10-04
New UI for setting Units and Size

yijiang1:
This is for #608 @Hovden 
The UI now looks like
screen shot 2016-10-04 at 9 40 27 am


cquammen:
LGTM.

PR #646, opened 2016-10-04
Use new full scene antialiasing support

cryos:
This enables the new full scene antialiasing added in VTK/ParaView,
improving the general look of rendered views in a post render step.


cquammen:
Tested, works nicely.

PR #649, opened 2016-10-04
Ensure traceback is displayed when a python exception is raised

### LINKED ISSUE #599: display python errors instead of crashing the program
Whenever there is an error in python transform, tomviz should simply display the error message instead of crashing the entire program...

cjh1:
Fixes #599. The traceback was not being shown because we were doing the error handling in a separate locking of the GIL.


cquammen:
Tested locally, tracebacks are now reported as expected.

LGTM

PR #652, opened 2016-10-05
Minor improvement to import error message

cquammen:


cquammen:
Nice, you reviewed before I could even ask for one, @cjh1!

PR #657, opened 2016-10-05
Install python files from source tree

cjh1:


cjh1:
@mathturtle @cryos  PTAL

cjh1:
I think I can simplify this a little ... give me a moment ..

cjh1:
ok, I think that is better

mathturtle:
LGTM

PR #665, opened 2016-10-07
Praxis fixes

mathturtle:
@cjh1 @cquammen 

With these changes it should build on Windows again.  Both changes are workarounds for bugs in VTK that I will be creating an MR for.

### DIFF
--- a/tomviz/InterfaceBuilder.cxx
+++ b/tomviz/InterfaceBuilder.cxx
@@ -25,6 +25,13 @@
 #include <QSpinBox>
 #include <QWidget>
 
+// Has to be included before vtk_jsoncpp.h
+// vtk_jsoncpp.h needs VTK_BUILD_SHARED_LIBS to be set correctly in order to
+// define the export macros for MSVC right. Otherwise you don't get the
+// dllimport on all the jsoncpp symbols. This was causing duplicate
+// symbol errors.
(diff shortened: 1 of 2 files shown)


tomviz/InterfaceBuilder.cxx
```diff
@@ -25,6 +25,9 @@
 #include <QSpinBox>
 #include <QWidget>
 
+// Has to be included before vtk_jsoncpp.h
+#include "vtkConfigure.h"
+
 #include "vtk_jsoncpp.h"
```
cjh1: Why?

mathturtle: `vtk_jsoncpp.h` needs `VTK_BUILD_SHARED_LIBS` to be set correctly in order to define the export macros for MSVC right.  Otherwise you don't get the `dllimport` on all the jsoncpp symbols.  This was the cause for the duplicate symbol errors.  @robertmaynard helped me track it down.

It is only a problem if `vtk_jsoncpp.h` is the first vtk header included (or in this case the only one).

cjh1: Can we add that explanation to the comment block, thanks.

mathturtle: Done

Issue #633, opened 2016-09-29
Expose advanced features in the volume rendering


cryos:
The default rendering using the volume looks like:

![image](https://cloud.githubusercontent.com/assets/42214/18964410/c0fec5c0-8646-11e6-99ab-f847b8380835.png)

Enabling lighting can offer this effect (parameters may need tweaking):

![image](https://cloud.githubusercontent.com/assets/42214/18964431/dc558be2-8646-11e6-98ee-c52e8fd9ffeb.png)

and maximum intensity can show the brightest parts of the interior (these are all using the same opacity/color map)

![image](https://cloud.githubusercontent.com/assets/42214/18964459/fad2717a-8646-11e6-94ee-6dd8d62574c2.png)

Hovden:
Very cool.

bdalevin:
I second that. Very cool.

Hovden:
Bumping this—would love to incorporate this into my talk on the variability of tomography visualization.

cryos:
I cleaned it up more, and pushed a branch. You should be able to try it out, I need to look at save/restore before it is really ready to be merged, and there are likely some corner cases I overlooked.

cryos:
There is certainly more to do, but I think the initial merge request will let you test out the functionality, and we can flesh out the exposed properties in future pull requests.

cryos:
Closing this issue. we can open new issues to track additional features, bugs, etc.

PR #671, opened 2016-10-10
Expose some additional volume rendering features

cryos:
This commit switches from the ParaView volume renderer to a locally
constructed one. This gives full access to the volume renderer,
properties, and mapper. So far I have just exposed a couple of the
features, and plan to add more in future commits.

I still need to add support for serialize/deserialize. The proposed
changes are being tracked in issue #633.

### DIFF
--- a/tomviz/ModuleVolume.cxx
+++ b/tomviz/ModuleVolume.cxx
@@ -45,6 +45,9 @@
 
 namespace tomviz {
 
+using pugi::xml_attribute;
+using pugi::xml_node;
+
 ModuleVolume::ModuleVolume(QObject* parentObject) : Superclass(parentObject)
 {
 }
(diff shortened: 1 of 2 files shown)


cquammen:
Looks nice! I think the lighting adds a lot in terms of shape perception for certain data sets. One needs to set a fairly sharp rise (close to step function) in the opacity function for the best effect, otherwise the shading can get blurred out in the semi-opaque transition. Effectiveness decreases when the opacity function is closer to a constant function.

Here is volume rendering without lighting for the small data set:

image

Here is volume rendering with lighting:

image

And for fun, here is maximum intensity:

image

Maximum intensity provides the best effect when interactively rotating the volume.

Only two comments at this point:

* The"Enable lighting" check box is not set properly when clicking a different item in the pipeline and then clicking on the Volume module.
* It would be great to make the "Enable lighting" checkbox visible or editable only when "Maximum intensity" is unchecked.

cryos:
Agreed, and they are things I had noticed, this was more to let Rob take a look. I will take a look later, and would like to add more of the lighting parameters to the panel too.

Side note - I feel like GitHub is marking branches out of date very aggressively or am I just imagining it? Maybe there is a lot more churn in our code base right now.

cquammen:
For this project, GitHub appears to disallow merging of any branches that do not include HEAD on master. Maybe it's controlled by a project setting? No setting visible to me indicates that only fast-forward merges are allowed.

I personally don't care for the history that results when merging master into branches, which is the one way GitHub provides for this project for out-of-date branches.

cryos:
Found it, and unchecked the box. I also added the requirement that pull requests must be reviewed while I was there. I think it fits well with what we want, and ensures everything has at least one reviewer.

Hovden:
@yijiang1 Do you mind taking a look?

Volume rendering enhancements are really exciting for viz. You mentioned controlling lighting, which is something I have wondered about for a while. Would this lighting be specific only to the volume render or to other renders (contour, orthoslice, etc.)?

cryos:
The lighting would be specific to the volume renderer, we could probably look at making some kind of linked set of properties, but that would be a little more work (and I am not sure everything translates perfectly).

yijiang1:
Visualization looks good to me.
I do found a minor UI bug:
1. In the property panel, check "Enable lighting" box.
screen shot 2016-10-11 at 12 22 07 pm
2. Switch property panel to some other rendering mode (orthogonal slice for instance).
3. Go back to the volume rending property panel, now the box is unchecked.

Same problem for "Maximum intensity" box.

cryos:
Yes, Cory mentioned the same issues above, part of the clean up I need to do, along with saving state, and disabling the lighting box when max intensity is enabled.

Hovden:
I would like to merge and use this feature.

cryos:
I should have time to finish it up this afternoon. It will break state files, but then there was not much to save from a state file before, so if you are OK with that and some of the shortcomings already pointed out then feel free to merge it and I can finish up save/restore, and interface pieces hopefully later today but most certainly this this next few days.

cryos:
This should now be ready to merge, there is still more I want to do, but I think it should work, and saves/restores its state.

cryos:
Thanks, I will get this merged, and plan on extending this out to be a little cleverer soon.

Issue #664, opened 2016-10-07
Consider using the Qt5 JSON classes for interface builder


cryos:
They are very nice, simple API, integrate well with the Qt APIs, and we already require Qt5 (first release they were a part of). It shouldn't be too hard to convert things over, especially while things are still at a relatively early stage.

cquammen:
I don't see a huge advantage to using the QJson* API. The API for [`QJsonValue`](http://doc.qt.io/qt-5/qjsonvalue.html) is really quite similar to jsoncpp's [`Json::Value`](http://jsoncpp.sourceforge.net/class_json_1_1_value.html), e.g.:

| QJsonValue val; | Json::Value val; |
| ------------ | ------------ |
| val.isArray() | val.isArray() |
| val.isBool() | val.isBool() |
| val.isDouble() | val.isDouble() |
| val.isNull() | val.isNull() |
| val.isObject() | val.isObject() |
| val.isString() | val.isString() |
| val.toBool() | val.asBool() |
| val.toDouble() | val.asDouble() |
| val.toInt() | val.asInt() |
| val.toString() | val.asString() |
| val.toArray()[index] | val[index] |

`Json::Value` combines the roles of `QJsonObject`, `QJsonValue`, and `QJsonArray`, so it's a little less clean design-wise, but it's not bad IMO.

cryos:
Totally aware of the API for both, and have used both. I think in Qt-based classes it makes far more sense to use the Qt APIs. They benefit from things such as copy-on-write, and use of QString and QVariant which are more natural to use with Qt widgets. Unless there is a compelling reason to not use it then I think we should migrate to it, just like we use QList, QString, etc in our Qt code.

cquammen:
Completed in #675.

cryos:
Thanks Cory

PR #675, opened 2016-10-12
Switch from JsonCPP to QJson API

cquammen:


cquammen:
@cryos Addresses #664.

cquammen:
Ready for review.

cjh1:
Looks like two references escaped the conversion:

```
tomviz/AlignWidget.cxx:834:  connect(&dialog, SIGNAL(applyPreset(const Json::Value&)),
tomviz/HistogramWidget.cxx:261:  connect(&dialog, SIGNAL(applyPreset(const Json::Value&)),
 ```

Everything complied and seems to run OK, but I probably didn't exercise these code paths.

cquammen:
The signal `applyPreset(const Json::Value&)` belongs to `pqPresetDialog`, a ParaView class, so we can't change these.

PR #676, opened 2016-10-12
Add support reporting progress from a Python operator

cjh1:


cquammen:
Tested locally, looks good! I tested Square Root and Invert and saw the progress bar move during processing.

How do I run the test(s)?

cjh1:
@cquammen The tests can be run using CTest. They need [gtest](https://github.com/google/googletest), so you would need to build that and then point our build at it.

cquammen:
Ok, thanks. I might try out the testing at a later time.

PR #677, opened 2016-10-12
Remove Modules from child data sources

### LINKED ISSUE #624: Child visualization not deleted with segmentation object
If I make a visualization from an object produced by a binary segmentation module and then delete the segmentation module the visualization module disappears from the pipeline but the visualization does not go away.

screen shot 2016-09-27 at 2 39 30 pm
screen shot 2016-09-27 at 2 38 33 pm

cquammen:
Modules created to visualize child DataSources were not being removed
when the Operator that created the child DataSource was removed, so
the Module visualizations would persist in the render view with no way
to modify or remove them. Fixed that by removing child DataSource
Modules while the Operator is being removed in the PipelineModel.

Fixes #624

### DIFF
--- a/tomviz/DataSource.cxx
+++ b/tomviz/DataSource.cxx
@@ -15,10 +15,12 @@
 ******************************************************************************/
 #include "DataSource.h"
 
+#include "ModuleManager.h"
 #include "Operator.h"
 #include "OperatorFactory.h"
 #include "PipelineWorker.h"
 #include "Utilities.h"
+
 #include <vtkDataObject.h>
 #include <vtkDoubleArray.h>
 #include <vtkFieldData.h>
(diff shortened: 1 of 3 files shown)


cquammen:
Thanks for the review! I made it recursive to handle more than one level of children.

cquammen:
One way to test the recursive version:

* Load SMALL data set
* Data Transforms -> Binary Threshold, default arguments
* Select the Thresholded Segmentation child data set
* Add Contour
* Data Transforms -> Convert to Float
* Data Transforms -> Binary Threshold
* Add Volume renderer
* Delete the initial Binary Threshold at the top of the Pipeline. All visualizations should be removed save the original Outline and Orthogonal Slice.
* Create a SpreadSheet View. There should be only three items left: `TrivialProducer1`, `Outline1`, `PassThrough1`.

cjh1:
Recursive version tests out as expected.

cquammen:
Thanks, @cjh1

PR #681, opened 2016-10-13
Use shared library for tomvizlib

cjh1:
I was lucky when I first introduced the wrapping, the method we used from Operator where implemented in Operator.h. However, updateProgress is not, so we now get a linking issue on Windows.

I am open to other suggestions, for example only doing this on Windows or another approach.


cjh1:
@cryos PTAL

PR #682, opened 2016-10-14
Fix a bug that non-integer values are passed to np.lib.pad function in direct fourier method

yijiang1:
Fix a small bug that non-integer values are passed to np.lib.pad function in direct Fourier reconstruction.


cquammen:
LGTM

PR #685, opened 2016-10-15
565 state file opacity map

cquammen:
This should fix the problem where the opacity map was not saved.

### DIFF
--- a/tomviz/DataSource.cxx
+++ b/tomviz/DataSource.cxx
@@ -26,6 +26,7 @@
 #include <vtkFieldData.h>
 #include <vtkImageData.h>
 #include <vtkNew.h>
+#include <vtkPiecewiseFunction.h>
 #include <vtkSMCoreUtilities.h>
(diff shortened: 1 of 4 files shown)


tomviz/DataSource.cxx
```diff
@@ -239,6 +241,29 @@ bool DataSource::serialize(pugi::xml_node& ns) const
   tomviz::serialize(this->colorMap(), node);
 
   node = ns.append_child("OpacityMap");
+
+  // Update the proxy values from the VTK object.
+  // TODO - find a nicer way to ensure these values are synchronized.
```
cjh1: Is this TODO for another PR?

cquammen: Yes. There isn't really a simple way to do this in ParaView, e.g., a single function call, but we may want to update the property in a slot that is invoked every time a signal is fired when the opacity map is changed.

cquammen:
Thanks for the review!

PR #698, opened 2016-10-21
Redesign property panel for contour visualization

yijiang1:
@Hovden , @ElliotPadgett and I had a discussion this afternoon and we want to tweak some UI to improve user experience. This PR changes the property panel for contour visualization to 
screen shot 2016-10-21 at 4 59 29 pm

### DIFF
--- a/tomviz/ModuleContour.cxx
+++ b/tomviz/ModuleContour.cxx
@@ -219,7 +219,7 @@ void ModuleContour::addToPanel(QWidget* panel)
 
   QFormLayout* layout = new QFormLayout;
 
-  //Solid color
+  // Solid color
   QHBoxLayout* colorLayout = new QHBoxLayout;
   colorLayout->addStretch();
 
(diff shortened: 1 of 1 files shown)


cquammen:
I don't have strong feelings about this, but I'm just a little concerned it isn't immediately clear that you need to check the check box to use the solid color. Did you consider a design like this? It might be ever so slightly clearer what it does.

image

One could even have the color button disabled when the checkbox is unchecked so that it is clear you can edit the color when you turn on solid coloring.

yijiang1:
We did, but I think @Hovden prefer to have checkbox on the left and label on the right.

ElliotPadgett:
Disabling the color button would help with clarifying things though.

ElliotPadgett:
Either design by @yijiang1 or @cquammen above seems fine to me.

Hovden:
Checkbox on the left is consistent with all other checkboxes used in tomviz.

I don't feel strongly. If we go with @cquammen 's proposal (which we original used) we need some whitespace between the check box and the color.

cquammen:
> Checkbox on the left is consistent with all other checkboxes used in tomviz.

That's true, so I withdraw my suggestion for moving it.

yijiang1:
@cquammen do you mind adding the disable feature later?
I don't know the best way to connect the QCheckBox to the pqColorChooserButton. Both of them are created inside the "addToPanel" function instead of the class.

cquammen:
Nope, not at all. Go ahead and merge

PR #699, opened 2016-10-22
Improve set tilt angle UI

yijiang1:
Added a title to the UI.
Made the UI non-resizeable.

### DIFF
--- a/tomviz/SetTiltAnglesOperator.cxx
+++ b/tomviz/SetTiltAnglesOperator.cxx
@@ -217,7 +217,8 @@ public:
   }
 
 public slots:
-  void updateAngleIncrement() {
+  void updateAngleIncrement()
+  {
     angleIncrement = (endAngle->value() - startAngle->value()) /
                      (endTilt->value() - startTilt->value());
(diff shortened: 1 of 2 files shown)


Hovden:
Good point. I would give it a little extra space just in case there was a change in font size.

yijiang1:
The current size looks like this on my Mac Pro (13'')
screen shot 2016-10-24 at 12 35 23 pm

ercius:
Any chance we can add a third box to set the tilt angle step size (1degree,
1.5 degree, etc)? It could be just a reporting box to let the user know
what the current values will produce.

yijiang1:
> It could be just a reporting box to let the user know what the current values will produce.

Done. The UI now looks like
screen shot 2016-10-25 at 4 11 28 pm

cquammen:
If this is ready to merge, please feel free to do so.

yijiang1:
I'm not entirely happy with current UI design but I'll merge this PR for now.
I'll discuss with people at Cornell about improving user experience and post a new issue in the future.

PR #701, opened 2016-10-22
Added gaussian filter for tilt series

yijiang1:
Added a data processing function that applies a 2D Gaussian filter to each tilt image.

### DIFF
--- a/tomviz/python/GaussianFilterTiltSeries.json
+++ b/tomviz/python/GaussianFilterTiltSeries.json
@@ -1,0 +1,14 @@
+{
+  "name" : "GaussianFilterTiltSeries”,
+  "label" : "Gaussian Filter Tilt Series”,
+  "description" : "Apply an isotropic Gaussian filter to tilt images. \nThe standard deviation\n(sigma) can be specified below:",
+  "parameters" : [
+    {
+      "name" : "sigma",
+      "label" : "Sigma",
+      "type" : "double",
+      "default" : 2.0,
+      "minimum" : 0.0
+    }
+  ]
+}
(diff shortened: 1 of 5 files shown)


tomviz/python/GaussianFilterTiltSeries.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name" : "GaussianFilter",
+  "label" : "Gaussian Filter",
```
yijiang1: What are the purposes of "name" and "label"? 
I don't quite understand how they control the UI.

yijiang1: Could anyone explain this to me? When I edit these two fields, the UI often disappear like this
screen shot 2016-10-24 at 10 39 02 am
Thanks!

cquammen: As described in https://github.com/OpenChemistry/tomviz/blob/master/tomviz/python/Operators.md, `name` is the internal name for the Operator, and `label` is the name that appears in the UI. `name` must not contain spaces.

Editing these fields should not cause the UI to disappear unless the JSON becomes ill formed. In that case, you should get an error message indicating the JSON is invalid.

yijiang1: Turns out the empty UI above is caused by wrong format in Mac's TextEdit (default for .json file). For example, the double quote character in TextEdit becomes 
screen shot 2016-10-26 at 12 37 11 pm

Note to self: use Xcode to edit .json file.

cquammen: Ah, good find. Thanks for tracking this down.

PR #705, opened 2016-10-24
Update autopep8 command to ignore comments

cjh1:
These ignores are already defined in flake8.cfg but autopep8 doesn't
seem to respect them.


cjh1:
@yijiang1 PTAL

Issue #679, opened 2016-10-13
Child data set label and icons not showing up


cjh1:
If I apply the "Binary Threshold" I see the following:

![empty](https://cloud.githubusercontent.com/assets/1031053/19360883/5dbed712-914e-11e6-8496-4ebd30caad2b.png)

@cquammen Are you seeing this on master too? Its totally possible I broke this somehow.

cjh1:
I don't seem to be able to add operators to them either ( child data sources that is ).

cquammen:
I cannot reproduce on master - everything seems to work as expected.

Two comments:

1). When it works correctly, the name/icon doesn't appear until after the operator completes *and* you click on the item. Not ideal.

2). Check the output at the console. My guess is there was an error and no child data set was generated. In that case, the name/icon will not appear at all. Again, not ideal.

cjh1:
The only thing I am seeing in the console is

```
Warning, array does not have Fortran order, making deep copy and fixing...
...done.
```

cquammen:
Hmm, that looks like it completed just fine.

cryos:
I took a look and see it complete, and that it is blank while processing, but shows up after. This should be looked at, but I can't reproduce your error here.

cjh1:
I recompiled using RelWithDebInfo ( I was running Debug before ) and see it update when the processing is complete, looks like some sort of timing issue, I can take look.

cjh1:
So #710 fixes my issue where the child data source was never showing up, I now see what everyone else was seeing ( it appears when the operation is complete ). So this behavior is a result of us creating the row in the table before we have a data source to populate it, so it appears blank until the child data source has been set. Now we could resolve this in by delaying the creation of the row until the data source is available. The problem with this is it is not consistent with our handling on results ( they appear up front ) although we could change that. I guess another option would be to create a "dummy" data source that would be used to render that table during the execution of the operator as we only really need the "filename". Thoughts?

cryos:
I would favor probably go with delayed creation - the new operator is there, and once it completes, and there is data, we should then see what it is. Not sure how much utility there is in the promise of data, especially when we have an operator that we can mark as executing in the UI.

cquammen:
Note that the child data source name should be available from the JSON.

cryos:
[...]

PR #710, opened 2016-10-26
Fix QJson parsing issue

cjh1:
The pattern ```jsonArray[0].toObject()["key"]``` results in a ```QJsonValueRef``` to a ```QJsonObject``` that has gone out of scope. ```childDatasetArray[0].toObject()``` returns a ```QJsonObject``` object, we then take a reference to one of the properties the underlying ```QJsonObject``` object then goes out of scope and we are left with references that refer to a dangling pointer. I was only seeing this when building in Debug mode. 

Fixes part of #679


cjh1:
@cquammen PTAL

PR #711, opened 2016-10-26
Don't automatically change image extent when setting array

cquammen:
Setting the minimum extent to (0, 0, 0) in utils.set_array(...)  is not always the right thing to do, especially after an Operator changes the size of the image. This branch makes setting the extent optional, splitting it into a separate function `tomviz.util.set_extent(...)` that should be called before `tomviz.util.set_array(...)`. The extent need only be set if the size of the image is changed by the transform.

### DIFF
--- a/tomviz/python/Crop_Data.py
+++ b/tomviz/python/Crop_Data.py
@@ -26,5 +26,5 @@ def transform_scalars(dataset):
                            START_CROP[2]:END_CROP[2]]
 
     # Set the data so that it is visible in the application.
-    utils.set_array(dataset, data_cropped)
+    utils.set_array(dataset, data_cropped, START_CROP)
     print('Data has been cropped.')
--- a/tomviz/python/Pad_Data.py
+++ b/tomviz/python/Pad_Data.py
@@ -26,7 +26,10 @@ def transform_scalars(dataset):
     array = np.lib.pad(array, pad_width, padMode)
 
     # Set the data so that it is visible in the application.
-    utils.set_array(dataset, array)
+    extent = list(dataset.GetExtent())
+    start = [x - y for (x, y) in zip(extent[0::2], pad_size_before)]
+
+    utils.set_array(dataset, array, start)
 
     # If dataset is marked as tilt series, update tilt angles
     if padWidthZ[0] + padWidthZ[1] > 0:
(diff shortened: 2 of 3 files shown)


cquammen:
Name and function calling pattern suggestions welcome.

cryos:
This feels like adding a lot of boilerplate code, and I can see why, but is there a reasonable way to do the right thing in the default case, and only require setting the extents when necessary? Our origin at zero assumption was bad, maybe we could use the original origin by default for example rather than zero.

cquammen:
> This feels like adding a lot of boilerplate code

Yeah, I'd love to not have to touch most of these transforms, and indeed some of the transforms may not need to set the extent at all if they leave the dimensions untouched. Could I get feedback from transform authors @yijiang1 @ElliotPadgett @Hovden @mathturtle as to which of the transforms I have changed here promise to not change the dimensions of the image? In those transforms, I can remove the `utils.set_extent` call.

cquammen:
> maybe we could use the original origin by default for example rather than zero.

Yes, that will be the case if you do not call `utils.set_extent`.

Another way to do this would be to add an `extent` parameter to `utils.set_array` that defaults to `None` but if it is set is expected to contain the new extent of the image data? That might be cleaner.

tomviz/python/Shift3D.py
```diff
@@ -21,5 +21,6 @@ def transform_scalars(dataset):
 
     data_py_return = ndimage.interpolation.shift(data_py, SHIFT, order=0)
```
yijiang1: `ndimage.interpolation.shift` does not change matrix size.

yijiang1:
> Setting the minimum extent to (0, 0, 0) in utils.set_array(...) is not always the right thing to do, especially after an Operator changes the size of the image.

What's the bad consequence of this?

> Another way to do this would be to add an extent parameter to utils.set_array that defaults to None but if it is set is expected to contain the new extent of the image data? That might be cleaner.

I prefer this way over having a separate `utils.set_extent` function.

cquammen:
> What's the bad consequence of this?

The consequence is that the image may be translated in a way you don't want.

cquammen:
I pushed a modification that removes the `utils.set_extent()` function and augments `utils.set_array()` with an optional argument to specify the data object's new minimum extent. This geatly simplifies usage such that only two transforms Python scripts need to be updated, one of which does not appear to be in use.

cquammen:
Ready for another look, @cryos

cquammen:
Just had to build one to throw away to see the better solution, I guess. Thanks for the feedback.

PR #712, opened 2016-10-27
Remove label map array

cquammen:
Remove code for when label maps were stored as separate arrays

Now that Tomviz supports child datasets, the kludge for storing
segmentation results as LabelMap arrays in the dataobject is no longer
needed. The functionality this commit removes is the ability to switch
between coloring visualizations by image data or by label maps. For
child datasets resulting from segmentation algorithms, the scalar
array is the label map, so it is indeed possible to color these
datasets by label map.
    
This commit does remove the ability to color a Contour of a dataset by
a label map. A future improvement could be to modify the Contour
visualization to enable selection of a child data set containing
a label map for coloring.

### DIFF
--- a/tomviz/Module.cxx
+++ b/tomviz/Module.cxx
@@ -76,7 +76,7 @@ public:
 
 Module::Module(QObject* parentObject)
   : Superclass(parentObject), UseDetachedColorMap(false),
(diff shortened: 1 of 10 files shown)


cquammen:
Ready for review @cryos @cjh1 @mathturtle

PR #713, opened 2016-10-27
Handle cases where data is added when spreadsheet view is active

### LINKED ISSUE #587: Crash when SpreadSheet view is active and data is loaded
I suspect tomviz is trying to add proxies defined in a `Module` to the spreadsheet view, but have not investigated to confirm.

cquammen:
There was an assumption in LoadDataReaction and
PythonGeneratedDatasetReaction that the current active view is a
Render View. When that isn't the case, visualization modules are not
added correctly, signals and slots are not set up properly, and a
crash might occur. In any case, Tomviz becomes unstable.

Fixed this problem by checking that the active view is a Render View,
and if not, look for one. If there isn't one, create a Render View.

Fixes #587

### DIFF
--- a/tomviz/ActiveObjects.cxx
+++ b/tomviz/ActiveObjects.cxx
@@ -19,9 +19,14 @@
 
 #include <pqActiveObjects.h>
 #include <pqApplicationCore.h>
+#include <pqObjectBuilder.h>
 #include <pqPipelineSource.h>
 #include <pqServer.h>
 #include <pqView.h>
+
+#include <vtkNew.h>
+#include <vtkSMProxyIterator.h>
+#include <vtkSMRenderViewProxy.h>
 #include <vtkSMSourceProxy.h>
 #include <vtkSMViewProxy.h>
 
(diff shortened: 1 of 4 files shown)


cquammen:
Ready for review, @cryos @cjh1 @mathturtle

tomviz/ActiveObjects.cxx
```diff
@@ -121,6 +126,41 @@ void ActiveObjects::setActiveModule(Module* module)
...
+void ActiveObjects::createRenderViewIfNeeded()
+{
+  vtkNew<vtkSMProxyIterator> iter;
+  iter->SetSessionProxyManager(proxyManager());
+  iter->SetModeToOneGroup();
+  for (iter->Begin("views"); !iter->IsAtEnd(); iter->Next()) {
+    vtkSMRenderViewProxy* renderView =
+      vtkSMRenderViewProxy::SafeDownCast(iter->GetProxy());
+    if (renderView) {
+      std::cout << "Found renderview\n";
```
cjh1: Debug message?

cquammen: :blush: Removed.

PR #714, opened 2016-10-27
Name generated data

cquammen:
Generated data sources had empty file names, so in the UI their items
in the UI were blank. Set the file name from the generated data source
label to give it a nice name.

Also, fixed a crash when generating an Electron Beam Shape data set.


cquammen:
Ready for review @cryos @cjh1 @mathturtle

cquammen:
Results look like

image

PR #719, opened 2016-10-29
Updates to the ParaView cmake lines

cryos:
Added new PARAVIEW_USE_QTHELP, then reorganized into related options
for VTK, ParaView, etc.


cryos:
Thanks for the late night review :-)

PR #720, opened 2016-10-29
Operator activation

cryos:


cryos:
I totally agree, I was thinking of trying to do it, but thought this was a decent start.

PR #723, opened 2016-11-01
Fix issue #565, volume now saves separate color map

cryos:
The serialize method for the base class was not being called.

### DIFF
--- a/tomviz/ModuleVolume.cxx
+++ b/tomviz/ModuleVolume.cxx
@@ -129,7 +129,7 @@ bool ModuleVolume::serialize(pugi::xml_node& ns) const
   bool maxIntensity =
     m_volumeMapper->GetBlendMode() == vtkVolumeMapper::MAXIMUM_INTENSITY_BLEND;
   maxIntensityNode.append_attribute("enabled") = maxIntensity;
-  return true;
+  return Module::serialize(ns);
 }
 
 bool ModuleVolume::deserialize(const pugi::xml_node& ns)
@@ -154,7 +154,7 @@ bool ModuleVolume::deserialize(const pugi::xml_node& ns)
     }
   }
 
-  return this->Superclass::deserialize(ns);
+  return Module::deserialize(ns);
 }
 
 void ModuleVolume::addToPanel(QWidget* panel)


tomviz/ModuleVolume.cxx
```diff
@@ -154,7 +154,7 @@ bool ModuleVolume::deserialize(const pugi::xml_node& ns)
     }
   }
 
-  return this->Superclass::deserialize(ns);
+  return Module::deserialize(ns);
```
cquammen: Curious: why do you prefer `Module` here instead of `Superclass` (which is typdefed to `Module`)?

cryos:
That is weird, just tested this again on a different machine with and without separate color map and it works here too. It also saved the color map that I changed to the rainbow in my tests, these are both Linux machines, can you try it again and confirm this is on Mac.

cquammen:
Source version:

```
terminus:tomviz cory.quammen$ git log -1
commit 9ac021515664561b7f767dd69c03e7bb6fd59918
Author: Marcus D. Hanwell <marcus.hanwell@kitware.com>
Date:   Tue Nov 1 14:51:58 2016 -0400

    Fix issue #565, volume now saves separate color map
    
    The serialize method for the base class was not being called.
```

Testing procedure on Mac OS X 11.6:

* Launch tomviz, confirm version is "tomviz 0.9.1-465-g9ac0215
* Load `SMALL_PtCu_NanoParticle.tif
* Add Volume visualization
* Check separate color map
* Choose a different color map from the presets, e.g., Black-Body Radiation
* Add three control points to the Opacity map, move them to various locations
* Save state file.
* Close tomviz
* Relaunch tomviz
* Load state file. Separate color map for Volume is restored appropriately. The opacity map, however, has been reset to the default with just 2 control points.

cryos:
I missed you were talking about the opacity function, rather than the color map which is saved. In my local testing the opacity function is not saved for volume or any other module in a separate color map scenario. If you add a contour, set some points, save, that will also not be restored. That seems to be a separate issue that what was done to save opacity was only effective for the global color map.

cquammen:
I see. Well the separate color map seems to work fine for Volume modules now. I recently put in a fix for restoring the opacity map for the main color map in the dataset.

I am still wondering about the `Superclass` - `Module` swap to see if there is a programming practice I should adopt.

cryos:
Thanks Cory, I prefer the more explicit call to the intended classes, and prefer it over the non-standard typedef. In C++11 or greater we should also prefer the new using as it is easier to read and preferred over typedefs now, i.e. using Super = Base;

PR #728, opened 2016-11-02
Add pipeline state column

cjh1:
This adds an extra column to the pipeline view to indicate the state of the operator

pipe

Currently the running operator icon is not animated, but I would like to fix that.


cryos:
Looking very nice, about to test it out, but wanted to point out what I said when looking over your shoulder. I think when I carried over the delete button from the operator pipeline it was a mistake, and it makes the rightmost column inconsistent. Module visibility mixed with deleting operators and nothing for the data sources.

Happy to address this in a follow up commit, but I really think the status icons should be moved to where the delete symbol is now, and the delete removed. The data source and modules would be far more consistent if operators were deleted in the way way using the context menu or the delete key. This also saves precious space for labels, etc.

cryos:
I get the following error - it looks like there is a missing icon that is referenced.

ninja: error: '/home/marcus/src/tomviz/tomviz/icons/hourglass_32.png', needed by 'tomviz/CMakeFiles/tomviz_automoc', missing and no known rule to make it

tomviz/Operator.cxx
```diff
@@ -38,12 +38,15 @@ DataSource* Operator::dataSource()
 
 bool Operator::transform(vtkDataObject* data)
 {
-  m_finished = false;
-  m_canceled = false;
+  this->m_state = OperatorState::RUNNING;
```
cryos: m_ shows it is a member, shouldn't use this-> as well.

tomviz/Operator.cxx
```diff
@@ -38,12 +38,15 @@ DataSource* Operator::dataSource()
...
   emit transformingStarted();
   emit updateProgress(0);
   bool result = this->applyTransform(data);
-  m_finished = true;
+  // TODO we may want to change the return type applyTransform
+  // to OperatorState so we don't have todo this.
+  if (this->m_state != OperatorState::CANCELED) {
```
cryos: As above

cjh1:
I have pushed a commit that removes that image.

tomviz/PipelineModel.cxx
```diff
@@ -300,7 +336,7 @@ QVariant PipelineModel::data(const QModelIndex& index, int role) const
 
   // Data source
   if (dataSource) {
-    if (index.column() == 0) {
+    if (index.column() == 1) {
```
cryos: It might be easier and more readable if we upgrade these to const ints or similar internally so that they can be changed, and the meaning is clearer, i.e.

namespace {
const int stateColumn;
const int labelColumn;
}

tomviz/PipelineWorker.cxx
```diff
@@ -116,6 +117,11 @@ void PipelineWorker::RunnableOperator::cancel()
   this->m_operator->cancelTransform();
 }
 
+bool PipelineWorker::RunnableOperator::isCanceled()
+{
+  return this->m_operator->isCanceled();
```
cryos: Don't need this->

cryos:
Building now, thanks Chris.

cryos:
I think the interaction and informational cues look great, it is a shame that the progress icon doesn't animate, but it is still pretty clear what is happening.

cjh1:
Yes, I would like to get the progress icon animated, I have a few things I would like to try.

cquammen:
Agreed, looking good.

tomviz/PipelineView.cxx
```diff
@@ -142,8 +142,8 @@ void PipelineView::deleteItem(const QModelIndex& idx)
 
 void PipelineView::rowActivated(const QModelIndex& idx)
 {
-  if (idx.isValid() && idx.column() == 1) {
-    auto pipelineModel = qobject_cast<PipelineModel*>(model());
+  if (idx.isValid() && idx.column() == 2) {
+    auto pipelineModel = qobject_cast<PipelineModel*>(this->model());
```
cquammen: I don't think we are using `this->` when calling member functions anymore.

cjh1:
@cryos I have moved the state column and addressed the other issues

cryos:
This looks mostly good, but the pipeline on my machine now has roughly a 50:50 split when we want column 0 to take up the majority of the space.

tomviz/PipelineModel.cxx
```diff
@@ -324,7 +360,7 @@ QVariant PipelineModel::data(const QModelIndex& index, int role) const
         default:
           return QVariant();
       }
-    } else if (index.column() == 1) {
+    } else if (index.column() == Column::state) {
```
cryos: This is so much easier to read!

tomviz/PipelineModel.cxx
```diff
@@ -455,7 +496,7 @@ int PipelineModel::rowCount(const QModelIndex& parent) const
 
 int PipelineModel::columnCount(const QModelIndex&) const
 {
-  return 2;
+  return 3;
```
cryos: This needs to be 2

PR #730, opened 2016-11-02
Add progress bar to more tomography reconstructions

yijiang1:
I want to add progress bar to more reconstruction transforms. 
This PR currently adds progress bar to "Direct Fourier Method" but it doesn't work properly and I don't know how to fix it...

Here is the problem:
1. Load a dataset, mark it as tilt series.
2. Run "Direct Fourier Method" reconstruction. The progress bar does not update during reconstruction. It looks like this: 
screen shot 2016-11-02 at 2 19 36 pm

3. After reconstruction finishes, double-click "Reconstruct (Direct Fourier)" in the pipeline.
4. Make some modification to the code (for example, add an empty line), then click "OK".
5. Now reconstruction is re-run and the progress bar seems working:
screen shot 2016-11-02 at 2 19 44 pm

### DIFF
--- a/tomviz/Operator.h
+++ b/tomviz/Operator.h
@@ -131,8 +131,11 @@ public:
   /// that QProgressBar interprets the progress as unknown.
   int totalProgressSteps() const { return m_totalProgressSteps; }
 
-  void setTotalProgressSteps(int steps) { m_totalProgressSteps = steps;
-    emit totalProgressStepsChanged(steps);}
+  void setTotalProgressSteps(int steps)
+  {
+    m_totalProgressSteps = steps;
+    emit totalProgressStepsChanged(steps);
+  }
 
 signals:
   /// Emit this signal with the operation is updated/modified
--- a/tomviz/Operator.h
+++ b/tomviz/Operator.h
@@ -131,7 +131,8 @@ public:
   /// that QProgressBar interprets the progress as unknown.
   int totalProgressSteps() const { return m_totalProgressSteps; }
 
-  void setTotalProgressSteps(int steps) { m_totalProgressSteps = steps; }
+  void setTotalProgressSteps(int steps) { m_totalProgressSteps = steps;
+    emit totalProgressStepsChanged(steps);}
 
 signals:
   /// Emit this signal with the operation is updated/modified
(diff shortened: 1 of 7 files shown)


cjh1:
@yijiang1 I have fixes for this if you give me push access to you PR I can push the changes

yijiang1:
> give me push access to your PR

How can I do this?

cjh1:
@yijiang1 Looks like I already can :smile: There was a missing signal ( may have got lost in a merge ). Also setting progress.maximum as soon as we get into the transform_scalars(...) methods, otherwise we get a busy progress bar ( at 100% ) ... Let me know if these changes make sense to you.

yijiang1:
> Also setting progress.maximum as soon as we get into the transform_scalars(...) methods, otherwise we get a busy progress bar ( at 100% )

I don't quite understand why we need to do this?

cjh1:
If we don't do that the progress bar will default to a busy indicator until you set a maximum progress, which might be OK, but i thought it looks a little odd. As first you would see the progress bar at 100% and then it would jump back to zero when you the set the maximum progress. Comment out the line and you will see what I mean.

From the Qt doc:

"If minimum and maximum both are set to 0, the bar shows a busy indicator instead of a percentage of steps"

yijiang1:
Later in the script (line 49) I set self.progress.maximum to a non-zero integer. So why do I need to do this at the beginning?

cjh1:
Because executing lines 8 - 49 take a couple of seconds ( the first time ), during this time the progress bar will be displayed as a busy indicator ( at 100% ). I would comment out that line and you see what I mean ...

yijiang1:
Ah I see. Thanks for the explanation!

cjh1:
No problem

yijiang1:
Travis is telling me this error:
screen shot 2016-11-03 at 12 48 39 pm
What is my problem and how should I fix it?

cjh1:
@yijiang1 Basically it because your transform_scalars has good too [complicated](https://en.wikipedia.org/wiki/Cyclomatic_complexity) ( to much branching ). I would refactor it to use some more methods, for example encapsulate each branch of the if statement in method.

yijiang1:
I see. I'll refactor the script.

yijiang1:
I've added progress bar to all python-based reconstruction methods. Ready for review and merge.

PR #733, opened 2016-11-03
Keep the scalar opacity function proxy updated

cquammen:
When the scalar opacity function was changed in the opacity editor,
the proxy was left none the wiser. Now, we explicitly update the
scalar opacity function property in the slot for scalar opacity
function changes.


cquammen:
@cryos @cjh1 This should be ready for review. TIA

cryos:
This works great

Issue #449, opened 2016-06-24
Orthoslice slider is too coarse


ElliotPadgett:
When looking at a 512X512X512 reconstruction with an orthoslice visualization the slider takes steps of about 10.  I often want to scroll around in 1 pixel steps around a region of interest, and typing the value is not very convenient.  Is it possible to make the sider have finer control?

ercius:
I think you can use the keyboard arrow keys to take single steps once the
slice slider bar is clicked.

Hovden:
I cannot adjust with the keyboard on OSX. Maybe other OS's treat it different. Keyboard adjustment would address this issue.

This may be an opportunity to also discuss improvement UX on the orthoslice:
<img width="306" alt="screen shot 2016-06-27 at 9 08 46 am" src="https://cloud.githubusercontent.com/assets/6197997/16380760/ee114ce4-3c46-11e6-9e89-30236f518541.png">
1) "Slice Direction" should read "Direction"
2) Value box should be no larger than 5 characters.

This will make more space for the slider and look better. I know this is may be a Paraview request, but it should be easy and beneficial.

mathturtle:
This should be fairly simple now.  I'll just do the same thing I did for the Contour module for the orthoslice.

Hovden:
bump (I'm orthoslicing frequently this week)

mathturtle:
Can this be closed now?

ElliotPadgett:
Still an issue for me. I'm on Mac, and I have a ~400 voxel long dataset and the orthoslice slider jumps 5 slices at a time and there's no easy way to move it with the keyboard except typing in the number.

mathturtle:
Hmmm. I'll try this on a mac and see if it is different there.

mathturtle:
Mac is different.  The slider jumps to where you click on it.  I can make it get keyboard focus when you click on it and then you will be able to use the arrow keys to adjust it.

ElliotPadgett:
Great.  This solves it!

PR #734, opened 2016-11-03
Make the slider grab keyboard focus when clicked on

mathturtle:
This allows stepping by 1 (using the keys) when using an orthoslice on a large dataset.  This is needed since Qt on OSX doesn't behave the same when you click on the slider above or below its current value.

@Hovden @ElliotPadgett This is for #449.

### DIFF
--- a/tomviz/IntSliderWidget.cxx
+++ b/tomviz/IntSliderWidget.cxx
@@ -37,6 +37,7 @@ IntSliderWidget::IntSliderWidget(bool showLineEdit, QWidget* p) : QWidget(p)
   l->setMargin(0);
   this->Slider = new QSlider(Qt::Horizontal, this);
   this->Slider->setRange(this->Minimum, this->Maximum);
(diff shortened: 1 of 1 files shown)


mathturtle:
> on Linux the step size may still need some adjustment


Yes, the step size was adjusted before since it is what the slider steps by on Linux/Windows when you click on it above or below the current value.  But Mac is different and this was the best way I could find to let Mac users have fine control.  I can take out the line that sets the step size and it will default back to 10 (or 10% of the range, I forget).

cryos:
So for me this is stepping by 3 on the SMALL data set (unsigned char). Float might be interesting too, I get the motivation and expected the step size to be 1.

cryos:
In fact it is 2 or 3, but it is pretty granular, is it mapping a step size of 1 onto a range of 255?

cryos:
I will just merge this as it is an improvement, it can be tweaked further in follow ups.

Issue #735, opened 2016-11-03
Add Tomviz version number to the state file


cryos:
This will aid in future versions being able to point out it was saved with an earlier version of Tomviz, and will offer metadata that can aid in publication/reproducibility.

cryos:
Got to it quicker than I thought, this is now merged and will be accessible in future so that we can add further checks based upon the Tomviz version.

PR #736, opened 2016-11-03
Add the Tomviz version and extra version to state files

cryos:
We don't do anything with it just yet, but adding it to the files will
help in future releases. This results in a compact xml node that was
near the top in my testing, such as,

  `<version` full="0.9.1" extra="481-g67e512b" />

This fixes issue #735.


cryos:
Thanks for the near instantaneous review @cquammen :-)

PR #738, opened 2016-11-03
Show the current dataset's units in the cube axes labels

mathturtle:
This does have a side effect of clearing the user-set axis labels, but
the units are visible now.  I am open to suggestions for how to avoid clobbering the user's axis labels if they specify any.

### DIFF
--- a/tomviz/DataPropertiesPanel.cxx
+++ b/tomviz/DataPropertiesPanel.cxx
@@ -28,6 +28,7 @@
 #include <vtkDataSetAttributes.h>
 #include <vtkPVArrayInformation.h>
 #include <vtkPVDataInformation.h>
+#include <vtkSMPropertyHelper.h>
 #include <vtkSMSourceProxy.h>
 #include <vtkSMViewProxy.h>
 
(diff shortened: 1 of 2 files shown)


cquammen:
@mathturtle Can you provide instructions how to test? I turned on the option View -> Show Axis Grid, but I don't see any unit labels.

mathturtle:
Did you have a dataset loaded?  You should see something like "X Axis (nm)" on all the axis titles.

tomviz_axis_labels

cquammen:
Ooops, PEBKAC. I fetched this branch but didn't check it out. I see the units now.

I don't think you are going to be able to avoid clobbering customized labels, and there is the further  complication that if have two datasets loaded with different units, the second one loaded will be shown. Maybe that is just an unlikely use case, but something to keep in mind.

mathturtle:
> there is the further complication that if have two datasets loaded with different units, the second one loaded will be shown. Maybe that is just an unlikely use case, but something to keep in mind.

It will show the currently selected dataset's units, which I thought made the most sense in that case.  I made it update based on the active dataset.  But we may want a different approach to displaying units to avoid clobbering the labels.

tomviz/DataPropertiesPanel.cxx
```diff
@@ -353,4 +361,30 @@ void DataPropertiesPanel::updateZLength()
...
+  vtkSMViewProxy *view = ActiveObjects::instance().activeView();
+  if (!view) {
+    return;
+  }
+  vtkSMProxy *axesGrid = vtkSMPropertyHelper(view, "AxesGrid").GetAsProxy();
+  DataSource *ds = ActiveObjects::instance().activeDataSource();
+  if (!axesGrid || !ds) {
+    return;
+  }
+  QString xTitle = QString("X Axis (%1)").arg(ds->getUnits(0));
```
cryos: I think we could just label these X, Y, and Z - the axis label feels a little redundant to me.

Hovden: +1

Hovden:
Merge?

mathturtle:
Ok, so that should fix the formatting issues.  And I shortened the labels to "X (nm)" as per @cryos's suggestion.

Issue #742, opened 2016-11-04
Volume render forgets show/hide status when state restored


ElliotPadgett:
If I make a volume render, hide it, and then save/restore the state it forgets that it is hidden, and is shown anyway.  Other renderings do not have this problem.

cryos:
This should now be resolved.

PR #743, opened 2016-11-04
Save the visibility value for the volume module

cryos:
This fixes issue #742 where the visibility was not correctly restored
for volumes.


cquammen:
Okay, locally tested after all. Still looks good.

cryos:
Thanks @cquammen

PR #749, opened 2016-11-07
A few more tilt series related fixes

cjh1:


cjh1:
@cryos @cquammen A few more tilt angles related fixes. I would appreciate some review and QA.

PR #763, opened 2016-11-21
Exposed the UseJittering volume renderer parameter

cryos:
This can be useful in reducing artifacts in certain scenarios, exposing
it required uising the GPU volume mapper directly. This isn't likely to
have a large impact as we already require OpenGL2 rendering which
implies an OpenGL 3.2 compliant graphics card. There are some other
elements of the GPU mapper I am looking at exposing in the future.


cryos:
Thanks Cory, I did not see any big differences

PR #766, opened 2016-11-21
Pass parameter values as arguments to transform_scalars(...)

cjh1:
Remove templating of Python scripts in favor of parameter passing.

I have converted the following  operators, I would appreciate some help in confirming that they are still operating as design ( please add a check mark, if you have successfully test one ... )

- [x] AddPoissonNoise.py
- [x] BinaryClose.py
- [x] BinaryDilate.py
- [x] BinaryErode.py
- [x] BinaryOpen.py
- [x] BinaryThreshold.py
- [x] ClearVolume.py
- [X] GaussianFilter.py
- [x] ConnectedComponents.py
- [x] Crop_Data.py ( is this still used? )
- [x] deleteSlices.py
- [x] GaussianFilterTiltSeries.py
- [x] GenerateTiltSeries.py
- [x] MedianFilter.py
- [x] OtsuMultipleThreshold.py
- [x] Pad_Data.py
- [x] PeronaMalikAnisotropicDiffusion.py
- [x] Recon_ART.py
- [x] Recon_DFT_constraint.py
- [x] Recon_SIRT.py
- [x] Recon_TV_minimization.py
- [x] Recon_WBP.py
- [x] RemoveBadPixelsTiltSeries.py
- [x] Resample.py
- [x] Rotate3D.py
- [x] Shift_Stack_Uniformly.py
- [x] Shift3D.py
- [x] ShiftTiltSeriesRandomly.py
- [x] Subtract_TiltSer_Background.py


cjh1:
OK, so BinaryThreshold fails with: 

```
Exception encountered while running BinaryThreshold
in method 'itkBinaryThresholdImageFilterIUC3IUC3_SetLowerThreshold', argument 2 of type 'unsigned char'
```

cjh1:
Looks like in the JSON the threshold values are set to doubles, but the should be int?

cquammen:
No, they should be doubles in case you have a data set of float or double values. We'll need to cast inside the filter. I'm working on the ITK code needed now.

cjh1:
@cquammen Thanks for looking at this.

cquammen:
NP. It turns out to be a bit tricky.

cjh1:
@cquammen Why did we not see this problem before ( before I made these changes )?

cquammen:
You can actually see the same error if you enter a threshold with a fractional number in the Binary Threshold dialog box while processing an integral-typed data set. In my testing I apparently just used numbers without fractional parts. In the string substitution, the `.0` got chopped off, making the values legal `unsigned chars`.

cjh1:
OK, thanks, that makes sense.

cquammen:
I have a patch for `utils.py` and `BinaryThreshold.py` that should resolve this. Mind if I push it onto this branch?

cjh1:
@cquammen Go right ahead, thanks!

yijiang1:
I'm having trouble compiling the branch:

In file included from /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/iostream:38:
In file included from /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/ios:216:
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/__locale:468:15: error: C++
      requires a type specifier for all declarations
    char_type toupper(char_type __c) const
              ^
/usr/include/python2.7/pyport.h:731:29: note: expanded from macro 'toupper'
#define toupper(c) towupper(btowc(c))
                            ^
In file included from /Users/yijiang/Visualization/tomviz/tomviz/DataTransformMenu.cxx:28:
In file included from /Users/yijiang/Visualization/tomviz/tomviz/Utilities.h:25:
In file included from /Users/yijiang/Visualization/paraview/ParaViewCore/ServerManager/Core/vtkSMSourceProxy.h:37:
In file included from /Users/yijiang/Visualization/paraview/ParaViewCore/ServerManager/Core/vtkSMProxy.h:133:
In file included from /Users/yijiang/Visualization/paraview/ParaViewCore/ClientServerStream/vtkClientServerID.h:29:
In file included from /Users/yijiang/Visualization/paraview/VTK/Common/Core/vtkIOStream.h:31:
In file included from /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/iostream:38:
In file included from /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/ios:216:
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/__locale:474:48: error: too
      many arguments provided to function-like macro invocation
    const char_type* toupper(char_type* __low, const char_type* __high) const
                                               ^
/usr/include/python2.7/pyport.h:731:9: note: macro 'toupper' defined here
#define toupper(c) towupper(btowc(c))
        ^

...

tomviz/python/tomviz/utils.py
```diff
@@ -154,6 +160,72 @@ def vtk_cast_map():
...
+                vtk.VTK_FLOAT: float,
+                vtk.VTK_DOUBLE: float
+            }
+
+        pd = dataset.GetPointData()
+        scalars = pd.GetScalars()
+
+        return _vtk_to_python_types[scalars.GetDataType()]
+    except:
+        pass
```
cjh1: Can add the type of the exception ( KeyError? ) I , otherwise this might hide errors such  as being unable to import vtk?

cquammen: Sure, good call.

cquammen:
Okay @cjh1, want to see if this resolves the problem?

cjh1:
@cquammen Thanks for this fix. Is this something that might be fix in ITK in the future? Seems like it would be a good improvement? I still see the error if I select a value that can't be represented as an unsigned char, for example 300, not sure what is going on there?

cjh1:
@yijiang1 I'm not sure those errors are related to this branch did you upgrade your version of XCode recently? Can you try compiling master?

yijiang1:
Yeah I am able to compile master with no errors. I'm using the latest xcode (v8.1)

cjh1:
@cquammen Do you have any insight as to why this might be failing for @yijiang1?

cjh1:
@yijiang1 I just pushed a change, can you try it, its a total guess! But might fix your problem ...

yijiang1:
@cjh1 I'm able to compile the code now.

cjh1:
Awesome!

yijiang1:
Is there a way to display and edit user-specified parameters in python script?

cjh1:
@yijiang1 Yes, that is the end goal of this PR, when you click on a Python operator it will bring up that parameters rather than the templated code.

yijiang1:
@cjh1 I see. Sounds good. I tested "AddPoissonNoise.py" and it works fine.

cjh1:
@yijiang1 Great, thanks for the QA.

cquammen:
> Is this something that might be fix in ITK in the future? Seems like it would be a good improvement?

Agreed. I'll talk to @thewtex about it. Certain a pain to have to do typecasting of member function arguments in Python(!)

> I still see the error if I select a value that can't be represented as an unsigned char, for example 300, not sure what is going on there?

If your input dataset has unsigned char voxels, the ITK binary threshold filter threshold arguments need to be unsigned char (or convertible to unsigned char). That's just how the filter is defined for better or worse. The good news is that 300 should be a totally valid value for unsigned short or unsigned int image types, etc. What I am seeing when entering 300 is that the argument value passed in to `transform_scalars` is truncated to 255.0.

cjh1:
@cquammen I am not seeing that, made be I have a different version of ITK? I have add the following print statements to the operator:

```
diff --git a/tomviz/python/BinaryThreshold.py b/tomviz/python/BinaryThreshold.py
index f87fdf7..7062daa 100644
--- a/tomviz/python/BinaryThreshold.py
+++ b/tomviz/python/BinaryThreshold.py
@@ -26,10 +26,16 @@ def transform_scalars(dataset, lower_threshold=40.0, upper_threshold=255.0):
         # representation.
         itk_output_image_type = itk.Image.UC3
 
+        print upper_threshold
+
         # ITK's BinaryThresholdImageFilter does the hard work
         threshold_filter = itk.BinaryThresholdImageFilter[
             itk_input_image_type, itk_output_image_type].New()
         python_cast = utils.get_python_voxel_type(itk_image)
+
+        print python_cast(upper_threshold)
+        print type(python_cast(upper_threshold))
+
         threshold_filter.SetLowerThreshold(python_cast(lower_threshold))
         threshold_filter.SetUpperThreshold(python_cast(upper_threshold))
         threshold_filter.SetInsideValue(1)

```

This is the output I am seeing if I enter 300 ( this is with out small test dataset ):

```
300.0
300
<type 'int'>
Exception encountered while running BinaryThreshold
in method 'itkBinaryThresholdImageFilterIUC3IUC3_SetUpperThreshold', argument 2 of type 'unsigned char'
```

cquammen:
Okay, I see the same thing. I accidentally rebuilt my ParaView build for tomviz rather than tomviz itself when I made my earlier comments. Investigating....

cquammen:
Ah, it's an overflow error, which makes sense. 300.0 is cast to 300, which is greater than the supported values in an unsigned char, obviously. 

Not sure what the best approach is, here. I suppose for thresholding, clamping to the representable range would be reasonable.

cjh1:
I think clamping to the representable range would a good idea, outside of the scope of the PR I think.

cquammen:
Sounds good. Doing so is another little knot of Python-wrapped ITK code.

cjh1:
@cquammen Any chance you could confirm I didn't break the other BinaryXXX operators?

cquammen:
Sure thing. Have to run now but will look at tonight.

cjh1:
No rush, thanks

cquammen:
The other `Binary*` operators work as expected.

tomviz/python/deleteSlices.py
```diff
@@ -1,14 +1,9 @@
-def transform_scalars(dataset):
+def transform_scalars(dataset, firstSlice=None, lastSlice=None, axis=None):
     """Delete Slices in Dataset"""
 
     from tomviz import utils
     import numpy as np
     axis = 0
```
cjh1: So we should just remove that line?

cjh1: @yijiang1 By the way, thanks so much for all the testing!

yijiang1: > So we should just remove that line?

Yeah I think so.

yijiang1:
For Crop_Data.py, we want to connect it with the "SelectVolumeWidget" (see the code for "Background Subtraction (Manual)" in AddPythonTransformReaction.cxx for example).

cjh1:
@yijiang1 I will add issue for doing that, thanks.

yijiang1:
@cjh1  Sounds good. Thanks!

yijiang1:
I'm having some ITK-related problems currently so I can't test PeronaMalikAnisotropicDiffusion.py...

cjh1:
@cquammen I get the following error when running PeronaMalikAnisotropicDiffusion.py:

```
Exception encountered while runningPeronaMalikAnisotropicDiffusion
"itkTemplate : No template (<class 'itkImagePython.itkImageUC3'>, <class 'itkImagePython.itkImageUC3'>) for the itk::GradientAnisotropicDiffusionImageFilter class"

```

Do we need some casting here as well?

@yijiang1 Is this what you are seeing?

yijiang1:
> @yijiang1 Is this what you are seeing?

Yes!

cquammen:
Yup, I see the same thing with `PeronaMalikAnisotropicDiffusion.py`. Not an issue with this PR, though.

The ITK filter is templated only over float images, so a workaround for unsigned char image data is to apply `Convert to Float` first. I'll submit a separate PR that automatically casts the image data to float in this operator to avoid this exception in the future.

cjh1:
@cquammen Thanks

tomviz/Utilities.h
```diff
@@ -18,6 +18,7 @@
 
 // Collection of miscellaneous utility functions.
 
+#include "vtkPython.h" // must be first
```
cryos: This shouldn't be needed now you moved Python utilities to their own file should it?

cjh1: Good catch, I will remove it.

thewtex:
Will chat with @cquammen after I return from the holiday to think about any improvements we can make here.

PR #771, opened 2016-11-23
Make Perona-Malik anisotropic apply to additional image types

cquammen:
Add auto-casting of input data set to a type supported by the ITK filter.


cjh1:
@cquammen I don't see the errors anymore, but the operator doesn't seem to return! I am probably not giving it the right parameters, could you give some example ones for the small dataset?

cquammen:
It's a compute-intensive algorithm. Try reducing the number of iterations to some small single-digit number instead of 100.

cquammen:
Thanks for the review, @cjh1!

PR #772, opened 2016-11-23
Restore view camera settings to those saved in state files

cquammen:
The camera settings were getting clobbered after animation cues where
loaded. This commit saves the camera settings after the state has been
loaded but before the animation cues are set and then reapplies the
settings after the cues are loaded.

### DIFF
--- a/tomviz/ModuleManager.cxx
+++ b/tomviz/ModuleManager.cxx
@@ -30,6 +30,8 @@
 #include "pqApplicationCore.h"
 #include "pqDeleteReaction.h"
 #include "pqPVApplicationCore.h"
+
+#include "vtkCamera.h"
 #include "vtkNew.h"
 #include "vtkPVXMLElement.h"
 #include "vtkPVXMLParser.h"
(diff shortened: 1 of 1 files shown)


cquammen:
Ready for review, @cryos @cjh1 .

tomviz/ModuleManager.cxx
```diff
@@ -546,6 +572,35 @@ void ModuleManager::onPVStateLoaded(vtkPVXMLElement* vtkNotUsed(xml),
     }
   }
 
+  // Save camera settings for each view
+  vtkNew<vtkSMProxyIterator> iter;
+  iter->SetSessionProxyManager(pxm);
+  iter->SetModeToOneGroup();
+  for (iter->Begin("views"); !iter->IsAtEnd(); iter->Next()) {
+    // if (xmlName == "RenderView") {
+    vtkSMRenderViewProxy* viewProxy =
```
cjh1: Not sure i understand this comment or is it left over from dev?

cquammen: Yep, leftover, thanks for catching.

Issue #768, opened 2016-11-23
Crash when closing script window after deleting the transform


yijiang1:
1. Load a dataset.
2. Do a transform. For example, "Delete Slices".
3. Open the python script by double-clicking it in the pipeline.
<img width="500" alt="screen shot 2016-11-23 at 12 50 50 pm" src="https://cloud.githubusercontent.com/assets/13088588/20572992/cbec154a-b17b-11e6-8eb9-ef9c2cf44c97.png">

4. Go to main window and delete the transform (keeping the script window open).
<img width="500" alt="screen shot 2016-11-23 at 12 51 01 pm" src="https://cloud.githubusercontent.com/assets/13088588/20573006/d7473582-b17b-11e6-8ebb-9e2b8cf49635.png">

5. Go back to the script window, click "OK" or "Cancel". 
6. Tomviz crashes.

cquammen:
Thanks for the report. The script dialog is now closed if the operator is deleted.

PR #773, opened 2016-11-24
Guard against invalid access to the Operator

cquammen:
Addresses #768 .


cquammen:
Ready for review, @cryos @cjh1 . Thanks.

PR #774, opened 2016-12-01
Add Python operator widget to properties panel

cjh1:
This PR adds the OperatorWidget to the properties panel, which allows the user to update the properties of an operator without editing the Python code. Currently the widget is added to a QScrollArea as the widget doesn't seem to resize correctly, a following PR will be needed to correct this.


cjh1:
@cryos @cquammen PTAL

cryos:
This is a great change, I think there are a number of enhancements needed to enable the generated interface to fit in the narrow panel. This is a great step in that direction, and we can work on some of the other issues in follow up commits.

PR #779, opened 2016-12-01
  Refactor itk python

cquammen:


cquammen:
@cjh1 PTAL

cjh1:
@cquammen LGTM, one thought/suggest. What about making utils a package, then we could have a itk module called utils.itk? And general utils in utils/_init__.py? We could eventually separate out utils further and have a utils.vtk ...

cquammen:
That sounds reasonable to me, but I defer to @cryos .

cquammen:
@cryos Please weigh in.

PR #785, opened 2016-12-06
Update dataset generation scripts to use parameter passing

### LINKED ISSUE #777: Convert generate_dataset(...) functions to use parameter passing
Python scripts such as https://github.com/OpenChemistry/tomviz/blob/master/tomviz/python/RandomParticles.py still use templating we should do the same conversion we did for transform_scalars(...).

cjh1:
Fixes #777


cjh1:
@cquammen Thanks!

PR #786, opened 2016-12-06
Add support for setting progress message from Python operator

### LINKED ISSUE #782: Add label to progress bar dialog
Could we add a label to the progress bar dialog that displays messages and can be updated in python script?
This is very useful for reconstructions which often have several steps. I'd like to print messages such as "Initialization...", "Iteration 1/10" and "Slice 1/256".

cjh1:
This PR fixes #782. It adds support for setting a progress messages that can be used in the progress dialog title, here is an example: 

progress

It also makes the progress Python API a little more consistent using properties for the progress value and progress message. So update the progress and message looks like this.

```python

self.progress.value = 10
self.progress.message = 'We are getting there..'
```

Note it changes the Operator progress methods, again to make the name more consistent.

### DIFF
--- a/tests/cxx/OperatorPythonTest.cxx
+++ b/tests/cxx/OperatorPythonTest.cxx
@@ -135,14 +135,14 @@ TEST_F(OperatorPythonTest, update_progress)
     file.close();
     pythonOperator->setScript(script);
 
-    QSignalSpy spy(pythonOperator, SIGNAL(updateProgress(int)));
+    QSignalSpy spy(pythonOperator, SIGNAL(progressStepChanged(int)));
     TransformResult result = pythonOperator->transform(dataObject);
     ASSERT_EQ(result, TransformResult::COMPLETE);
 
     // One from applyTransform() and one from our python code
     ASSERT_EQ(spy.count(), 2);
 
-    // Take the signal emission from our python code
+    // Take the signal emitted from our python code
     QList<QVariant> args = spy.takeAt(1);
     ASSERT_EQ(args.at(0).toInt(), 100);
 
--- a/tests/cxx/OperatorPythonTest.cxx
+++ b/tests/cxx/OperatorPythonTest.cxx
@@ -163,8 +163,8 @@ TEST_F(OperatorPythonTest, update_progress_message)
 
     QSignalSpy spy(pythonOperator,
                    SIGNAL(progressMessageChanged(const QString&)));
-    bool result = pythonOperator->transform(dataObject);
-    ASSERT_TRUE(result);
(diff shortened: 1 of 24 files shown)


cquammen:
I tried this out on the Invert Data operator by adding a progress message:

```
import tomviz.operators

NUMBER_OF_CHUNKS = 10


class InvertOperator(tomviz.operators.CancelableOperator):

    def transform_scalars(self, dataset):
        from tomviz import utils
        import numpy as np
        self.progress.maximum = NUMBER_OF_CHUNKS

        scalars = utils.get_scalars(dataset)
        if scalars is None:
            raise RuntimeError("No scalars found!")

        result = np.float32(scalars)
        max = np.amax(scalars)
        step = 0
        for chunk in np.array_split(result, NUMBER_OF_CHUNKS):
            if self.canceled:
                return
            chunk[:] = max - chunk
            step += 1
            self.progress.value = step
            self.progress.message = 'Step %d' % step

        utils.set_scalars(dataset, result)
```

The progress message does not appear in the progress dialog, but it is printed to stdout.

cjh1:
@cquammen I rebased on master and removed the debug. Try adding a time.sleep(1) in the invert operator it make be finishing to quick to render the title?

cquammen:
@cjh1 When run on a large enough data set, Invert Data runs slow enough to watch the progress bar update about once per second, so it is slow enough to see the progress update.

If I resize the progress dialog while it is running, the title does get updated with the expected message, so it looks like a repaint issue.

cjh1:
@cquammen I added an update after setting the title, can you try that?

cquammen:
Same behavior.

cjh1:
@cquammen take two?

cquammen:
Nope.

cjh1:
:frowning: I will have to go find myself a mac ....

cquammen:
`QDialog::repaint()` didn't work either.

cquammen:
Clarification on what I said earlier: I have to resize the dialog so as to cover or expose some part of the title to get the title to refresh. Simply resizing the dialog is insufficient to get the title to update.

cryos:
I think this is an OS specific issue, and setting messages in the status bar will work well (and it works as expected on all OSes).

cjh1:
@cquammen I couldn't get the title update correctly on the mac. So I am now showing the progress messages in the status bar as well.

yijiang1:
> I see the progress messages in the status bar on the Mac.

Where is the status bar?

cryos:
It is the bar at the bottom of the user interface (a bar across the bottom of the main window with the test appearing at the left side). On Windows you should see the message appear in both the status bar and the title of the progress dialog.

PR #791, opened 2016-12-07
Separate labeled object analysis from connected components operator

cquammen:


cjh1:
I like the new request review GitHub feature.
I am getting the following exception from ITK:

Exception encountered while running Label Object Attributes
Cannot get an instance of NumPy array.
Failed to execute the script. (/home/cjh/work/source/tomviz/tomviz/OperatorPython.cxx:348, virtual bool tomviz::OperatorPython::applyTransform(vtkDataObject*))
  File "Label Object Attributes", line 67, in transform_scalars
RuntimeError: Cannot get an instance of NumPy array.

cquammen:
Hmm, haven't seen that one. Can you briefly outline the steps you used to test it? (Sorry I didn't provide any)

cryos:
That is a really cool feature. I can take a look at this in the morning, thanks for pointing out the new feature!

cjh1:
I loading a dataset and then ran the operator, am I supposed to run another operator before it?

cquammen:
Which data set? I'm curious to the array type.

I imagine this operator to normally run on a dataset that results from a segmentation operation. So, integral type with different objects labeled with different integral values.

cjh1:
If I print the stacktrace I get:

```
  File "Label Object Attributes", line 28, in transform_scalars
  File "/home/cjh/work/build/ITK/Wrapping/Generators/Python/itkLazy.py", line 44, in __getattribute__
    itkBase.LoadModule(module, namespace)
  File "/home/cjh/work/build/ITK/Wrapping/Generators/Python/itkBase.py", line 49, in LoadModule
    if sys.version_info >= (3, 4):
RuntimeError: Cannot get an instance of NumPy array.
```

cjh1:
SMALL_PtCU...

cquammen:
Thanks. I couldn't reproduce the error by:

* Loading SMALL_PtCu_NanoParticle
* Applying "Label Object Attributes"

That's kind of a confusing error. What version of ITK are you running?

cjh1:
master from October (e19ddf0c)

cquammen:
Ah, okay. I'm back on ITK 4.9.0 (the only version I've been able to get to work locally). If you get a chance, can you do the following:

* Launch Tomviz
* Open the Python Console
* Run the following

```
import itk
itk.LabelImageToShapeLabelMapFilter.values()
```

What do you get for output?

cjh1:
```python>> import itk
>>> itk.LabelImageToShapeLabelMapFilter.values()

[<class 'itkLabelImageToShapeLabelMapFilterPython.itkLabelImageToShapeLabelMapFilterIUC3LM3'>, <class 'itkLabelImageToShapeLabelMapFilterPython.itkLabelImageToShapeLabelMapFilterIUS2LM2'>, <class 'itkLabelImageToShapeLabelMapFilterPython.itkLabelImageToShapeLabelMapFilterIUS3LM3'>, <class 'itkLabelImageToShapeLabelMapFilterPython.itkLabelImageToShapeLabelMapFilterIUC2LM2'>]
>>> 

```

cjh1:
should I move back to 4.9, not sure why I am on master?

cquammen:
Thanks for running that test. That's what I would expect to see at the console. It's odd because according to the traceback you have provided that command is leading to the NumPy array error, unless I'm reading that wrong.

I wouldn't roll back to 4.9.0. Ideally, I would get closer to your checked out commit, but I'm not quite willing to endure the pain of a rebuild just yet.

If you run the Connected Components operator on Tomviz master, do you get a similar error?

cjh1:
Yes, master 'Connected Component' gives the same error.

cquammen:
Ah ha, that's an interesting clue. I should probably update my ITK after all to make sure I have all the wrapping updates. Thanks for all your help with this so far!

cjh1:
No problem!

cryos:
I do not see any errors on my machine, although I am trying to remember what the best way to use this operator is - what is the model pipeline?

cquammen:
* Load data
* Apply Connected Components operator
* Apply Label Object Attributes operator

cryos:
Then just look at the table? I thought we had this coloring the nanoparticles in the LARGE data sample? I tried running it there, and got 

Exception encountered while running ConnectedComponents
/home/marcus/src/ITK/Modules/Core/Common/src/itkMultiThreader.cxx:420:
itk::ERROR: MultiThreader(0x7fc9d8eaf780): Exception occurred during SingleMethodExecute
/home/marcus/src/ITK/Modules/Segmentation/ConnectedComponents/include/itkConnectedComponentImageFilter.hxx:347:
itk::ERROR: ConnectedComponentImageFilter(0x7fc9d9461450): Number of objects greater than maximum of output pixel type

It also ate up all my memory, and started swapping (24GB of RAM). Maybe I am missing something obvious on my system.

cryos:
https://blog.kitware.com/segmentation-and-analysis-features-in-tomviz/ shows what I was thinking of, but I can't seem to make that work on the LARGE data sample.

cquammen:
Okay, so the pipeline I provided just happens to work for the SMALL dataset because there are relatively few connected components with the raw labeled data. For the LARGE dataset, Binary Threshold should be applied first to minimize the number of connected components probably arising from background noise. The pipeline should look like

* Load data
* Binary Threshold (or Otsu threshold if you don't want to supply the thresholds manually)
* Connected Components operator on the child data set produced by the threshold
* Label Object Attributes operator on the connected components output
* Double-click the child table produced by the above operator to view it in the spreadsheet view

cryos:
So there is no way to apply the label colors to the contours anymore? You can just look at it in a table? I see the original unsigned char go through binary threshold, result in an unsigned short, but not sure what the values represent (0 to 1329 in this case). Connected components operates on that, and then I do label object attributes as in the screenshot.
image

cquammen:
> So there is no way to apply the label colors to the contours anymore?

Not since label maps started being stored as child datasets. Coloring contours by the label value was fairly easy when label maps were stored as a second array in the dataset - that was kind of a hack that was used because we didn't support child datasets. Now that label maps are distinct child datasets, we would have to do some additional work to provide the child dataset to the Contour module for coloring. It shouldn't be too hard to implement. Perhaps a combobox that lists valid child datasets containing labeled voxels could be presented to change the coloring? I suggest we create a separate issue and have that discussion there.

> I see the original unsigned char go through binary threshold, result in an unsigned short, but not sure what the values represent (0 to 1329 in this case).

0 represents background, 1 to 1329 are the labels for each of the 1329 connected components. 1 is assigned to the smallest component, 1329 is assigned to the largest component.

> I think we need a reasonable way to take an isosurface of the data, and apply the label colors generated by the labeled connected components.

Let's take that up in another issue if you don't mind.

> I am also not sure I see the value in applying connected components on its own create a label map.

The value is that Connected Components is a common operation in image analysis, the results of which may be used in a number of ways. Currently, the only way we have imagined using it in Tomviz is computing attributes (size, surface area) and for coloring particles so that they are distinct. It's not too hard to imagine other uses - isolating the largest connected component by applying this operator followed by a Binary Threshold, for example, is the first that comes to mind. Then you might want to find the surface area of this largest connected component after all, for which you could use Label Object Attributes. For another example, you might want to apply the Label Object Attributes operator to a label map dataset that comes from an entirely different set of operations that doesn't involve the Connected Components operator.

> I don't want to see the voxels in a countour as are clearly visible in this data, but the contour of the raw data with the result of a label map.

Fair enough, but orthogonal to this merge request.

> Would the label object attributes be better as a helper in the ITK Python code that is always applied to connected components rather than separating them out.

You could do that, but I don't think it is desirable to always calculate the label object attributes of the connected components. Computing the label object attributes comes with some computing expense. I contend that there is value in separating out the Label Object Attributes as its own operator so that it may be applied to label map-producing operators other than Connected Components.

cryos:
Thanks for the detailed reply, there are certainly some higher level issues that GitHub isn't ideally suited to discussing. I will add an issue for coloring a contour by a label map, I think that is an important user-visible feature that we should prioritize restoring, and I wasn't sure if I was missing the new way of doing it.

I want to make sure our operators are at the right level of granularity, and appreciate the examples of isolated use I had not considered. I still fear the common use case if to always label them, and wouldn't want to make the majority of users add two operations every time they use this - hence my question.

Issue #648, opened 2016-10-04
Progress dialog does not appear when applying operator to child data set


cquammen:
Steps to reproduce:
- Load data set
- Data Transforms -> Binary Threshold, accept default settings
- Select "Thresholded Segmentation" child data set
- Data Transforms -> Otsu Multiple Threshold

Notice that the progress dialog does not appear

cjh1:
Fixed by #811

PR #800, opened 2016-12-15
Label object distance from axis

cquammen:
Added Label Object Distance From Principal Axes operator

Computes the distance from the centroid of each connected component in a label object with the given label_value to the given principal axis and store that distance in each voxel of the label object connected component. A principal_axis of 0 is first principal axis, 1 is the second, and 2 is third.


cjh1:
@cquammen Code looks good can you give me the recipe to test?

cquammen:
Sure, sorry for leaving that out.

* Load SMALL_PtCu_NanoParticle
* Data Transforms -> Binary Threshold, default parameters
* Click on Thresholded Segmentation child dataset. Data Transforms -> Label Object Principal Axes, default arguments
* Data Transforms -> Label Object Distance from Principal Axes, default arguments
* Add Volume module to the Thresholded Segmentation child dataset.
* Add a Contour module to the SMALL_PtCu_NanoParticle, Select Color checked to white, set Opacity to 0.04.

You should see the following:

image

The first principal axis of the object runs left to right in the image. Small connected components close to this axis have colors from the low end of the color map. The largest connected component centroid is very close to the centroid of the entire object, so its distance is near zero, hence it maps to the background color, i.e., is not visible.

cjh1:
@cquammen Thanks for the instructions, I am not seeing a progress dialog for 'Label Object Principal Axes' any idea why or is that by design?

cquammen:
For some reason the progress dialog does not show when operators are applied to child datasets (#648).

PR #808, opened 2016-12-21
Add a Ruler visualization module

mathturtle:
@cryos This is still WIP... there is a crash when it is deleted that I haven't tracked down yet.  I'm making the ruler a visualization module for now (it really should be something like a module but not associated with any data source... thats a lot of extra refactoring I may do later if we really want it).

remaining issues:
* ruler module needs an icon (I left it blank in this first pass)
* There is a crash when deleting the module, something in the UnRegisterProxy function goes wrong. I plan to debug it this afternoon and I'll update with progress.

### DIFF
--- a/tomviz/CMakeLists.txt
+++ b/tomviz/CMakeLists.txt
@@ -94,6 +94,8 @@ set(SOURCES
   ModuleOutline.h
   ModulePropertiesPanel.cxx
   ModulePropertiesPanel.h
+  ModuleRuler.cxx
+  ModuleRuler.h
   ModuleSegment.cxx
   ModuleSegment.h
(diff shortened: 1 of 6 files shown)


mathturtle:
So the crash has to do with closing tomviz while the ruler module is present... deleting the module via the context menu doesn't crash.  So I put code in the main window's close event to delete the module before closing it and the crash is "fixed".  We should look more at this when redoing the shutdown/close sequence.

mathturtle:
I was looking through the ParaView icons... what do you think of these for the Ruler module icon?

pqmeasure24
pqruler16

cryos:
I prefer the second one, I will try and take a look at this soon.

PR #811, opened 2017-01-02
Ensure progress dialog is displayed for child data

cjh1:
Add signal to ensure that operators added to child data sources display a progress dialog.


cjh1:
Fixes #648

PR #812, opened 2017-01-02
Delay adding child data source until operator is complete

cjh1:
Builds on #811


cjh1:
Fixes #679

cryos:
I am going to merge this, it looks reasonable to me and I would like to see how the dashboards do.

PR #815, opened 2017-01-11
Use BlockingQueuedConnection for operationStarted

cjh1:
This means that the worker thread blocks until the slot returns.
So ProgressDialogManager gets to connect up the correct signals
and show the progress dialog before the worker starts running the
operator so we don't lose any progress messages.


cjh1:
@yijiang1 Can you test out this branch? I think it may fix the missing progress messages you where seeing.

cjh1:
Excellent, thanks for the testing.

Issue #802, opened 2016-12-16
Progress animation does disappears when its row is selected on Mac


cquammen:
When an operator is running, select the operator's row in the Pipelines panel. You cannot see the progress animation.

mathturtle:
Is it just that the progress dialog is behind the main window?  That is a known issue on OSX, dialogs are not kept in front of their parent and it is a bug in Qt (if it is a bug).

cquammen:
No, I mean the little spinning icon in the Pipeline panel, not the progress dialog.

![image](https://cloud.githubusercontent.com/assets/360056/21274225/06fc25dc-c395-11e6-988c-a8fb7d2491c7.png)

cjh1:
I had seen that, not sure why

cryos:
This is now fixed

PR #818, opened 2017-01-12
Call QItemDelegate::paint(...) before painting progress

cjh1:
This ensures that the progress icon is painting on top of any selection style changes.

Note: Currently this looks a little bad as the spinner seems to have a lot white around the edges that shows up when a row is selected, this needs to cleaned up.


cjh1:
Fixes #802

cjh1:
@cquammen I have pushed another commit that should remove the goofy look ...

PR #820, opened 2017-01-16
Remove/hide histogram when data source is removed

### LINKED ISSUE #816: Histogram remains even after dataset has been removed
After removing a dataset from the pipeline its histogram remain. See screenshot below.

screen

cjh1:
@cryos You know the charts better than I so may know of a better way to do this, but this gives us the desired effect. Fixes #816

### DIFF
--- a/tomviz/vtkChartHistogram.cxx
+++ b/tomviz/vtkChartHistogram.cxx
@@ -138,6 +138,29 @@ void vtkChartHistogram::SetHistogramInputData(vtkTable* table,
 {
   this->HistogramPlotBar->SetInputData(table, xAxisColumn, yAxisColumn);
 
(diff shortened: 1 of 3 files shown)


tomviz/vtkChartHistogram.cxx
```diff
@@ -138,6 +138,29 @@ void vtkChartHistogram::SetHistogramInputData(vtkTable* table,
 {
   this->HistogramPlotBar->SetInputData(table, xAxisColumn, yAxisColumn);
 
+  // vtkPlotBar doesn't seem to behave well when given a null table,
+  // so we just hide the components.
+  auto setItemsVisible = [this](bool vis) {
```
cquammen: Nifty!

Issue #793, opened 2016-12-08
Support coloring contours by nearest label map value


cryos:
We wrote a blog post showing this feature off, and it would be great to bring it back. We might want to break this out into more issues but going beyond coloring by the ID I would like to color by radius, volume, surface area, or other calculated values from the connected components label for example.

tjcorona:
This issue is partially addressed by MR #941.

cquammen:
@tjcorona Do you have a set of steps to run in Tomviz to show off the patch in #941, e.g., load data set, apply transform, create contour, set color map range, etc. I'm having trouble getting something reasonable visualized.

tjcorona:
Sure! I just followed your rubric: https://github.com/OpenChemistry/tomviz/pull/823

cquammen:
Another issue: we should disable point-to-cell data and nearest neighbor interpolation when coloring by "This Data".

![image](https://cloud.githubusercontent.com/assets/360056/23527484/20a72f76-ff64-11e6-97bc-e7aea01d24a1.png)

tjcorona:
I agree. Would you like me to put up a MR for it?

cquammen:
Sure. I'm happy to test it.

cryos:
I think we have resolved a lot of issues here, please reopen if I am mistaken.

PR #823, opened 2017-01-16
Enable coloring contours by data from a child dataset

cquammen:
When operators that computed label maps were first added to Tomviz, it was possible to change the scalar coloring of contours to the labeling. The refactoring of child datasets to being added as separate data sources removed this capability. This branch restores it by adding a combo box to the Contour module's property panel to enable selection of a child of the current dataset. The selected child dataset is resampled onto the contour and set as the color array, restoring the ability to color a contour by a label map.

To test (takes a while):

* Load LARGE_PtCu_NanoParticles.tiff
* Run Otsu Multiple Threshold, default parameters
* Add Contour module
* Select Label Map child dataset from Otsu Multiple Threshold
* Run Connected Components
* Select Contour module
* Set the "Color By" combo box to "Label Map"
* Rescale color map data range to [600, 822]

Expected result:

image

### DIFF
--- a/tomviz/Module.cxx
+++ b/tomviz/Module.cxx
@@ -91,9 +91,11 @@ bool Module::initialize(DataSource* data, vtkSMViewProxy* vtkView)
     // FIXME: we're connecting this too many times. Fix it.
     tomviz::convert<pqView*>(vtkView)->connect(
       m_activeDataSource, SIGNAL(dataChanged()), SLOT(render()));
-    this->connect(m_activeDataSource,
-                  SIGNAL(displayPositionChanged(double, double, double)),
-                  SLOT(dataSourceMoved(double, double, double)));
+    connect(m_activeDataSource, SIGNAL(dataChanged()), this,
+            SIGNAL(dataSourceChanged()));
+    connect(m_activeDataSource,
+            SIGNAL(displayPositionChanged(double, double, double)),
+            SLOT(dataSourceMoved(double, double, double)));
   }
   return (m_view && m_activeDataSource);
 }
--- a/tomviz/Module.h
+++ b/tomviz/Module.h
@@ -138,6 +138,9 @@ protected:
   virtual vtkSMProxy* getProxyForString(const std::string& str) = 0;
 
 signals:
+  /// Emitted when the represented DataSource is updated.
+  void dataSourceChanged();
+
   /// Emitted when the UseDetachedColorMap state changes or the detatched color
   /// map is modified
   void colorMapChanged();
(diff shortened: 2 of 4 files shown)


cquammen:
Addresses #793

cjh1:
Tested using the instruction provided and get the same result. One issue I did notice is that when I add the contour module I get a strange unresponsive dialog that hangs around, see screenshot below:

dialog

Are you seeing that too? Not sure its related to these changes.

Code looks good.

cquammen:
@cryos Want to weigh in on the UI choice of a combobox listing the child data sets? I'm open to other suggestions.

cryos:
I was just about to take a look at this branch, building now.

cquammen:
> Are you seeing that too? Not sure its related to these changes.

@cjh1 I was not seeing that. Hmm, could you confirm that window doesn't pop up in master? I'm a little worried it is an errant vtkRenderWindow.

cjh1:
@cquammen Just checked on master and I'm not seeing the window when I add the contour module.

cquammen:
Hrmmm... I'll have a look.

cryos:
This caused my poor 24GB desktop to swap a lot, so I don't know if the box popped up (didn't notice it). I do see (and your screenshot seems to show some of that too) multiple colors per particle which isn't what we got before when choosing to color by the label map:

image

cryos:
It also looks like we are showing the wrong color map for the contour, but can view it when clicking on the connected components. This is something we should probably track in an issue, and not pin on this PR, but I see the same thing in your screenshot.

cryos:
Sorry, distracted by a ParaView/VTK bug. I think if you change the color by attribute in the contour then it needs to select the data set being colored by rather than the source as you want to manipulate the transfer function for the active data set being used for coloring. This may require some bigger changes.

It is odd, if I visualize the volume of the labels then it has the colors, but they lack any of the texture of the data. I think even for the volume I would want to see the original volume with the label map controlling the colors. Maybe it would be useful to go over how Slicer and other tools are doing this, as I see some great visuals that appear to be colored by label maps.

I think this is a good exploration, but would like to refine the interaction and visualization output further before we merge it in.

cryos:
Just pinging on this, we can merge what we have and iterate. I would like to figure out what we can do with labels and visualization as this is an important area once we have segmented data that people often want to show.

cquammen:
I'm okay with merging as is if you wish. I probably won't get back to this for a few days at the earliest.

cryos:
Let's loop back to this in a couple of days, I am fine merging what we have but would like to figure out how we plan to get this into a more generally useful form for our end users.

cquammen:
Yep, I have some ideas how to do this and may be able to get to this sooner.

PR #824, opened 2017-01-17
Force integral types for Otsu threshold output labels

cquammen:
Limitations in the Otsu threshold filter wrapping meant that the
output of the Ostu threshold filter had to be the same as the input.
For input images of type float or double, this produced output
labelmap datasets with float or double labels, respectively. Filters
such as connected components would not work with float or double
labels, so we now cast the output of the Ostu threshold filter to an
integral type (unsigned char) if it otherwise is not.

### DIFF
--- a/tomviz/python/OtsuMultipleThreshold.py
+++ b/tomviz/python/OtsuMultipleThreshold.py
@@ -8,6 +8,8 @@ def transform_scalars(dataset, number_of_thresholds=1,
 
     try:
         import itk
+        import itkExtras
+        import itkTypes
         import vtk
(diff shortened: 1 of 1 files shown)


cquammen:
Thanks! Should have mentioned how to test.

Sample Data -> Generate Random Particles, default settings
Data Transforms -> Otsu Threshold
Select child data set from Otsu Threshold
Data Transforms -> Connected Components

Should fail in master, pass with this change.

cjh1:
I can confirm that this fails on master and is fixed on the branch, thank @cquammen for the test instructions.

Issue #817, opened 2017-01-12
White Background As Default


Hovden:
I'd like to suggest a white background as the default. I prefer using a white background as the default for a couple reasons:
(1) It better suits use in a figure for scientific publication (we usually work on white)
(2) The UX looks nicer.
(3) Grey can be a difficult color to reproduce without the hex value.

<img width="1524" alt="screen shot 2017-01-12 at 1 03 57 pm" src="https://cloud.githubusercontent.com/assets/6197997/21901898/c52dc078-d8c7-11e6-94c4-e89f0db41a20.png">
<img width="1524" alt="screen shot 2017-01-12 at 1 04 23 pm" src="https://cloud.githubusercontent.com/assets/6197997/21901901/c7430e2c-d8c7-11e6-96ee-545971d118ad.png">

ercius:
I don't like the gray either, but I prefer black rather than white. Volume
visualizations seen to be better visible than on the gray. Ill try white
out to see how it looks though. Ultimately, adding a option to set the
default per user would be best.

Also, I always thought it interesting that data presented at conferences
prepared by Avizo was obvious due to blue default background. It's a subtle
advertisement for Avizo.

Hovden:
I use white for manuscripts (and web content), but often use black for presentations. There is a Viz argument for grey (e.g. Maya uses gray as default).

I agree about the advertisement. The biomed community is famous for really nice 3D Viz in their publications and they are most often on white backgrounds. I do think the EM community is advertising for Avizo when they use the blue gradient. It also suggests that people usually just use the default parameters (see also Digital Micrograph plots).

ElliotPadgett:
I would also vote for white, since it is most common (universal?) for manuscripts, and at least as common as dark backgrounds for web and presentations.

ercius:
I just tested white vs black. I'm a convert to white.

cquammen:
Please take a look at the latest master branch. It has a tool bar for picking the default ParaView color palette, a white background color palette, and a black background palette.

![image](https://cloud.githubusercontent.com/assets/360056/22112247/384fa7c0-de30-11e6-902c-f128376f887c.png)

If this is satisfactory, please feel free to close this issue.

yijiang1:
1. When first loading tomviz, I still see the old background.
2. Click color palette -> White Background. The background becomes white.
3. Click color palette -> Make Current Palette Default.
4. Click color palette -> Default Background. The background becomes the old one.
Furthermore,
[...]

PR #826, opened 2017-01-17
Use white background color palette

cquammen:
ParaView has the concept of color palettes, or selections of colors
for different elements in a typical scene (background, foreground,
text) that go well together. Make the color palette intended for
printing on white paper the default.

Addresses #817


cquammen:
Note that more than one color palette could be exposed via a combo box, for instance, so folks could switch from a printing color palette to a "presentation" color palette where the background is black.

cquammen:
Screen shot with default settings proposed in this patch:

image

cryos:
I would like to expose the palette we have now as the default, a printing palette such as this, and one with a black background. I would also like to make it simple for our users to set the default, and make sure the choice is saved in the state file.

cquammen:
Ready for another look, @cryos . The color palette tool button has a palette icon. Clicking on it brings up a menu of color palettes including default, black background, and white background. The current palette can be saved as default. Saving and loading to a state file appears to work.

PR #829, opened 2017-01-18
SVG spinner to PNG

cjh1:
@cryos this should alleviate your concerns about pulling in the SVG library.


mathturtle:
This was only an issue for the Linux binary and I can look at adding the qt-svg library to the Docker image if you want.

cjh1:
@mathturtle Don't worry about it, @cryos was concerned about pull in svg anyway and this seems to work fine, thanks for the offer.

PR #831, opened 2017-01-19
Change name of background to reflect its color

cryos:


cryos:
Yes, I think that would be useful, added (although not showing up here for some reason).

PR #835, opened 2017-01-23
Add support for loading image stacks

cjh1:
This PR adds support for loading image stacks. It allows multiple files to be selected from the file load dialog and makes to appropriate changes to display wildcarded versions of the filename on the data source and in the recently loaded files menu.

The allowing data set can be using to test this functionality:

https://data.kitware.com/api/v1/item/58867cf58d777f4f3f3084be/download


Hovden:
Great, useful feature.

cryos:
The same crash is happening with a normal TIFF volume being restored. We should open an issue, but this has nothing to do with the TIFF stack addition here.

Issue #837, opened 2017-01-25
Add support for MetaImage and NRRD Image IO


thewtex:
I am having a hard time creating 3D images I can view in tomviz.

Most scientific processing tools, especially the ITK-based ones, have support for MetaImage (.mha) and NRRD (.nrrd) formats. They are N-Dimensional, support a variety of pixel types, and have good support for associated image metadata like the origin, etc. I see ParaView has support for these, so hopefully it would not be too much work to add them to tomviz.

cryos:
I have never heard of these formats, if it is a matter of exposing the ParaView readers it shouldn't be a big deal. We can take a look, but @cjh1 just added support for reading TIFF image stacks, and TIFF volumes have worked well for quite some time. There is also XDMF (HDF5 based) that many people use.

Where are these file formats generated? What domains are you thinking of? It isn't something I have come across as a ParaView developer either.

thewtex:
MetaImage is the most common format I work:

  https://github.com/Kitware/MetaIO

NRRD is pretty common the medical domain:

  http://teem.sourceforge.net/nrrd/format.html

NRRD may not be as applicable to tomviz, but MetaImage is extremely common in the ITK/VTK community.

thewtex:
Confirmed that the TIFF stacks work with the latest Nightly binary :+1: .

Still would like MetaImage support ;-)

cryos:
We will look into it, do you know where some sample files with volumes in them might be? I have seen that library name, but know little else about them. Looking at the links they seem very medical which might explain it :-) We have a simple file reader white list, so it shouldn't take much work to test and add them once we have a sample file.

thewtex:
Thanks for the consideration.

Here's a file from the VTK test suite:

  https://midas3.kitware.com/midas/folder/10149

There are a bunch more:

  https://midas3.kitware.com/midas/search/index?q=.mha

cryos:
So we have MetaIO, it should be in one of the next binaries. We should see if we can do the same for NRRD, https://midas3.kitware.com/midas/search/index?q=.nrrd doesn't show any obvious volumes, is there a good sample file you have to hand there?

thewtex:
Thank you!

Here's a testing dataset for NRRD:

  https://data.kitware.com/#item/588275628d777f4f3f30816e

cquammen:
[...]

PR #839, opened 2017-01-25
Enable reading MetaIO image files

mathturtle:
@cryos @thewtex For #837, enable the MetaIO reader in tomviz.  I tested it with one of the VTK test data files.


thewtex:
@mathturtle Awesome, thanks!

I saw that .mhd was already supported in writing -- could .mha be added there, too?

Issue #764, opened 2016-11-21
Create child datasets in python script


yijiang1:
Is it possible to have a utility function that creates child datasets in python script (.py)?

It seems that child datasets creation is currently done in json script. However, not all transforms (especially tomography reconstruction) can use json script.

cquammen:
You'll likely still need to define the JSON for the Operator to let Tomviz know a child data set is expected. The missing piece is being able to specify that a custom UI is used.

mathturtle:
So I'm going to use this issue to discuss what I was asked at the meeting... convert the reconstruction operators to create child datasets.  This means converting them to use JSON files.  @yijiang1 I'm curious why you don't think they can use JSON files.

I have a question about the weighted back projection operator.  Right now the output size is a parameter and defaults to the Y size of the input image.  @cquammen is there a way to set a default like this with the current JSON files or do we need to add something to the format?

yijiang1:
> I'm curious why you don't think they can use JSON files.

Some reconstruction methods have very complicated UI like this
<img width="500" alt="screen shot 2017-01-13 at 4 49 15 pm" src="https://cloud.githubusercontent.com/assets/13088588/21946893/63f8a0d6-d9b0-11e6-9db7-28e4ba6f7a52.png">

I have no problem using JSON files if they can create this kind of UI.

cquammen:
Two approaches:

1). Remove the custom UI and replace it with one generated in JSON. You couldn't reproduce the exact UI shown here with the JSON, but you could probably get close. The labels can contain long descriptive strings, but they currently show up just on the left-hand side while the widgets show up on the right side.

2). Add to the JSON description something that indicates a custom UI is to be used. The logic would need to be added to AddPythonTransformReaction.cxx. You'll also need to handle passing parameters from the UI to the script.

2 seems like the better option to me because there will be some custom UI's that will be too awkward to express in JSON.

cquammen:
> I have a question about the weighted back projection operator. Right now the output size is a parameter and defaults to the Y size of the input image. @cquammen is there a way to set a default like this with the current JSON files or do we need to add something to the format?

No, there isn't anything to set defaults from the input data at the moment. It would be useful to have, though.

cryos:
This is now working.

PR #845, opened 2017-01-30
Recon child datasets

mathturtle:
This is most of #764... I have a commit that fixes the DFT_constraint one too (but not on this branch yet), but it changes the UI to make it look like the others, and the discussion on the issue indicates that we may want to keep it the way it looks now.

@cryos @yijiang1


cryos:
There are a few minor Python formatting issues reported, I will give this a try after lunch.

cryos:
Just testing this now, I think it will need a rebase, and it has some formatting issues flagged in the travis testing.

cryos:
It looks like we need some more work on the pipeline from the user perspective. I get something like the following, and the panel doesn't seem to know what it is, so nothing is shown. I am also unable to visualize the reconstructed volume, and see

    QObject::connect: Cannot queue arguments of type 'vtkSmartPointer<vtkDataObject>'
    (Make sure 'vtkSmartPointer<vtkDataObject>' is registered using qRegisterMetaType().)

yijiang1:
We also need to make simulated tilt series a child data set.

cryos:
This still doesn't work for me, I am loading the tilt-series, running the C++ reconstruction, and I see the same as I pasted above. I can only work with/visualize the parent data set. Same warning messages, this is on Linux.

mathturtle:
... Different test cases.  In my case I was making the tilt series via the python operator that creates one.  And the Python Operator's constructor registers the qMetaTypes that it needs.  But if you don't run a python operator first then the metatypes aren't registered and it fails.  I'm adding a commit to fix it.

PR #846, opened 2017-01-31
Disable the double slider widget for now

cryos:
It is causing an issue on Windows, and is not currently used as far
as I can tell.


mathturtle:
Yes, I had to change to use a custom subclass of vtkPropertyWidget since we wanted to set the width of the text fields.  See `tomviz/DoubleSliderWidget.{h,cxx}` for the replacement for this.  I guess I missed taking this out.

cryos:
I figured there was no point in tracking down the cause of the error if we don't use it, although it seems like an odd one that only hits Windows for any ParaView plugin - maybe a new issue for Qt 5.8.0 @utkarshayachit may want to be aware of when they move forward.

PR #847, opened 2017-01-31
Use a custom plist with keys to use retinal displays

cryos:


cryos:
This activates retina capability in the app bundle, addressing #19 with support from a number of patches merged into VTK and ParaView.

PR #852, opened 2017-01-31
If an operator is running ask before exiting tomviz

mathturtle:
This will help alleviate the bug I found where tomviz is exited during a transform and the worker thread is not cleaned up (you end up with a zombie background thread sitting in its event loop keeping the process alive).  This will help warn users away from this, but we need to fix it...  Putting in a `QCoreApplication::exit(1)` in the else clause of `if (response == QMessageBox::No)` didn't help.

### DIFF
--- a/tomviz/MainWindow.cxx
+++ b/tomviz/MainWindow.cxx
@@ -170,8 +170,7 @@ MainWindow::MainWindow(QWidget* _parent, Qt::WindowFlags _flags)
           SLOT(expandAll()));
 
   // connect quit.
-  pqApplicationCore::instance()->connect(ui.actionExit, SIGNAL(triggered()),
-                                         SLOT(quit()));
+  this->connect(ui.actionExit, SIGNAL(triggered()), SLOT(close()));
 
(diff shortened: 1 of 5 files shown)


tomviz/MainWindow.cxx
```diff
@@ -501,6 +501,15 @@ void MainWindow::showEvent(QShowEvent* e)
 
 void MainWindow::closeEvent(QCloseEvent* e)
 {
+  if (ModuleManager::instance().hasRunningOperators()) {
+    QMessageBox::StandardButton response = QMessageBox::question(
+      this, "Close tomviz?", "You have transforms that are not completed.  Are "
```
cryos: Just use on space after a full stop, if they click yes is there any hope of exiting cleanly? If not, then maybe we should consider warning about that.

cjh1:
@mathturtle This  may be of some help http://doc.qt.io/qt-5/qthread.html#terminate. We could add a terminate method on the PipelineWorker. I can take a look if you would like?

PR #853, opened 2017-02-01
Add web export

jourdain:


cjh1:
@jourdain I pushed a commit to exclude JavaScript form clang-format :smile:

jourdain:
Cool thanks... Obviously, I still need to fix some C++ formatting. Is there any good way to do it?

cjh1:
https://llvm.org/svn/llvm-project/cfe/trunk/tools/clang-format/git-clang-format

See https://github.com/OpenChemistry/tomviz/blob/master/STYLE.md for details

cryos:
Which ParaView/VTK modules/options need to be turned on to make this work? I tried VTK_Group_Web and that didn't work here. Would it be the PARAVIEW_ENABLE_WEB? I am giving it a try now, but wondered what you were using there.

cryos:
That did the trick, so I guess right now to test we need PARAVIEW_ENABLE_WEB turned on, I successfully generated a data folder in the specified folder, but I don't see any html, or an obvious way of looking at what was generated. I tried copying web/www/* over but see the error 

XMLHttpRequest cannot load file:///home/marcus/build/tomviz/web/data/index.json. Cross origin requests are only supported for protocol schemes: http, data, chrome, chrome-extension, https, chrome-extension-resource.

In my chrome console. That is where I stopped, but I guess it is saying I cannot simply open the index.html file and must use a web server protocol.

jourdain:
Just ParaViewWeb ON

On Sat, Feb 4, 2017 at 12:05 PM, Marcus D. Hanwell <notifications@github.com

cryos:
Chris showed me 'python -m SimpleHTTPServer' and I was able to test it out, it looks great. It seems to ignore the number of angles requested in my testing, but I was able to look at some exported volumes and get basic interaction working well. I only see the build flag PARAVIEW_ENABLE_WEB and couldn't find a module named ParaViewWeb.

It doesn't work in a Python 3 build, so that is something that would be nice to figure out in ParaView separately. Once we can get it copying over the tomviz JavaScript, index.html and loading using the file location in the browser it would be great to show off to our collaborators.

jourdain:
PARAVIEW_ENABLE_WEB is the right flag name. 

Regarding the `number of angles`, it is not the number of angles but what is the delta angle you want to have between two positions. I'll double check, to make sure it works for me. I you rather provide the number of steps instead of delta angle, that's fine too. Just let me know.

You can use it locally without a web server if you follow https://github.com/jourdain/tomviz/blob/3995a9bccd4a939bfec2206c44a8fee5c365f4ea/web/README.md

For the JavaScript, I need to figure out the path from where I should get the JS and HTML or at least the basepath of the application.

Right now, I'm trying to fix the formatting ;-)

cryos:
Is there nothing we can do in the HTML/JavaScript? Otherwise I think we should look at embedding everything in the index.html to avoid the request entirely. We want to let them double click on the index.html without passing flags/changing settings to really achieve the goal of easily sharing an interactive export of the data.

jourdain:
Those are browser security things therefore you can not change them within the JavaScript.

If we try to embed the data to the file, we will have to base64 it which will increase its size.
 
As an intermediate approach, we could try to package the data into a single ZIP file (which could have a tomviz specific extension) and require the user to drag and drop the file to the web page that they opened by double clicking on the index.html like we did here: http://kitware.github.io/vtk-js/examples/StandaloneSceneLoader.html

Once that part is working, we can try to embed that base64/zip into the index.html.

But this will require some rework of the data access code within the ParaViewWeb library so we can swap the backend from HTTP/url to ZIP/filePath.

cjh1:
@jourdain I will take a look why the clang-format checks are failing with: "Invalid symmetric difference expression master"

jourdain:
Cool thanks... I didn't even notice. I was looking at the flake output while trying to fix the copy of the JS/HTML at the same time.

cryos:
I think if we need to ask users to create custom shortcuts, change browser settings etc that is really a non-starter. Whatever we offer has to be something they can open up in an unmodified browser, I was hoping there might be a way to support this with the format you have. If dragging/opening a file is the best we can do then it sounds preferable to making them look up their browser and make a change to their settings/shortcut.

jourdain:
I agree. For me that two step process is the most vanilla one that I came up with so far.
Especially when you have binary content on the side.
 
Other options will require system specific application/batch/shell script to start a local web server just to get around that security limitation.

As for now, I'll try to wrap up my current branch and start focusing on bundeling the export into 2 files the `html application` and the `data` one.

jourdain:
@cjh1 @cryos can any of you review that part while I'm focusing on the loading of a zip file in the web app?

cjh1:
I am working on fixing up clang-format ...

jourdain:
Cool thanks, just wanted to let you know that the changes I made seems to be reasonable now.

For our records, this list all the command line options to start a web server.

https://gist.github.com/willurd/5720255

jourdain:
I disabled flake8 for the web/* directory which is the javascript one, not web.py inside tomiviz/...?

For some reason when I was running flake8 locally, it was grabbing the javascript. But I ran several time flake8 after I've excluded the web/* directory to fully fix the formatting issue on my web.py.

cjh1:
Sorry my mistake!

jourdain:
No worries, it is better to be sure. ;-)

cryos:
The path search logic doesn't seem to work, in my case I launch from the Tomviz build directory, and enter the path '/home/marcus/build/tomviz/web' as the output directory, which is empty, and then I get the following error where I am guessing it is attempting to search directories for the files to copy:

Generate/home/marcus/build/tomviz/web
  File "/home/marcus/build/tomviz/lib/site-packages/tomviz/web.py", line 18, in web_export
    copy_viewer(destinationPath)
  File "/home/marcus/build/tomviz/lib/site-packages/tomviz/web.py", line 43, in copy_viewer
    os.path.join(destinationPath, 'tomviz'))
  File "/usr/lib/python2.7/shutil.py", line 171, in copytree
    names = os.listdir(src)
OSError: [Errno 2] No such file or directory: '/home/marcus/build/tomviz/openchemistry/thirdparty/VTK/Web/Applications/FileBrowser/www/tomviz'

jourdain:
Hum, interesting. Could you try again but from the install tree, since I have some cmake rules for installing those files? It is possible that the build tree is incomplete and therefore fails. Moreover, I can try to be sure we find the right index.html as yours seems to have picked up one in the VTK/web/...

jourdain:
Ok, I've pushed a new web app for the drag and drop of a zip file.

I still need to fix the export to zip the directory but otherwise, we could live with only two files.

cryos:
Yes I was just working from my build tree, and I guess it kept walking up until it found something. I don't do an install when developing, if that is all that is missing then we can copy the relevant files to the build directory to make testing easier. I will give it a try when I am back in the office tomorrow.

jourdain:
Yes we can copy the file to the build tree. I guess we can just run another cmake command to do it.
But I'm not sure how to do it properly.

cryos:
If that is the only issue remaining I can take a look, I have done it for the Python files and other bits.

jourdain:
Yes for the image based export.

Otherwise I still need to figure out that segfault due to some GL code when we export the composite data.

jourdain:
We might be all set now even for the build tree. 

I guess we should talk tomorrow to validate the next priorities.

tomviz/SaveWebReaction.cxx
```diff
@@ -0,0 +1,100 @@
...
+  if (dialog.exec() == QDialog::Accepted) {
+    this->saveWeb(dialog.getOutputPath(), dialog.getExportType(),
+                  dialog.getDeltaPhi(), dialog.getDeltaTheta());
+  }
+}
+
+bool SaveWebReaction::saveWeb(const QString& filename, int type, int deltaPhi,
+                              int deltaTheta)
+{
+  cout << "Generate" << filename.toLatin1().data() << endl;
```
cjh1: Left over debug?

tomviz/python/tomviz/web.py
```diff
@@ -0,0 +1,167 @@
...
+
+# -----------------------------------------------------------------------------
+# Helpers
+# -----------------------------------------------------------------------------
+
+
+def zipData(destinationPath):
+    dstFile = os.path.join(destinationPath, 'data.tomviz')
+    dataDir = os.path.join(destinationPath, 'data')
+    zf = zipfile.ZipFile(dstFile, mode='w')
```
cjh1: Probably better to use context manager here, then you don't need the try finally block, something like this

```python
with zipfile.ZipFile(dstFile, mode='w') as zf
   # Do stuff with zip here
```

tomviz/python/tomviz/web.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+def get_proxy(id):
+    session = simple.servermanager.ActiveConnection.Session
+    remoteObj = session.GetRemoteObject(int(id))
+    return simple.servermanager._getPyProxy(remoteObj)
+
+
+def copy_viewer(destinationPath):
+    searchPath = os.getcwd()
```
cjh1: Performing a search based on the directory tomviz was started in seems a little fragile. ```__file__``` will give you local of this file, which you could used to orientate your self in the install tree,

jourdain: I started by that but, the recorded path was the one from the source tree not the build tree.

tomviz/python/tomviz/web.py
jourdain: Good idea, thx

tomviz/python/tomviz/web.py
```diff
@@ -0,0 +1,167 @@
...
+
+def add_scene_item(scene, name, proxy, view):
+    hasNormal = False
+    hasColor = False
+    colors = {}
+    representation = {}
+    rep = simple.GetRepresentation(proxy, view)
+
+    # Skip volume
+    if rep.Visibility == 0 or rep.Representation == 'Volume':
```
cjh1:
```python
not rep.Visibility 
```
is a little more pythonic ( I assume Visiibility is a boolean flag? )

tomviz/SaveWebReaction.cxx
jourdain: yes...

tomviz/python/tomviz/web.py
```diff
@@ -0,0 +1,167 @@
...
+        return
+
+    for prop in ['Representation']:
+        representation[prop] = rep.GetProperty(prop).GetData()
+
+    pdInfo = proxy.GetPointDataInformation()
+    numberOfPointArrays = pdInfo.GetNumberOfArrays()
+    for idx in xrange(numberOfPointArrays):
```
cjh1: I would use ```range``` here as xrange will go away with python 3.

tomviz/python/tomviz/web.py
```diff
@@ -0,0 +1,167 @@
...
+        array = pdInfo.GetArray(idx)
+        rangeValues = array.GetRange(-1)
+        if array.Name == 'Normals':
+            hasNormal = True
```
cjh1:
```python
hasNormal = arrayName == 'Normals'
```

tomviz/python/tomviz/web.py
```diff
@@ -0,0 +1,167 @@
...
+            else:
+                colors[array.Name] = {
+                    'location': 'POINT_DATA',
+                    'range': [i for i in rangeValues]
+                }
+
+    # Get information about cell data arrays
+    cdInfo = proxy.GetCellDataInformation()
+    numberOfCellArrays = cdInfo.GetNumberOfArrays()
+    for idx in xrange(numberOfCellArrays):
```
cjh1:
```
range
```

tomviz/python/tomviz/web.py
jourdain: No, it's a Proxy property which represents an Integer. But I guess that might work. 
Although I'm not sure how Python work with those wrapped properties.

jourdain:
Thanks Chris for those great feedback. I'll fix them tomorrow...

tomviz/python/tomviz/web.py
cjh1: OK, yes we symlink the python files into the build tree. In which case you may have to configuring a python file to contain CMAKE_BINARY_DIR and search using that as well as __file__ to cover both built and installed environments.

tomviz/python/tomviz/web.py
cjh1: Visibility sounds like it would 0 or 1

cryos:
We could simply pass the executable location into the Python script from the application, it always knows where it is started, and that is used in several places to get relative paths right.

cjh1:
@cryos Good idea!

cryos:
It gets further, but I get this error now and am left with just Tomviz.html in the directory I specified: 

    shutil.Error: `/home/marcus/build/tomviz/web/Tomviz.html` and `/home/marcus/build/tomviz/web/Tomviz.html` are the same file

cjh1:
Just as another data point I get the following:

``` python
  File "/home/cjh/work/build/tomviz-py2/lib/site-packages/tomviz/web.py", line 63, in copy_viewer
    shutil.copy(os.path.join(root, 'Tomviz.html'), destinationPath)
  File "/usr/lib/python2.7/shutil.py", line 119, in copy
    copyfile(src, dst)
  File "/usr/lib/python2.7/shutil.py", line 83, in copyfile
    with open(dst, 'wb') as fdst:
IOError: [Errno 21] Is a directory: '/tmp/test/'
Failed to execute the script. (/home/cjh/work/source/tomviz/tomviz/SaveWebReaction.cxx:93, bool tomviz::SaveWebReaction::saveWeb(const QString&, int, int, int))

```

cjh1:
@jourdain One way to achieve the single page use case we are aiming for would be to embed all the images in the html as separate elements ( hidden, base64 encoded ).  The data loading code could then lookup the appropriate element to get the data ( and decode ) for each image. One advantage of this approach over encoding a zip file is that data would be decoded as need, rather than up front having to decode the entire zip.

tomviz/SaveWebReaction.cxx
jourdain: ok

tomviz/python/tomviz/web.py
jourdain: ok

tomviz/python/tomviz/web.py
jourdain: So you are suggesting to use `__file__` as it will be better for the long run and just do a work around when running from the build tree?

tomviz/python/tomviz/web.py
jourdain: yes it should only be 0 or 1

tomviz/python/tomviz/web.py
jourdain: ok

tomviz/python/tomviz/web.py
jourdain: I don't agree as we are in a loop and we want to know if one of the array name was `Normals`.

tomviz/python/tomviz/web.py
jourdain: ok

tomviz/python/tomviz/web.py
cjh1: I would do what @cryos suggested, pass in the executable location.

tomviz/python/tomviz/web.py
cjh1: sorry I meant:

```python
hasNormal = array.Name == 'Normals'
```

jourdain:
@cjh1 I'm not sure it will be better to rely on a DOM to keep track of all the resources.

cjh1:
@jourdain Why?

jourdain:
Ok, I think I tackle all the suggestions you've listed except:

- Start with a better path for the search of `tomviz.html`
  - If the C++ is giving the executable path, how can I get that runtime information?
  - Is using `__file__` still an option?

- What is remaining:
  - Fix Python 3 support when enabling ParaViewWeb (We may want to target that fix before PV 5.3?)
  - Fix composite rendering crash?
  - Provide other types of export (which could be still image base.

jourdain:
Because the dataset could have thousands of thousands files. Which will translate with that many node in the DOM. While for the zip, it will just be a memory map.

jourdain:
And the parsing will still happen in the DOM too.

cjh1:
>  - If the C++ is giving the executable path, how can I get that runtime information?

```
QApplication::applicationDirPath()
```

>   - Is using `__file__` still an option?
This could only work for the install tree, so @cryos suggestion is a better solution

jourdain:
zip is just a binary list of resources without compression in our case.

jourdain:
Ok, I'll start using `QApplication::applicationDirPath()`, thx!

cjh1:
But the zip will have to be decoded first if its embedded into the html?

jourdain:
Yes but that will be quicker than thousand of thousand of nodes in the DOM that will be parsed at load time and decoded later.

cjh1:
OK, I was under the impression we where dealing with smaller numbers of images, currently the export seem generate around 40?

jourdain:
That's because we don't do volume exploration or composite export.

cjh1:
@jourdain To remove those python3 errors rebase on master ....

jourdain:
Do you mean ParaView/master has fix for ParaViewWeb and Python 3?

cjh1:
I mean the error travis is reporting on this branch.

jourdain:
Oh, ok ;-)

cjh1:
@jourdain So as a little test, I generate a html file containing 10,000 hidden DOM elements containing base64 encoded images. The document is 363M and loads in a couple of seconds. I then have some JavaScript that gets one of these divs by id and decodes the data, it can do this in less the 2 milliseconds on my machine. So this is definitely another approach that is feasible.

cjh1:
Export now works for me, looks good :smile: I noticed that it seems to reset the current camera position?

jourdain:
What do you mean it reset the current camera position. The export start with the current position and finish with the last one of the export, which obviously change the camera once done.

jourdain:
For the export of the data in the DOM, I was not saying it was not possible, I was just saying that it was not providing any kind of benefit.

cjh1:
Ideally it should restore the original camera position, but that is polish ...

Regarding DOM export, the benefit is you don't have to decode 300MB of data up front before you display anything ...

jourdain:
I get what you are saying but I don't believe that decoding 300MB up front is an issue (as we just read the meta data).
I guess the best test will be to try out and see if we get a bad load time or not.

cjh1:
Agreed

cryos:
Great debate, I think it would be good to explore this as you both agreed and see which approach gives the best end user experience. Glad to have you both thinking about this in depth.

jourdain:
@cjh1 @cryos , I think I'm all set.

Will probably need some polish, but I'll focus next on PV and Python3

PR #862, opened 2017-02-06
Update README.md

Hovden:


cryos:
I don't think we need this, and will have a fix shortly to work on older versions of macOS. We specify the minimum versions of each OS on the downloads page.

cryos:
Closing for now, I think we want to allow freedom and consider the requirement of macOS Sierra a bug. These feels a little too vague, and more specific information is already offered on the download page.

Issue #869, opened 2017-02-07
Crash when closing RenderView window


yijiang1:
Binary: 0.9.3-16-g6b94623 on Mac
1. Load a data set
2. Close RenderView by clicking the x button:
![picture1](https://cloud.githubusercontent.com/assets/13088588/22715245/14f720b8-ed5e-11e6-8878-3e2674d41a14.png)
3. Crash

yijiang1:
After #871 , clicking the X button no longer crash tomviz.

However, there is a new bug:
After closing the RenderView, information about the dataset still exists in the pipeline. Clicking it can crash tomviz.

cryos:
Yes, we were looking at this, the visualization modules need to be removed if a render view is removed otherwise things will crash. This was the simplest fix, but we will follow up with more fixes to resolve outstanding issues.

cryos:
This should now be resolved.

PR #871, opened 2017-02-07
Guard for when the representation is null

cryos:
This can happen when the view is closed for the module. We may want
to remove the modules if the view they belong to is closed too, but
this makes the module classes less prone to crashing in odd situations
lie this.


cryos:
This prevents crashes such as those described in #869.

cjh1:
I open this issue for the layout #874

PR #879, opened 2017-02-08
Added progress status to all auto tilt series alignment methods

### LINKED ISSUE #432: Progress indicator for Image Alignment (Auto)
It would be nice if there was a progress indicator of some type for the Image Alignment (Auto) data transform since it may take a long time if the user has a large dataset.  The python transform has a for loop in it iterating over images in the stack so there is a clear way to measure the progress.

yijiang1:
This closes #432

### DIFF
--- a/tomviz/python/AutoTiltAxisShiftAlignment.py
+++ b/tomviz/python/AutoTiltAxisShiftAlignment.py
@@ -40,9 +40,10 @@ class AutoTiltAxisShiftAlignmentOperator(tomviz.operators.CancelableOperator):
             shiftedTiltSeries = np.roll(
                 tiltSeries[slices, :, :, ], shifts[i], axis=1)
             for s in range(numberOfSlices):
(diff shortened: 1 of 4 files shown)


tomviz/python/AutoTiltAxisShiftAlignment.py
```diff
@@ -1,46 +1,61 @@
...
 
-    result = np.roll(tiltSeries, shifts[np.argmax(I)], axis=1)
+        for i in range(shifts.size):
+            if self.canceled:
+                return
+            shiftedTiltSeries = np.roll(
+                tiltSeries[slices, :, :, ], shifts[i], axis=1)
+            for s in range(numberOfSlices):
+                self.progress.message =
+                    'Reconstructing slice No.%d with %d pixels shift' %
```
yijiang1: How should I separate such long code?
autopep8 doesn't fix it for me :(

Issue #880, opened 2017-02-09
C --> Fortran array deep copy


ercius:
The console keeps printing out the following information:

> Warning, array does not have Fortran order, making deep copy and fixing...
...done.

This seems very inefficient. I am not sure why tomviz has to keep changing the array from Fortran -->C-->Fortran ordering. This happens for almost any numpy operation since numpy is C ordering.

As a test I modified the rotation script to print out the ndarray flags at various points:
```
print('Rotating Dataset...')
print(data_py.flags) #print the ndarray flags
print('...')
axis1 = (rotation_axis + 1) % 3
axis2 = (rotation_axis + 2) % 3
data_py_return = ndimage.interpolation.rotate(
    data_py, rotation_angle, axes=(axis1, axis2),order=1,mode='nearest')
print(data_py_return.flags) #print the ndarray flags
utils.set_array(dataset, data_py_return)
print('Rotation Complete')
```

Here is the output:
> Rotating Dataset...
  C_CONTIGUOUS : False
  F_CONTIGUOUS : True
  OWNDATA : False
  WRITEABLE : True
  ALIGNED : True
  UPDATEIFCOPY : False
  ...
  C_CONTIGUOUS : True
  F_CONTIGUOUS : False
  OWNDATA : True
  WRITEABLE : True
  ALIGNED : True
  UPDATEIFCOPY : False
Warning, array does not have Fortran order, making deep copy and fixing...
...done.
Rotation Complete

So, the array starts as Fortran and ndarray.interpolation.rotate returns a C-ordering array. Then the utils sets it back to fortran ordering. I guess tomviz internals (based on VTK?) need Fortran ordered arrays? It would be better if we can use 1 ordering to avoid deep copies.

As a side note: Some of the scripts seem to use fortran indexing [X,Y,num] rather than the numpy default C indexing [num,Y,X]. This might be correct initially, but some operations might return a C-ordered array, It is important to be clear of the array ordering during image-by-image operations (such as cross-correlation) because you can get a large speed up with the correct indexing.

cjh1:
@ercius #882 resolves this issue for rotation, but we will have to apply a similar fix else where.

cjh1:
Fixed by #880

PR #882, opened 2017-02-09
Preserve array order by passing in empty copy

cjh1:
Fixes issue reported in #880, a similar change need to applied to other operators.

### DIFF
--- a/tomviz/python/AutoTiltAxisRotationAlignment.py
+++ b/tomviz/python/AutoTiltAxisRotationAlignment.py
@@ -87,8 +87,9 @@ class AutoTiltAxisRotationAlignOperator(tomviz.operators.CancelableOperator):
         rot_ang = coarseAngles[minIntensityIndex]
 
         self.progress.message = 'Rotating tilt series'
-        result = ndimage.interpolation.rotate(
-            tiltSeries, -rot_ang, axes=((0, 1)))
+        result = np.empty_like(tiltSeries)
+        ndimage.interpolation.rotate(
+            tiltSeries, -rot_ang, axes=((0, 1)), output=result)
         #step += 1
         #self.progress.value = step
         print("rotate tilt series by %f degrees" % -rot_ang)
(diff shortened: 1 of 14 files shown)


cjh1:
I think I have fixed all the operators. @yijiang1 PTAL

tomviz/python/AutoTiltAxisRotationAlignment.py
```diff
@@ -87,8 +87,9 @@ def transform_scalars(self, dataset):
         rot_ang = coarseAngles[minIntensityIndex]
 
         self.progress.message = 'Rotating tilt series'
-        result = ndimage.interpolation.rotate(
-            tiltSeries, -rot_ang, axes=((0, 1)))
+        result = np.empty_like(tiltSeries)
```
yijiang1: If rot_ang is not a integer multiple of 90 (i.e. 90, 180, 270...), the output's shape will be different from input array.
Same issue for Rotate3D transform.

cjh1: Good catch

cjh1: @yijiang1 Looking at the [doc](https://docs.scipy.org/doc/scipy-0.16.1/reference/generated/scipy.ndimage.interpolation.rotate.html) for ndimage.interpolation.rotate, it will reshape the output array if necessary.

yijiang1: For scientific datasets we prefer to change the array size after rotation as doing so keeps the same unit.

cjh1: No sure I understand? What is your proposed fix?

yijiang1: Ah never mind I misunderstood the document.

yijiang1:
Rotate transform with angles that are NOT integer multiples of 90 will cause "RuntimeError: output shape not correct"

ercius:
If the different shape is a problem then set reshape=False. You generally
do not need the the parts of the image that will be cut off.

cjh1:
@yijiang1 Would this work OK for GenerateTiltSeries and AutoTiltAxisRotationAlignment, if so that is probably the easiest solution.

yijiang1:
For GenerateTiltSeries: the current implementation already sets reshape=False because I pad the array before rotation. No error here.

For AutoTiltAxisRotationAlignment: I think setting reshape=False is fine for most of electron tomography data sets. 

For Rotate3D: I'm not sure about this one. As an user I prefer to change array size.

cjh1:
@yijiang1 I have added a function to calculate the rotated shape, this should be good to go.

yijiang1:
LGTM

PR #884, opened 2017-02-10
Add support for compiling and running tests using CircleCI

cjh1:
This PR is the first step in improving our CI building and testing our code. It currently compiles and run the few non UI tests we have. The cycle time seems to be just under 8 minutes which it not too bad. A couple of possible improvements for the future:

- [x] Upload result of build/tests to CDash
- [ ] Add some UI tests
- [ ] Make some changes the superbuild so the dependencies in the tomviz-dependencies image could be built using the superbuild. This would potential allow us to automatically rebuild this image if the superbuild changes. Currently this is a manual process.


cjh1:
@cryos Can you approve this branch for me?

PR #893, opened 2017-02-13
Add Ctrl+O shortcut to open files

thewtex:


Hovden:
Great. Adding shortcuts is a good idea. Save State would be another obvious one.

thewtex:
Thanks for the reviews.

@Hovden Yes, agreed. It would be nice to have keyboard shortcuts for many of the actions in the *File* menu in my humble opinion. Not sure what combination to give the others.

@cjh1 Thanks for the tip. Rebased and we are green now.

PR #897, opened 2017-02-14
Make BinaryThreshold cancelable and report its progress

cquammen:
This shows how to make ITK-based data transforms cancelable and report
thir progress. We add a progress observer command that invokes a local
function each time a progress event is emitted. This function relays
the progress to Tomviz and checks to see if the ITK filter should be
canceled, or aborted in ITK parlance. If canceled, the filter throws a
RuntimeError that we catch and ignore, then return from the
transform_scalars function early.

This example also splits the transform into different steps, each of
which is assigned a somewhat arbitrary progress percentage that is
completed when the step is finished.

### DIFF
--- a/tomviz/python/BinaryThreshold.py
+++ b/tomviz/python/BinaryThreshold.py
@@ -1,56 +1,86 @@
-def transform_scalars(dataset, lower_threshold=40.0, upper_threshold=255.0):
-    """This filter computes a binary threshold on the data set and
-    stores the result in a child data set. It does not modify the dataset
-    passed in."""
+import tomviz.operators
 
-    returnValue = None
 
-    # Try imports to make sure we have everything that is needed
-    try:
-        import itk
-        import vtk
-        from tomviz import itkutils
-    except Exception as exc:
-        print("Could not import necessary module(s)")
-        print(exc)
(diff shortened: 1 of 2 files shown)


cquammen:
Some improvements coming...

cquammen:
Okay, ready for review now.

tomviz/python/BinaryThreshold.py
```diff
@@ -1,56 +1,85 @@
...
+
+            itkutils.set_array_from_itk_image(label_map_dataset,
+                                              threshold_filter.GetOutput())
+            self.progress.value = STEP_PCT[4]
+
+            returnValue = {
+                "thresholded_segmentation": label_map_dataset
+            }
+
+        except Exception as exc:
```
thewtex: We still want to add

```
print("Exception encountered while running BinaryThreshold")
```
here?

cquammen: Yes, you are right, catching and re-raising the exception isn't all that useful. I'll print the message.

cquammen:
Thanks! You made the progress/cancel options easy, and the ITK Python wrapping thankfully has that wonderful callback mechanism available.

PR #898, opened 2017-02-15
Add Windows CI with Appveyor

cjh1:


cjh1:
@cryos PTAL this one should be ready to go as well

PR #900, opened 2017-02-15
Add badges for Travis and CircleCI

cjh1:


cjh1:
@cryos I think we might want the little table here explaining want the different badges shown?

cjh1:
@cryos This should be good now

cryos:
So what are the two travis and two appveyor builds here, will they become just one once we merge?

cjh1:
Yes, that is what made me make the change :-)

PR #913, opened 2017-02-20
Update Web app to support data resources inside HTML

jourdain:


cjh1:
@jourdain I restarted the CircleCI build, it failed due to a network issue, but its passing now.

jourdain:
Thanks @cjh1 , feel free to merge it if you feel like it.

web/src/tomviz-viewer.js
```diff
@@ -134,19 +134,18 @@ export function load(container, zipFile, options) {
...
 const container = document.querySelector('.react-content');
 const fileSelector = document.querySelector('.fileSelector');
 
 const bodyStyle = document.querySelector('body').style;
-bodyStyle.background = `url(${background})`;
-bodyStyle.backgroundRepeat = 'no-repeat';
-bodyStyle.backgroundPosition = '50% 50%';
+// bodyStyle.background = `url(${background})`;
+// bodyStyle.backgroundRepeat = 'no-repeat';
+// bodyStyle.backgroundPosition = '50% 50%';
```
cjh1: Should be yank these commented lines?

jourdain: It depend if we want to provide a background image at startup or not. 

And we may change our mind later on. But I guess we can always look at the history to find out how to do it again...

jourdain: We use to provide the drop image, we could provide a loading image instead now.

tomviz/python/tomviz/web.py
```diff
@@ -65,21 +58,47 @@ def web_export(executionPath, destPath, exportType, nbPhi, nbTheta):
...
+                        content = data.read()
+
+                webResources.append(
+                    '<div class="webResource" data-url="%s">%s</div>'
+                    % (relPath, content))
+
+    webResources.append('<script>ready()</script></body>')
+
+    # Create new output file
+    with file(srcHtmlPath) as srcHtml:
```
cjh1: Its preferable to use open(...) here rather than calling the file constructor directly ( according to the python doc ...).

jourdain:
The last commit should make @cjh1 happier... ;-)

cjh1:
@jourdain I'm always happy, just suggestion :smile:

jourdain:
Those are good suggestions and I'm happy that you provide them so I can learn to write better python... I just wanted to mention that the last commit should be tackling your suggestions. ;-)

cjh1:
@cryos If you are happy I think @jourdain can now merge this.

PR #914, opened 2017-02-21
ITK NumPy array conversion requires C order arrays

thewtex:
Newer ITK (4.11) has improved NumPy support that allows for array data
manipulation in place. It also has more stringent checks on the C vs Fortran
order arrays -- ITK assumes C order and will throw an error if this is not the
case.

Extend tomviz.utils.get_array to specify C versus Fortran order. Also, correct
the flattening in tomviz.utils.set_array that was not accounting for C versus
Fortran order.

### DIFF
--- a/tomviz/python/tomviz/utils.py
+++ b/tomviz/python/tomviz/utils.py
@@ -47,37 +47,41 @@ def set_scalars(dataobject, newscalars):
     do.PointData.SetActiveScalars(name)
 
 
-def get_array(dataobject):
+def get_array(dataobject, order='F'):
     scalars_array = get_scalars(dataobject)
-    scalars_array3d = np.reshape(scalars_array, (dataobject.GetDimensions()),
-                                 order='F')
+    if order == 'F':
+        scalars_array3d = np.reshape(scalars_array,
+                                     (dataobject.GetDimensions()),
+                                     order=order)
+    else:
+        scalars_array3d = np.reshape(scalars_array,
+                                     (dataobject.GetDimensions()[::-1]),
+                                     order=order)
     return scalars_array3d
 
 
 def set_array(dataobject, newarray, minextent=None):
-    # Ensure we have Fortran ordered flat array to assign to image data. This
-    # is ideally done without additional copies, but if C order we must copy.
-    if np.isfortran(newarray):
-        arr = newarray.reshape(-1, order='F')
-    else:
-        print('Warning, array does not have Fortran order, making deep copy '
-              'and fixing...')
-        tmp = np.asfortranarray(newarray)
-        arr = tmp.reshape(-1, order='F')
-        print('...done.')
-
     # Set the extent if needed, i.e. if the minextent is not the same as
     # the data object starting index, or if the newarray shape is not the same
     # as the size of the dataobject.
+
+    # Account for indexing and memory layout. VTK uses i,j,k indexing, where i
+    # is the fastest moving index in memory.
+    arr = newarray.ravel(order='A')
+    if newarray.flags.f_contiguous:
+        vtkshape = newarray.shape
+    else:
+        vtkshape = newarray.shape[::-1]
+
     if minextent is None:
         minextent = dataobject.GetExtent()[::2]
     sameindex = list(minextent) == list(dataobject.GetExtent()[::2])
-    sameshape = list(newarray.shape) == list(dataobject.GetDimensions())
+    sameshape = list(vtkshape) == list(dataobject.GetDimensions())
     if not sameindex or not sameshape:
         extent = 6*[0]
         extent[::2] = minextent
         extent[1::2] = \
-            [x + y - 1 for (x, y) in zip(minextent, newarray.shape)]
+            [x + y - 1 for (x, y) in zip(minextent, vtkshape)]
         dataobject.SetExtent(extent)
 
     # Now replace the scalars array with the new array.
--- a/tomviz/python/tomviz/utils.py
+++ b/tomviz/python/tomviz/utils.py
@@ -64,9 +64,9 @@ def set_array(dataobject, newarray, minextent=None, isFortran=True):
     # Set the extent if needed, i.e. if the minextent is not the same as
     # the data object starting index, or if the newarray shape is not the same
     # as the size of the dataobject.
-    # isFortran indicates whether the NumPy array has Fortran-order indexing, i.e.
-    # i,j,k indexing. If isFortran is False, then the NumPy array uses C-order
-    # indexing, i.e. k,j,i indexing.
+    # isFortran indicates whether the NumPy array has Fortran-order indexing,
+    # i.e. i,j,k indexing. If isFortran is False, then the NumPy array uses
+    # C-order indexing, i.e. k,j,i indexing.
 
     if isFortran is False:
         # Flatten according to array.flags
--- a/tomviz/python/tomviz/utils.py
+++ b/tomviz/python/tomviz/utils.py
@@ -60,18 +60,31 @@ def get_array(dataobject, order='F'):
     return scalars_array3d
 
 
-def set_array(dataobject, newarray, minextent=None):
+def set_array(dataobject, newarray, minextent=None, isFortran=True):
     # Set the extent if needed, i.e. if the minextent is not the same as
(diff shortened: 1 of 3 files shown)


cryos:
It is a real shame that vtkImageData uses Fortran ordering, and ITK uses C. There are some formatting issues travis found I think too.

thewtex:
Both `vtkImageData` and `itk::Image` use C ordering for their internal arrays. The order in which indices are specified in NumPy and C arrays are different from the way they are specified in the ITK or VTK API's.

Addressing the style issue in the next push.

thewtex:
More background on the C versus Fortran issue and the indexing order issue is here:

  https://docs.scipy.org/doc/numpy/reference/internals.html

cryos:
The last I checked vtkImageData used Fortran ordering in its internal arrays, hence all the work we have done to keep things in Fortran order despite it not being the default. I am familiar with the issue, Eigen also defaults to C but supports Fortran ordering with a template parameter.

tomviz/python/tomviz/utils.py
```diff
@@ -47,24 +47,23 @@ def set_scalars(dataobject, newscalars):
...
-        arr = newarray.reshape(-1, order='F')
-    else:
-        print('Warning, array does not have Fortran order, making deep copy '
-              'and fixing...')
-        tmp = np.asfortranarray(newarray)
-        arr = tmp.reshape(-1, order='F')
-        print('...done.')
+    # Ensure we have flat array to assign to image data. Account for Fortran vs.
+    # C order.
+    arr = newarray.reshape(-1, order='A')
```
cryos: So reading the documentation, the 'A' ordering doesn't change the underlying array, but instead adapts the indexing into the array? It will be a new view object where possible, so what does this end up doing at the VTK level?

cryos: Also, not entirely convinced we want to unconditionally drop this warning as it lets developers know about potentially expensive copies, which are best avoided.

cryos:
If you think we are misinterpreting the numpy API then it would be good to have a discussion, and ensure what we are doing makes sense. Berk and Utkarsh have worked on a fair bit of the numpy integration, and that is where much of our treatment of the arrays has come from, with the goal if minimal copies being made while preserving the Fortran order for efficient interaction with the vtkImageData API.

thewtex:
Let's discuss this more at the 11 AM Hangout.

tomviz/python/tomviz/utils.py
thewtex: Good point about the view -- will switch to `ravel` instead.

thewtex:
To summarize: indexing and memory layout must be consistent. vtkImageData PointData Scalars can be read in as "Fortran" indexed NumPy array if the shape is the same as `GetDimensions()`, since Fortran indexing in NumPy is indexed as i,j,k, where is the fastest moving index in memory and GetDimenions() returns i,j,k.
Equivalently, vtkImageData PointData Scalars can be read in as a "C" indexed array if the shape is the reverse of `GetDimensions()`, since C indexing in NumPy is indexed as k,j,i, where i is the fastest moving index in memory and `GetDimenions()` returns i,j,k.

So, code that creates multi-dimensional image data views just has to be consistent, which this patch makes possible without unnecessary copies / memory reordering.

The VTK NumPy code works with flat arrays. With this patch, we are doing a `ravel` prior to `numpy_to_vtk`, but `numpy_to_vtk` will do re-ordering of its input unless they are in "C" order -- it calls `ravel` with its default arguments, which is `order = "C"`:

  https://github.com/Kitware/VTK/blob/9b7c9b7c354bc5f1e7d26175af240b6c82207351/Wrapping/Python/vtk/util/numpy_support.py#L165

https://docs.scipy.org/doc/numpy/reference/generated/numpy.ravel.html

This patch was missing the shape setting in `set_array` -- added in the latest revisions.

cjh1:
@thewtex I think you may need to rebase on master to pick up CI changes ...

cryos:
@berkgeveci, @utkarshayachit does this look reasonable to you both? I will see if we can add some tests do do some verification too, but you both know this VTK API/infrastructure far better than I do.

berkgeveci:
Two comments:

* It looks to me that those reshape() calls would do deep copy. I am not sure if that's what you want.
* Is there no support for multi component scalars? Multiple components make things very complicated as VTK uses a hybrid Fortran-C ordering for those (component idx goes first, then i, then j and finally k)

cryos:
Looking at the documentation it would appear that at least newer numpy claims not to do a deep copy on reshape, but I haven't verified that.

We are currently only concerned with single component, and I was most concerned that the changes to set_array look reasonable. If nothing jumps out then I figured we could merge and do some further verification. The numpy_to_vtk part of this is the one I think I understand the least.

cquammen:
I did some testing of the ITK filters with and without this MR and recorded Tomviz's memory usage.

| step | master | this MR |
|-----|--------|---------|
| Startup | 117M | 116M |
| Load small dataset (~60M) | 272M | 270M |
| Binary Threshold | 808M | 669M |
| Delete Binary Threshold | 686M | 674M |
| Binary Threshold (again) | 936M | 805M |
| Delete Binary Threshold | 810M | 803M |
| Binary Threshold (again) | 1060M | 930M |

There is apparently a memory leak, but it looks like with Matt's changes, roughly 2 images worth of memory are saved each time the Binary Threshold filter is applied.

thewtex:
Thanks to all for the investigation and reviews.

I think this branch should still be considered if just for the sake of numpy/scipy Python processing. But, following up on Cory's suggestion the other day that we should look into whether the ITK NumPy bridge could convert Fortran-ordered NumPy arrays -- support for getting an ITK Image from a Fortran-contiguous NumPy array is added here:

  https://github.com/InsightSoftwareConsortium/ITKBridgeNumPy/pull/14

Also, if memory leaks are coming from the ITK NumPy bridge and the tests were using ITK 4.9.0, an update may improve things -- the ITK NumPy bridge was re-written since then to use the [Python Buffer Protocol](https://docs.python.org/3/c-api/buffer.html).

mathturtle:
There is an issue to update our ITK OpenChemistry/tomviz-superbuild#129.  Do we want to do this now?   To what new ITK version?  @cryos @cquammen

cquammen:
I am happy to try a more recent version locally to find out what version to tell Shawn.

thewtex:
ITK 4.11.0 will be a jump forwards. After a few weeks, binary size  / memory should be slightly improved in master.

cryos:
I think at the hackathon we can hack on other local versions of ITK, but I would rather do the ITK move soon after the hackathon as it is the end of Thursday right now and we will all get together early on Monday.

cryos:
I also have a desire to look at moving ITK up, and jumping to Python 3, which could be two good things to investigate, if we have branches that require newer ITK we can hold off merging until they have moved forward. Let's target the end of next week tentatively, with the possibility of me tagging another release before we make the move.

cryos:
@thewtex I had a chat with Berk, and we looked over the code change proposed. It looks like in the general case what is happening on set_array would still transpose the geometry, but for the ITK case it does what is expected.

Why not leave the general purpose set_array as it was, and just make a specific path for ITK where we know we flipped things, and flip them back on the way back in with zero copies? That would avoid the ITK copies, and not risk flipping our geometry in the general case where a numpy result maintained our geometry but used the default C ordering (which is what that warning was designed to guard against).

I guess we could add an optional argument to avoid the copies for ITK, and make that a case in the branching at the start of that function (or refactor it a little to make the general and ITK specific functions call the same code once we have flattened it). I will see if I can write a quick test case to test out our concerns as it would be nice to have some testing coverage on such a critical code path.

thewtex:
@cryos, yes, we could retain the old code path, and I will work on a patch if you think it is worthwhile.

I would expect that any spatially-aware code that uses C ordering with the vtkImageData Dimensions directly for the array shape (without reversing them) will not correctly process the array, so I think the proposed path would be more widely used for non-ITK transforms. Are there example transforms that use the 
```
      tmp = np.asfortranarray(newarray)
      arr = tmp.reshape(-1, order='F')
```
code path?

> I guess we could add an optional argument to avoid the copies for ITK, and make that a case in the branching at the start of that function.

So an `isFortranOrder=True` argument to `set_array`?

PR #920, opened 2017-02-21
Added some code to offer friendly module names

cryos:
This manipulates the pqProxy names where available, and should offer
more easily used names in the spreadsheet view, and from Python.


jourdain:
If that works looks good to me. I didn't know about the pqProxy rename API. ;-)

cryos:
The property name you gave me didn't work, but when asking Utkarsh he said that should do what we want, and it did for the proxies I tested.

jourdain:
Cool, thanks...

Issue #906, opened 2017-02-16
Crash when deleting recon child data


yijiang1:
Binary: 0.9.3-146-gd5e4a10 macOS

yijiang1:
This might be related to #881

cjh1:
@yijiang1 What is resolved by #925 as well?

yijiang1:
It's fixed by #925

Issue #881, opened 2017-02-09
Crash when deleting data transforms


yijiang1:
Binary: 0.9.3-61-g4ff32c6.  macOS

1. Load a data set.
2. Do two data transforms.
3. Right-click the first transform and delete it.
4. Crash

cjh1:
@yijiang1 Is the transformation running when you do the delete? Following the instruction provided above seems to work on linux.

yijiang1:
> Is the transformation running when you do the delete?

No. I delete the transforms after they finishes.

Maybe it's only Mac's problem. Here is part of error message I get:

Process:               tomviz [10483]
Path:                  /Applications/tomviz.app/Contents/MacOS/tomviz
Identifier:            ???
Version:               ??? (0.9.3)
Code Type:             X86-64 (Native)
Parent Process:        ??? [1]
Responsible:           tomviz [10483]
User ID:               501

Date/Time:             2017-02-09 11:24:27.579 -0500
OS Version:            Mac OS X 10.12.3 (16D32)
Report Version:        12
Anonymous UUID:        E0591590-A875-050F-DEE0-519EA2C8DDF4


Time Awake Since Boot: 3200 seconds

System Integrity Protection: enabled

Crashed Thread:        8  Thread (pooled)

Exception Type:        EXC_BAD_ACCESS (SIGSEGV)
Exception Codes:       EXC_I386_GPFLT
Exception Note:        EXC_CORPSE_NOTIFY

Termination Signal:    Segmentation fault: 11
Termination Reason:    Namespace SIGNAL, Code 0xb
Terminating Process:   exc handler [0]

cjh1:
What particular transform are you using?

yijiang1:
Pretty much any transform can cause crash.
Also, it seems resetting tomviz may temporarily fix the problem.
For example,
1. Open tomviz. Click File-->Reset.
2. Load a sample data set.
3. Do Rotate transform.
4. Do another Rotate transform.
5. Delete the first transform. No crash.
6. Do a third Rotate transform.
7. Delete the first transform again, tomviz crashes.
@Hovden Can you reproduce this problem?

Hovden:
Confirmed. Repeating 1-7 causes a crash on OSX v0.9.3-61.

mathturtle:
Step 5 crashes for me... I had noticed this when testing the reconstruction child datasets branch and was looking at it a bit, but I got pulled off on the OME-TIFF thing.

[...]

PR #925, opened 2017-02-22
Make sure deletion only processes once per row

mathturtle:
Since the PipelineView is a multi-column table, we can get multiple
selected indices per row.  This makes sure that each row is only
processed once by the delete code, which avoids deleting objects twice.

Related: #881 #906  -- I am unable to reproduce these crashes with this patch.

@cjh1 @yijiang1


yijiang1:
I can't reproduce these crashes either 👍

PR #930, opened 2017-02-23
Ome tiff reader

mathturtle:
@cryos 

The reader is essentially the vtkTiffReader hacked to support OME-XML in the image description.  I tried not to delete very much.  It doesn't support time yet, but is aware of it being in the file so it only reads the first timestep's data.

Supports simple case of OME-TIFF files and pulls out the physical size of the dataset from the OME metadata.  Right now it defaults to 1um if no size data is specified (um is the default if no units are specified in the OME file format specification).  We should make sure this is the behavior we want.

### DIFF
--- a/tomviz/Behaviors.cxx
+++ b/tomviz/Behaviors.cxx
@@ -64,16 +64,18 @@ namespace tomviz {
 Behaviors::Behaviors(QMainWindow* mainWindow) : QObject(mainWindow)
 {
   Q_ASSERT(mainWindow);
+
+  PV_PLUGIN_IMPORT(tomvizExtensions)
+
   vtkSMReaderFactory::AddReaderToWhitelist("sources", "JPEGSeriesReader");
   vtkSMReaderFactory::AddReaderToWhitelist("sources", "PNGSeriesReader");
   vtkSMReaderFactory::AddReaderToWhitelist("sources", "TIFFSeriesReader");
+  vtkSMReaderFactory::AddReaderToWhitelist("sources", "OMETIFFReader");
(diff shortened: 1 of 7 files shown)


mathturtle:
Some sample files:
https://www.openmicroscopy.org/site/support/ome-model/ome-tiff/data.html

cryos:
Which sample file are you using? I tried single-channel.ome.tif and that causes a crash that seems to originate in tomviz/CentralWidget.cxx:62, trying to get the scalar range. We should try to fail gracefully on what we can't open, I can poke around a little more.

cryos:
z-series.ome.tif worked locally, maybe it is just that first one as I tried a few others too. I did see that the volume render refused to render 

ERROR: In /home/marcus/src/ParaView/VTK/Rendering/Volume/vtkGPUVolumeRayCastMapper.cxx, line 352
vtkOpenGLGPUVolumeRayCastMapper (0x4842e00): scalar of type VTK_CHAR is not supported because this type is platform dependent. Use VTK_SIGNED_CHAR or VTK_UNSIGNED_CHAR instead.

I guess we can easily map this to one type or the other in order to support volume rendering?

mathturtle:
z-series.ome.tiff was one of the ones I tested.  I also used 4D-series.ome.tif to test that only the first timestep is read.  I'll look into why it isn't reading the single-channel.ome.tif file.

mathturtle:
I found the bug where single-channel.ome.tif was failing to read.  We should handle empty images better in the histogram code though.

cryos:
We should probably filter for this case in the load reaction, I am not sure there is any value in reading in an empty image.

Issue #894, opened 2017-02-13
Accessing shift/rotate values?


ercius:
After aligning the tilt axis using the manual tilt axis widget two operators are added to the pipeline: Shift and Rotate. The code of these operators does not indicate how much shift or rotation will be applied. 

In order to track what has been done to a data set, it would be useful to retrieve the shift and rotate values. Suggestions:
1) Write the values into the operator code as the default input value or comment
2) Provide to the user in the Properties window (below the pipeline).

These values are written into the state file, but not easy to extract unless you know where the look:
`<variant type="double" value="-10" />`

ercius:
The proposed change in #932 sounds good to me. The suggestion to allow the user to reopen the tilt align GUI is the ideal situation as also mentioned.

ercius:
Working nicely in 0.9.4

PR #932, opened 2017-02-23
Read in JSON so manual tilt align shows values in UI

mathturtle:
This at least allows the user to see (and edit) the manual tilt align
shift and rotation values after they exit the manual tilt align UI.

@ercius This should help with #894.  I think a real solution would let you get back to the manual tilt axis UI, but this will at least let you view the values it is using for now and was quick to do.  It won't help with old state files, but for new ones it should let you see the values in the properties panel (option 2 from your issue).

@yijiang1

### DIFF
--- a/tomviz/CMakeLists.txt
+++ b/tomviz/CMakeLists.txt
@@ -291,6 +291,7 @@ set(json_files
   Recon_TV_minimization.json
   Recon_SIRT.json
   Recon_WBP.json
+  Shift3D.json
   ShiftTiltSeriesRandomly.json
   )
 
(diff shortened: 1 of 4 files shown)


yijiang1:
If the shift is positive in the manual UI:
screen shot 2017-02-23 at 1 22 16 pm

then the property panel shows all zeros:
screen shot 2017-02-23 at 1 22 31 pm

mathturtle:
Apparently it only works for positive values of the shift argument (which are negative numbers in the Shift Rotation Axis box).  Tracking down what is broken now.

mathturtle:
Yi found a bug in the InterfaceBuilder (fixed in the last commit on this branch).  Unless there is some reason we want to disallow negative values by default?  See [here](http://en.cppreference.com/w/cpp/types/numeric_limits/min) for the difference between `min` and `lowest`.

@cquammen @cjh1

PR #936, opened 2017-02-23
Made it possible for a palette to override custom background

cquammen:
Changed a ParaView-ism where setting a custom background color unlinks
the background property from the global color palette background
property. Now the link is re-established if it is not present whenever
a color palette is applied. This has the effect of applying the color
palette background to the view background.

### DIFF
--- a/tomviz/LoadPaletteReaction.cxx
+++ b/tomviz/LoadPaletteReaction.cxx
@@ -16,13 +16,16 @@
 #include "LoadPaletteReaction.h"
 
 #include <QAction>
+#include <QDebug>
 #include <QMenu>
 
 #include <pqActiveObjects.h>
 #include <pqApplicationCore.h>
 #include <pqApplicationSettingsReaction.h>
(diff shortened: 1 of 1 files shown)


cquammen:
Fixes #817.

@cryos This should be ready to test.

tomviz/LoadPaletteReaction.cxx
```diff
@@ -101,6 +104,25 @@ void LoadPaletteReaction::actionTriggered(QAction* action)
   vtkSMProxy* paletteProxy = pxm->GetProxy("global_properties", "ColorPalette");
 
   if (action->property("PV_XML_NAME").isValid()) {
+    // Setting the color property unlinks the global palette background property
+    // from the view background property. As a result, changes to the palette
+    // do not update the view background. To solve this, we re-link the global
+    // palette background color property to the background.
+    vtkSMGlobalPropertiesProxy* gbPaletteProxy =
```
cryos: Can't this just use auto too?

Issue #933, opened 2017-02-23
Crash when deleting multiple transforms


yijiang1:
0.9.3-212-gd7acad9 on Mac

tomviz crashes when deleting multiple data transforms.
One example:
<img width="381" alt="screen shot 2017-02-23 at 1 58 03 pm" src="https://cloud.githubusercontent.com/assets/13088588/23274382/8bd76764-f9d0-11e6-8be9-c6c44fea02dd.png">

mathturtle:
This should also be fixed since the change is merged.

PR #938, opened 2017-02-23
Make sure we only unpause each pipeline once

mathturtle:
@yijiang1 This should help with #933.

@cjh1 Another 'unpause is getting called twice' bug.


yijiang1:
Yup the bug is fixed.

Issue #935, opened 2017-02-23
Data type is not toggled properly with child reconstruction


yijiang1:
1. Reconstruct a tilt series.
2. If we select the Reconstruction child in the pipeline, we can see that it is marked as volume, which is expected.
<img width="300" alt="screen shot 2017-02-23 at 4 12 58 pm" src="https://cloud.githubusercontent.com/assets/13088588/23279524/9a5ba2ba-f9e3-11e6-93f6-6781e6a40e4e.png"> <img width="300" alt="screen shot 2017-02-23 at 4 13 10 pm" src="https://cloud.githubusercontent.com/assets/13088588/23279650/15d61ac4-f9e4-11e6-86d2-7513e602ebb4.png">

3. However, now if I click previous transforms, the data is still marked as volume:
<img width="300" alt="screen shot 2017-02-23 at 4 13 18 pm" src="https://cloud.githubusercontent.com/assets/13088588/23279691/37d22492-f9e4-11e6-9257-b99f311a41c9.png"> <img width="300" alt="screen shot 2017-02-23 at 4 13 34 pm" src="https://cloud.githubusercontent.com/assets/13088588/23279750/6c3682f0-f9e4-11e6-94e3-fefd501fc2f1.png">

We should mark it back to tilt series.

4. In order to toggle data type, I have to click the parent data in the pipeline:
<img width="300" alt="screen shot 2017-02-23 at 4 13 45 pm" src="https://cloud.githubusercontent.com/assets/13088588/23279811/bb7e3eca-f9e4-11e6-8848-35d56c72d8d3.png">

mathturtle:
This should be merged now.

PR #939, opened 2017-02-23
Update active datasource when module or operator is selected

mathturtle:
If the user clicked off onto another datasource and then clicks on the
active module or any operator, the active datasource should change to
the datasource of the selected object.

@yijiang1 For #935


yijiang1:
LGTM

PR #940, opened 2017-02-24
Add geometry web export

jourdain:


jourdain:
First step to get some geometry output...

cryos:
Cool, one thing I would like to address is removing the 3.5MB HTML page from the repository, it isn't something that is useful to diff, and if it can be generated easily we could do that and store it separately. Another option may actually be turning on git-lfs support, and storing it their along with the PDFs we update occasionally.

jourdain:
The requirements will be:
- Have `npm` + `node` but usually they come together on any build machine
- In CMake trigger the following set of commands

```
$ cd web
$ npm install
$ npm run bundle
```

cryos:
I wanted to rewrite the history of this branch to remove the .html entirely so it doesn't end up in our history, I can look at that though now everything else looks like it works.

PR #941, opened 2017-02-24
Support coloring contours by label map.

tjcorona:
Assign label map as categorical data, preventing spurious
interpolation between labels.

### DIFF
--- a/tomviz/ModuleContour.cxx
+++ b/tomviz/ModuleContour.cxx
@@ -105,7 +105,8 @@ bool ModuleContour::initialize(DataSource* data, vtkSMViewProxy* vtkView)
   controller->PostInitializeProxy(this->ContourFilter);
   controller->RegisterPipelineProxy(this->ContourFilter);
 
-  // Set up a data resampler to add LabelMap values on the contour
+  // Set up a data resampler to add LabelMap values on the contour, and mark
+  // the LabelMap as categorical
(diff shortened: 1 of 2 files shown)


tomviz/ModuleContour.cxx
```diff
@@ -410,6 +430,8 @@ std::string ModuleContour::getStringForProxy(vtkSMProxy* proxy)
     return "Representation";
   } else if (proxy == this->ResampleFilter.Get()) {
     return "Resample";
+  } else if (proxy == this->PointDataToCellDataFilter.Get()) {
+    return "Resample";
```
cquammen: According to the documentation for `Module::getStringForProxy(vtkSMProxy* proxy)`, the string returned should be unique for the various internal proxies, so this should probably return "PointDataToCellData" to match the logic in `ModuleContour::getProxyForString(const std::string& str)` below.

mathturtle: Yes, this is for saving animations.  I need to figure out which proxy is being animated and restore the animation to the right one.  ParaView's mechanism using proxy ids can't work here since we don't/can't save all of the pieces needed to make it work... so this was my workaround.  `ModuleX::getStringForProxy` and `ModuleX::getProxyForString` should be inverses to make this save/load logic work.

tjcorona: Yup, stupid mistake. Fixed!

PR #948, opened 2017-02-27
DataSource now emits its own signals.

tjcorona:


tjcorona:
@mathturtle Here is the MR you requested

PR #949, opened 2017-02-27
Measurement cube length change causes a rerender.

tjcorona:


tjcorona:
@cquammen I believe this will fix the issue you mentioned.

PR #951, opened 2017-02-27
Add ability to hide multiple modules

### LINKED ISSUE #908: Hide multiple visualization modules at once
It would be nice to be able to hide multiple visualization modules by a single click.

One way to do this is adding a "Hide" option to the right-click menu.

cjh1:
Fixes #908


yijiang1:
Tested well 👍 
btw, the new icons look really cool!

cryos:
This doesn't trigger a render for me, and so I have to interact with the scene to see them hide. I also then want a show option.

PR #952, opened 2017-02-27
More new icons

cjh1:


Hovden:
Shouldn't these new icons have more pixels? They all appear pixelated on OSX.

cryos:
There are 32x32 and 64x64 variants, The 2x variant is used when retinal is used, the one question was whether pixmap correctly does the substitution for the spinner.

PR #956, opened 2017-02-27
Add delete confirmation dialog

### LINKED ISSUE #945: Confirmation for deleting dat sets and operators
It is possible to delete data sets and operators in the pipeline by right clicking --> delete or simply pressing the delete key. I think a confirmation box before deleting these item(s) would stop the possibility of deleting things by accident. This could also be important when multiple items are selected (i.e. #873).

cjh1:
Fixes #945

### DIFF
--- a/tomviz/PipelineView.cxx
+++ b/tomviz/PipelineView.cxx
@@ -135,7 +135,7 @@ PipelineView::~PipelineView() = default;
 void PipelineView::keyPressEvent(QKeyEvent* e)
 {
   if (e->key() == Qt::Key_Delete || e->key() == Qt::Key_Backspace) {
-    deleteItems(selectedIndexes());
+    deleteItemsConfirm(selectedIndexes());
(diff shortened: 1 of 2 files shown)


tomviz/PipelineView.cxx
```diff
@@ -357,4 +357,14 @@ void PipelineView::setCurrent(Module* module)
 void PipelineView::setCurrent(Operator*)
 {
 }
+
+void PipelineView::deleteItemsConfirm(const QModelIndexList& idxs)
+{
+  QMessageBox::StandardButton response = QMessageBox::question(
+         this, "Delete pipeline elements?",
+         "You have sure you want to delete the selected pipeline elements?");
```
cryos: Are you sure you want to delete the selected pipeline elements?

cjh1: yes, that is what I meant :smile:

PR #960, opened 2017-02-28
Update itk build

cryos:


cryos:
We should choose a release or SHA for default checkout, need to determine if we can exclude rgba (leads to a configure error in master).

cryos:
Just taking a second look at this now, looking at targeting what works in ITK 4.11.

cryos:
Just building now, I backed off on removing the rgb/rgba types (I think we can live with them, and would rather use a release tag). Any thoughts, opinions? These would feed into updating our superbuild if they look reasonable.

cryos:
Thanks @cquammen, that is all I wanted. This is just the new recommendation I came up with when talking to @thewtex last week. I will merge this for now, but we can certainly revisit if issues arise.

thewtex:
+1 LGTM

PR #963, opened 2017-02-28
Fix web export to work with Python 3

cjh1:


cjh1:
Also requires [VTK change](https://gitlab.kitware.com/vtk/vtk/merge_requests/2536)

cjh1:
OK, the ParaView/VTK changes are in this is ready for merging ...

cryos:
Does the Python 2 path still work, I didn't check that in the end.

PR #971, opened 2017-02-28
BUG: Use ITK Cast filter in the operator function

fbudin69500:
Instead of converting the input pixel type using convert_vtk_to_itk_image,
convert input image type using the ITK Cast filter. Converting
the input image in convert_vtk_to_itk_image leads to a segfault
since the new object containing the image buffer is created in
the scope of convert_vtk_to_itk_image and removed once this
function is done executing.

### DIFF
--- a/tomviz/python/PeronaMalikAnisotropicDiffusion.py
+++ b/tomviz/python/PeronaMalikAnisotropicDiffusion.py
@@ -15,7 +15,7 @@ class PeronaMalikAnisotropicDiffusion(tomviz.operators.CancelableOperator):
 
         # Approximate percentage of work completed after each step in the
         # transform
-        STEP_PCT = [10, 20, 90, 100]
+        STEP_PCT = [10, 20, 30, 90, 100]
 
         try:
             import itk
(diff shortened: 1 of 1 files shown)


fbudin69500:
@cquammen

cryos:
Is this ready to be merged, does it need ITK 4.11 before we do so?

fbudin69500:
I only tested it with the most recent ITK on my machine.

tomviz/python/PeronaMalikAnisotropicDiffusion.py
```diff
@@ -32,8 +32,23 @@ def transform_scalars(self, dataset, conductance=1.0, iterations=100,
             # Get the ITK image. The itk.GradientAnisotropicDiffusionImageFilter
             # is templated over float pixel types only, so explicitly request a
             # float ITK image type.
-            itk_image = itkutils.convert_vtk_to_itk_image(dataset, itkTypes.F)
-            itk_image_type = type(itk_image)
+            itk_image = itkutils.convert_vtk_to_itk_image(dataset)
```
cquammen: This will create a new VTK data object if the input `vtkImageData` has no direct ITK image type counterpart, so this doesn't completely solve the problem you report.

fbudin69500: That is true, it does not solve all the issues.

cquammen:
The underlying problem has been resolved by #1053.

PR #972, opened 2017-02-28
Avoid segfault when ITK image goes out of scope

thewtex:
itk.PyBuffer.GetArrayFromImage returns a NumPy array view on the voxel buffer
data. When an ITK pipeline that is not entirely InPlace in a Tomviz Operator,
the output image of the pipeline also goes out of scope along with the voxel
buffer data it allocated.


thewtex:
CC: @fbudin69500

PR #975, opened 2017-02-28
Add camera linking

cquammen:
Enables linking/unlinking of camera views. Cribs entirely from ParaView's view linking mechanism.

**Adding a link**

To link the cameras of two views, click in one to make it active. Right-click to bring up a context menu and select "Add Camera Link". Select the default settings in the dialog box and click on the other view you want to link.

**Removing a link**

Right-click in a render view and select "Manage Camera Links". In the Link Manager dialog that appears, select the link from the list and click "Remove".


thewtex:
Nice!

PR #977, opened 2017-03-02
Add gradient opacity tf widget

alvarosan:


alvarosan:
Need to run clang format.

cryos:
Meeting now, but soon I am going to check this out!

cquammen:
Wow, you can really see the pores in the small data set! These weren't apparent to me until now.

image

The only problem I noticed is that the gradient opacity widget visibility is not sticky if you click on a different Module and then click back on the Volume module. Other than that it looks nice and works well!

alvarosan:
Ran clang format.  @cquammen , should I cache the visibility? it should be a quick thing to do.  Where can I find that dataset?

cquammen:
To me it seems reasonable to keep the state of the gradient editor visibility - having it disappear is a bit unexpected, to me at least. Ask @mathturtle, @cjh1, or @cryos for the SMALL_PtCu dataset.

cryos:
We just had a chat, and I think it should be global for the datasource, so if you enable it and click between modules it stays visible, if you load a second file then it is hidden by default there. It is looking nice, just going to have a play now.

cryos:
It is very nice, we should try and get Steve to create a nice icon for it, and I might need some explanation on the expected interaction, but I really like how you can pull out some of the edges from the data, and add a new dimension to how we can interact with the transfer function.

alvarosan:
Pushed some fixes for the interaction issues.

alvarosan:
My last patch is in.

cryos:
:+1:

Issue #891, opened 2017-02-13
Manual tilt series alignment not advancing reference image with keyboard a/s


ercius:
The manual tilt series alignment widget does not advance the reference image when a/s keyboard shortcuts are pressed. Clicking on the arrows next to the image number has the desired effect though.
![screen shot 2017-02-13 at 11 42 45 am](https://cloud.githubusercontent.com/assets/15657108/22900125/2112d706-f1e2-11e6-97a2-c2697a4ba398.png)

yijiang1:
Yo need to click the tilt image on the left first. Same for arrow keys.

ercius:
I just double checked that, but no luck. The arrow keys work for shifting but the a/s still does not advance the reference image. I have to click on the Static radio button and then back to Prev. Then it updates.

yijiang1:
> I have to click on the Static radio button and then back to Prev.

Yeah you have to do this for changing the REFERENCE image.

The A/S shortcut adjusts the CURRENT image.

ercius:
This is not the expected behavior.

What you describe is the expected behaviour when the Static button is selected. Then a/s should only advance the current image.

If the Prev. or Next radio button is selected then the Reference AND Current image should advance by +1 for a and -1 for s. 

Furhter, the behavior is different for clicking the Current image up/down arrows compared to the a/s keyboard shortcuts.

yijiang1:
> If the Prev. or Next radio button is selected then the Reference AND Current image should advance by +1 for a and -1 for s.

Yeah this is the feature we need but missing in tomviz currently.

ercius:
>Yeah this is the feature we need but missing in tomviz currently.

Im really confused then. What are Prev. and Next radio buttons for then?

cryos:
I noticed issues in the interaction, and it did not feel natural to me. It would be good to define what we want, and ensure that the interaction is as expected here. I would have expected what Peter expected, and the way it is written the next and prev buttons do not behave as I would have thought - I would have thought static would do what @yijiang1 is stating.

ercius:
Should we define how this should work? @Hovden might have some input here as well. Currently, this GUI is a little difficult to work with during alignment.

yijiang1:
bump

Hovden:
I agree with this: "If the Prev. or Next radio button is selected then the Reference AND Current image should advance by +1 for a and -1 for s."

[...]

PR #982, opened 2017-03-02
Ensure we update the reference slice

cryos:
When using the keyboard shortcuts (one of the most convenient methods),
ensure we update the reference slice respecting what was requested in
the user interface. Fixes issue #891.


cryos:
@cjh1 it looks like appveyor is not happy, but I think this patch is a step in the right direction.

PR #986, opened 2017-03-02
Add scale legend to 3D scene.

tjcorona:
The scale legend can be a cube or a ruler.

### DIFF
--- a/tomviz/CMakeLists.txt
+++ b/tomviz/CMakeLists.txt
@@ -170,6 +170,8 @@ set(SOURCES
   SaveWebReaction.cxx
   ScaleActorBehavior.cxx
   ScaleActorBehavior.h
+  ScaleLegend.h
+  ScaleLegend.cxx
   SetScaleReaction.cxx
   SetScaleReaction.h
   SetTiltAnglesOperator.cxx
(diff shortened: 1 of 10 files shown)


tjcorona:
I would love any/all constructive criticism, thanks!

cjh1:
@tjcorona Looks like there are build errors on Windows, do we need a updated version of ParaView?

https://open.cdash.org/viewBuildError.php?buildid=4791305

tjcorona:
@cjh1 No, I'm not exporting something correctly within tomviz itself. I will look into it.

cryos:
@cjh1 It looks like clang-format is passing when it should fail too, I noticed a number of places in ScaleLegend.cxx where running it locally produces quite a large diff. Any ideas on what might be wrong there as this was working consistently.

cryos:
This looks like it passes now, so maybe a red herring. I am seeing some initialization issues on Linux when I first request the cube it is very large
image

I love how fixed the text is now, that looks great, but wonder if it could be closer to the bottom, and if we might allow the cube to grow a little larger. Switching to ruler I see similar issues where the first render is:
image

Then it looks good, although same thought that it could be lower down on the screen.

tjcorona:
I have made the following changes:

- maximum size of the cube has increased
- ruler and cube sizes are now correct on initial render
- both the ruler and cube have been moved down in the window
- the new files have been clang-formatted

Mind reviewing again?

cryos:
That is great @tjcorona, I'll take a look soon, thanks for addressing all these points so quickly.

Hovden:
I really like both the scale free floating scale cube and the fixed cube / ruler overlay. They both look great and work well. This

I would pick an off white (light gray) to ensure it compatible with a white background.

cryos:
That is a great point, my initial attempts work great for changing the text color but the ruler seems to be ignoring my attempts at setting the color.

PR #988, opened 2017-03-03
Change Measurement Cube to Scale Cube, and add toggle for annotation.

tjcorona:


tjcorona:
@Hovden Would you mind reviewing this merge request?

PR #990, opened 2017-03-03
TRYING TO FIX FORMATING CHECK

cjh1:
The scale legend can be a cube or a ruler.


cryos:
I like that old Python failed, new Python passed :smile_cat:

PR #992, opened 2017-03-06
Make the histogram pretty again...multisampling

cryos:
Udpates to the QVTKOpenGLWidget had disabled multisampling, this meant
that line smoothing was disabled for all charts. Selectively turn this
back on for the histogram so that things don't look so blocky. This
needs some updates that were pushed into VTK a little while ago.


cryos:
This effectively updates the lines from
image
to
image
if a new enough VTK/ParaView is used. If not, then there is no change.

cryos:
Many thanks to @utkarshayachit for fixing this in VTK!

thewtex:
:+1:

PR #993, opened 2017-03-06
Make sure to update label when slice updates

mathturtle:
@yijiang1 When changing the slice via the arrow keys the text telling the current shift was not updating.  Here is a fix that makes it update.

### DIFF
--- a/tomviz/AlignWidget.cxx
+++ b/tomviz/AlignWidget.cxx
@@ -592,7 +592,7 @@ void AlignWidget::changeSlice(int delta)
     i = max;
   }
   this->currentSlice->setValue(i);
-  this->setSlice(i, false);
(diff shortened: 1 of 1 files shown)


yijiang1:
> When changing the slice via the arrow keys

Do you mean changing the current slice here screen shot 2017-03-06 at 12 08 01 pm?

mathturtle:
> Do you mean changing the current slice here screen shot 2017-03-06 at 12 08 01 pm?

No, that one worked, sorry.  When changing the current slice with the A/S keys (not the arrows on the text box above), this label didn't update:
label

cryos:
I missed that in my testing, this looks improved here.

yijiang1:
Ok I understand the bug now.

PR #996, opened 2017-03-06
Refactor contour ui

alvarosan:
Matching contour and volume module's UI.

### DIFF
--- a/tomviz/ModuleContourWidget.cxx
+++ b/tomviz/ModuleContourWidget.cxx
@@ -14,22 +14,21 @@
 
 ******************************************************************************/
 #include "ModuleContourWidget.h"
-#include "ui_ModuleContourWidget.h"
 #include "ui_LightingParametersForm.h"
+#include "ui_ModuleContourWidget.h"
 
 #include "vtkSMProxy.h"
 #include "vtkSMSourceProxy.h"
 
 #include "pqPropertyLinks.h"
-#include "pqWidgetRangeDomain.h"
 #include "pqSignalAdaptors.h"
-
+#include "pqWidgetRangeDomain.h"
 
 namespace tomviz {
 
 ModuleContourWidget::ModuleContourWidget(QWidget* parent_)
   : QWidget(parent_), m_ui(new Ui::ModuleContourWidget),
-  m_uiLighting(new Ui::LightingParametersForm)
+    m_uiLighting(new Ui::LightingParametersForm)
 {
   m_ui->setupUi(this);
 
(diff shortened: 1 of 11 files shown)


cjh1:
@alvarosan The -Wshadow warnings are not a problem if you use a modern compiler (GCC >= 5.4) In general we have taken the approach not to silence them by changing the code as this is something fixed in newer compilers.

alvarosan:
Thanks for the heads up @cjh1 , force of habit from ParaView.  I will stop correcting these.

cryos:
Looks good, we should also be able to expose ambient and diffuse in the UI. I dropped the shadow warning commit from this PR for you. In general it is always best to avoid mixing in other changes to a PR, and make a specific PR to clean up unrelated compiler warnings.

tomviz/ModuleContourWidget.cxx
```diff
@@ -0,0 +1,147 @@
...
+  m_ui->sliValue->setLineEditWidth(leWidth);
+  m_ui->sliOpacity->setLineEditWidth(leWidth);
+  m_uiLighting->sliAmbient->setLineEditWidth(leWidth);
+  m_uiLighting->sliDiffuse->setLineEditWidth(leWidth);
+  m_uiLighting->sliSpecular->setLineEditWidth(leWidth);
+  m_uiLighting->sliSpecularPower->setLineEditWidth(leWidth);
+
+  // TODO Ambient and diffuse are not playing well with coloring by data
+  // (non-solid color).
+  m_uiLighting->sliAmbient->hide();
```
cryos: This may be due to an issue @tjcorona is tracking - contours in general are a little odd right now due to some changes made for categorical data. We are hoping to restore the original default, and then enable the changes for categorical coloring when using a label map.

tomviz/ModuleContourWidget.h
```diff
@@ -0,0 +1,94 @@
...
+   * All proxy properties should use this signal.
+   */
+  void propertyChanged();
+  //@}
+
+private:
+  ModuleContourWidget(const ModuleContourWidget&) = delete;
+  void operator=(const ModuleContourWidget&) = delete;
+
+  std::shared_ptr<Ui::ModuleContourWidget> m_ui;
```
cryos: I can do a quick pass, but these should just be QScopedPointer, they are not shared and I would prefer to be explicit that they are not intended to be.

tomviz/ModuleVolumeWidget.h
```diff
@@ -81,12 +80,9 @@ class ModuleVolumeWidget : public QWidget
   bool usesLighting(const int mode) const;
 
   std::shared_ptr<Ui::ModuleVolumeWidget> m_ui;
+  std::shared_ptr<Ui::LightingParametersForm> m_uiLighting;
```
cryos: As above

alvarosan:
@cryos , I removed the hide() calls and exchanged shared_ptr for QScopedPointer.

alvarosan:
@tjcorona is going to hold on merging his branch until this is merged, so we should go ahead.

PR #997, opened 2017-03-07
Set the text color to a light gray

cryos:
Also attempts to set the ruler color, but this doesn't have any effect
locally.


cryos:
@tjcorona I am not sure if you have any ideas on why this would not work for the ruler, I found a few posts suggesting this should work for the class being derived from.

tjcorona:
The line color is being set here: https://github.com/OpenChemistry/tomviz/blob/master/tomviz/ScaleLegend.cxx#L152
If you change this, I think you should be all set.

cryos:
Thanks @tjcorona, that was the spot I was missing. Looks great now.

PR #998, opened 2017-03-07
Print error if connect fails

cjh1:
This should make is easier to debug when the test fails


tests/cxx/AcquisitionClientTest.cxx
```diff
@@ -99,6 +99,9 @@ private slots:
     QSignalSpy finished(request, &AcquisitionClientRequest::finished);
     finished.wait();
 
+    if (!error.isEmpty()) {
+          qDebug() << error;
```
cryos: Indentation looks off

cjh1:
@cryos Should be fixed up now

PR #999, opened 2017-03-07
Silent warning ( extra semicolon )

cjh1:


cjh1:
@cquammen Thanks for the review, I tend to agree, but we have to keep the compiler happy :smile:

cquammen:
I know, it's such a tyrant! I wonder if clang-format is smart enough to appease this tyrant.

cjh1:
I took a quick look at the doc and couldn't see anything obvious.

cryos:
In this one case I kind of agree, but they are macros, so they are different, and missing all the other places people put extra semicolons makes me sad...

PR #1001, opened 2017-03-07
In ModuleContour, switch between two reps for "color by" fields.

tjcorona:


tjcorona:
This commit does not include lazy initialization of the categorical components of the visualization pipeline. That part is a bit tricky, so I figured I would get this in first and then work on it.

cryos:
Weird bug here @tjcorona, when adding the contour it is visible, but the pipeline marks it as not. Clicking on the closed eye the contour turns white, clicking again it goes back to the color map color. Do you see that locally?

tjcorona:
@cryos I think it is addressed now. Also, the categorical data pipeline is now lazily constructed. There is still work to be done to connect the visualization components to both representations, but perhaps now is a good time for people to chime in on how the visualization components should be laid out?

tjcorona:
Superseded by https://github.com/OpenChemistry/tomviz/pull/1006

PR #1008, opened 2017-03-08
Only reset the zoom upon data loading

cryos:
It can be really irritating to lose your place whenever a module is
added to the scene, move that call from module creation to the data load
reaction.


cryos:
It had been irritating me for a while, and I think it had been the behavior since very early on. I am mixed on only doing it on first load, baby steps I suppose :smiling_imp: Part of my effort to fix up many small niggling issues.

Issue #870, opened 2017-02-07
Can't load datasets to multiple RenderViews


yijiang1:
Binary: 0.9.3-16-g6b94623 on Mac

When there are multiple RenderViews, dataset l load or generate always show up in the first RenderView.

cryos:
Confirmed, you can add stuff to the views but the data load only works in the first view in master.

cryos:
This has been resolved.

yijiang1:
Binary: 0.9.4-35-g807197b

Sample datasets that are created by python transforms still go to the first RenderView window.

cquammen:
Confirmed in master.

cryos:
So #1018 resolves this second issue, and has been merged. Hopefully this resolves the issue fully, as noted we may want to circle back around to this and refactor some code but I think there are other more pressing issues at this moment.

PR #1010, opened 2017-03-09
Prefer the active view if it is a RenderView

cryos:
Do some checking to make sure the active view is a RenderView, if it is
anything else set the active view to the first RenderView available.

### DIFF
--- a/tomviz/LoadDataReaction.cxx
+++ b/tomviz/LoadDataReaction.cxx
@@ -252,7 +252,7 @@ void LoadDataReaction::dataSourceAdded(DataSource* dataSource)
   ActiveObjects::instance().createRenderViewIfNeeded();
   auto view = ActiveObjects::instance().activeView();
 
-  if (QString(view->GetXMLName()) != "RenderView") {
+  if (!view || QString(view->GetXMLName()) != "RenderView") {
     ActiveObjects::instance().setActiveViewToFirstRenderView();
     view = ActiveObjects::instance().activeView();
   }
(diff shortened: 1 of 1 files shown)


cryos:
This fixed #870

cryos:
@cquammen this addresses some changes you made in 9948fd846e4d609bcacfc4504e6c2dee85dcf95b, and preserves using the active view when possible, would you mind taking a look to see if there are any cases you were thinking of there that I might have missed

cquammen:
Works great. I tested the following scenarios:

Load data into default view ✅ 
Create SpreadSheet view. Make it active. Load data. Modules added to first view ✅ 
Create RenderView. Make it active. Load data. Modules added to active view. ✅ 
Make SpreadSheet view active. Load data. Modules added to first view. ✅ 

One scenario where I get a crash:

Create new view panel, but do not create a RenderView or SpreadSheet view. Load data. Crash. ❌ 

Suggested fix: guard for null `view` at LoadDataReaction.cxx:255 and add Modules to first available render view.

cryos:
You sick :dog2:, I never thought of doing anything that perverse! I will add the guard, great catch - thanks @cquammen!

cquammen:
Woof!

cryos:
I don't get that crash, what are you doing, maybe I am missing a step. Line 252 seems to kick in and create a render view if I am sat at the choices, is that not happening for you when trying this?

cryos:
It's OK, I get what you mean now, have a render view, but then sit in another view but don't decide...

cquammen:
Right.

cryos:
Reproduced, fixed, and good to verify it isn't null in any case. Thanks @cquammen

PR #1013, opened 2017-03-10
Remove visualization modules when view is removed

cryos:
It is important not to leave behind orphaned visualization modules,
what is their purpose if they have no render view in which to show
themselves? It is a sad life, and one we should not put them though...


cryos:
This cleans up an ancillary issue pointed out in #869

PR #1018, opened 2017-03-10
Fix adding of sample data source modules to active view

cquammen:
Mimic logic in LoadDataReaction::dataSourceAdded().


cquammen:
Fixed remaining issue identified in #1010.

cquammen:
Ideally the common code would be factored out to a common function, but I was trying to understand where the camera orbit code fit in.

PR #1020, opened 2017-03-10
Fix display name of tool bar in View -> Toolbars menu

cquammen:
A toolbar's display name was "toolBar". Changed this to more descriptive, "Utilities Toolbar" and changed the Qt object name while at it.

Before:

image

After:

image


cquammen:
"Utilities" was the best name I could come up with, but surely there is a better one. The toolbar buttons do fairly different things, though - moving data for one and choosing a color palette for the other.

PR #1021, opened 2017-03-10
CI changes for python 3

cjh1:


cryos:
Is this still relevant @cjh1? We moved to Python 3 as the default, so it would be great to ensure the CI is doing the same. It is marked as WIP right now.

cjh1:
@cryos I will take a look, looks like there is a failure on windows.

cjh1:
@cryos This is now ready to go

Issue #979, opened 2017-03-02
Child datasets and Caching in Pipeline


Hovden:
I just wanted to follow up regarding the child datasets.

The primary intended functionality of child datasets was to cache the pipeline, not create a new branch. Branching is an added benefit but not something we need to make full featured—it is OK if modifications to a parent pipeline rerun entirely.

Using child datasets as a means to cache is important. So there should be a way to "create a child". This could just be a simple Python bit in the "Data Transforms" menu called "Create Child". Having it as a python transform would allow it to serve as example code for new programmers. The small downside is it creates an extra line in the pipeline.

It would be nice if "Create Child" transform was also accessible by right clicking on the dataset in the pipeline. By creating a child, users have effectively cached the transforms leading up to it.

cryos:
So, the point I was attempting to make earlier was that if we have a child dataset then it could be marked as such. If we make the assumption that a child dataset producing operation does not modify its parent then I think we can effectively mark it as frozen, use an :iceland: icon to denote that, and avoid rerunning that.

Alternately we could probably create a special operation object that effectively pins it to that point. By creating a special operation we can know for certain that it does not have any interaction, and so can be frozen without risk. You could then reconstruct that volume rather than the result of the linear pipeline.

Hovden:
I think we have added more functionality to child datasets than originally intended—we need simply to create marks (e.g. children) to prevent re-runing the whole pipeline. A child pipeline should rerun only the things after the beginning of its creation. A child doesn't need to see the rest of the parent's pipeline. Having many child datasets, and allowing more modifications to the parent, etc are extra features—they are also solvable with cloning.

yijiang1:
> it is OK if modifications to a parent pipeline rerun entirely.

I don't think it's OK to always rerun tomographic reconstructions that take a lot time in practice. 

> Having many child datasets, and allowing more modifications to the parent, etc are extra features—they are also solvable with cloning.

In this case, I'd suggest we clone reconstructions by default instead of making them child datasets.

mathturtle:
@Hovden I thought the original point of having a child dataset was to produce a dataset of a different shape (i.e. reconstruction) without destroying the original (parent) data so that the user can continue to work with both.  Thus having multiple child datasets was the point of the feature.  Label maps could just as easily have been extra arrays on the existing image data, so they didn't need to be child datasets.

[...]

Issue #937, opened 2017-02-23
Skip child-dataset-generating transforms when re-executing pipeline


yijiang1:
tomviz always executes the entire pipeline like this:
<img width="345" alt="screen shot 2017-02-23 at 4 41 55 pm" src="https://cloud.githubusercontent.com/assets/13088588/23280765/7b093224-f9e8-11e6-87c8-49a29c50ade0.png">

I think we should skip transforms that generate child data set.

cjh1:
@yijiang1 Not sure I understand why we could skip running the transform, does the Reconstruction step not working on the output of the Auto Tilt Axis Shift Align?

yijiang1:
In practice, tomographic reconstruction takes hours to finish. By setting reconstructions to child datasets, I want to avoid repeating time-consuming transforms. 

Here is a solution:
Say I have a pipeline like this:
Transform 1
Transform 2
Transform 3 -> Generates a child
Transform 4
Transform 5 -> Generates a child

If I edit Transform 4, tomviz only re-executes Transform 1, 2, 4, 5.
If I edit Transform 1 or 2, tomviz re-executes everything from 1 to 5.

cryos:
I think we should add the option to detach a child data set to avoid this, but I don't think this would be a good default behavior in the general case. If we wanted to "freeze" then we should mark it as such.

yijiang1:
> I think we should add the option to detach a child data set to avoid this,

Sounds good to me.

cryos:
We will prioritize this

cryos:
This is now resolved.

PR #1031, opened 2017-03-13
Add a caching operator to the pipeline

cryos:


cryos:
This addresses issues discussed in #937 and #979.

cryos:
@cquammen we were talking about naming/UI, and @cjh1 suggested renaming this operator to "Snapshot" which I think is a better reflection of what it is. I am happy to rename it, but would appreciate opinions before I do. This operator takes a deep copy of the data, and flatly refuses to do anything else after creation. It behaves largely as clone does, but sits nested in the operator pipeline.

cquammen:
I like "Snapshot". I'll play with this PR early this afternoon.

cjh1:
LGTM, I will let @cquammen have a play

cquammen:
Crash in `OperatorFactory::operatorType(Operator* op)` when state file is saved because there is no case corresponding to this new `Operator`.

cryos:
Thanks for the thorough review Cory. The state files are very crashy in general for cloned data, and this is the same kind of behavior. It is important for us to add checking when saving state to verify data nodes have files to back them up, let's see what we can do to address that in future PRs as you suggest.

PR #1036, opened 2017-03-13
Add Volume rendering export for web

jourdain:


jourdain:
@cryos Here is a sample data: https://drive.google.com/open?id=0B0nURBixaI44OFhqYzFBSVNCTmc

jourdain:
@cjh1  I need to know how to upload the [html](https://drive.google.com/open?id=0B0nURBixaI44UTBIc1lZSXNKWVU) file

cjh1:
@jourdain I have pushed a PR that adds some instructions #1037, hope that helps.

jourdain:
Yes thanks, I'm debugging some vtk-js issue and then will update the application.

cryos:
The sample looks great, I guess the bigger thing is figuring out how much to reduce the volume size by to support web rendering, and add some options for that in the export interface. I would also like to add an option (at least for the webGL based stuff) to export as separate files so that we can more easily integrate into web pages on a server.

jourdain:
For server integration, I guess we have 2 options.

1) Since we already export a single HTML file, that same file can be used inside an `iframe` to be embedded in an existing page.

2) If we want a deeper integration we could do the following:
 a) Create a new viewer embedded inside a JavaScript file that can be loaded by the given site.
 b) Add an export option to bundle the data as a single zip file
 c) The integration can then be performed in a web site either in those two ways:

```
<div class='tomviz-viewer' data-url='./tomviz-export.zip' data-viewerSize="200x400" />
```

```
var viewer = tomviz.createViewer('.class-selector', 200, 400);
viewer.loadData('./tomviz-export.zip');
```

Let me know if you want me to build the second one as the first one is already done.

Hovden:
The web export looks great. Especially on a white background.

One issue, the volume explore does not return the opacity map back to the state it was in before.

jourdain:
Thanks for the feedback @Hovden , I'll try to push something to revert back to its original state.

PR #1038, opened 2017-03-14
ModuleContour: switch histogram to reflect current coloring.

tjcorona:


tjcorona:
This MR causes the histogram to reflect the color-by data selected by the user during contouring. It does not rescale the color range (and I'm not sure that it should).

@cryos Would you mind reviewing this merge request?

cryos:
The one oddness I just noticed is clicking away to another module does as I would expect (set the original histogram back), but when returning it does not change to the segmented histogram in my test. Still a big improvement, but I think we want to restore the selected histogram on return.

tjcorona:
@cryos Mind taking another look?

tjcorona:
Clang formatted. Thanks @cjh1 !

cryos:
This looks good @tjcorona, and the interaction is as I would expect it to be. Thanks!

PR #1041, opened 2017-03-14
Update tests to return TIFF rather than RAW format

cjh1:


cjh1:
@cryos I need to make sure this works with python 3

cjh1:
Apparently not :-1:

cjh1:
Apparently I was making up package names :smile_cat:

cryos:
Works much better,
image

Issue #746, opened 2016-11-04
Advanced version information in about dialog


cryos:
Augment the about dialog to show some advanced version information, such as runtime Qt version, Python, ParaView, VTK, ITK, NumPy, etc where available. This should help make is easier to diagnose issues, and offer something easy to copy and paste more complete system state. I envision adding the same information to the state file so that is is all recorded, helping make even local builds offer more reproducibility.

cryos:
I would favor exporting versions to JSON in the about dialog (maybe copy to clipboard or similar). The state file serves a different purpose, to aid in transparency/reproducibility. I had envisioned gathering all of this information in a small helper function that would populate the about dialog, copy a JSON/simple key-value pair to a clipboard, and enable us to insert more verbose information in the state file.

cryos:
@cquammen The VTK version always says 7.1.0 for me, when I don't think it is. If this is an easy tweak I would fix it, otherwise maybe just drop it and keep the ParaView version for now at least - one is implied by the other as we do not manually move VTK back or forward.

cquammen:
You are technically right that VTK is not at version 7.1.0. As it is unimportant, I think I'll just remove it.

cryos:
Sounds great, I am happy with that. Once that is merged we can close this issue out - it satisfies my initial requirements and can be extended as needed in the future.

cryos:
I am going to close this, #1103 removes the VTK version information as discussed.

PR #1042, opened 2017-03-14
Add a slew of version info to the About dialog

cquammen:
Partially addresses request in #746.

Looks like:

image


cquammen:
It would be good to test this against Python 3 if anyone has a Tomviz built against Python 3 laying around.

cjh1:
@cquammen I can test against Python 3

tomviz/AboutDialog.cxx
```diff
@@ -24,23 +24,95 @@
...
+#if PY_MAJOR_VERSION >= 3
+      const char* cdata = PyUnicode_AsUTF8(attr);
+#else
+      const char* cdata = PyBytes_AsString(attr);
+#endif
+      AddItem(tree, itemName, QString::fromStdString(cdata ? cdata : ""));
+    }
+  } else {
+    AddItem(tree, itemName, "unavailable");
+  }
```
cjh1: @cquammen This doesn't compile with Python 3, because Qt defines slots as well as Python 3. I have already created classes in [PythonUtilities] (https://github.com/OpenChemistry/tomviz/blob/master/tomviz/PythonUtilities.h) then encapsulates our python operations, for example you can import a module using this [function](https://github.com/OpenChemistry/tomviz/blob/master/tomviz/PythonUtilities.h#L142). You make have to add some API to do what you want. The ideal is to not leak any python types.

cquammen: Not sure I follow the comment about Qt and Python 3 slots. I should be able to use the Python wrapper. By leaking python types, do you mean exposing them within the core Tomviz code itself?

cjh1: Correct, python type should only be used in PythonUtilities.cxx. I can provide assistance if you run into any issues.

cquammen:
Okay, @cjh1, I think we're on the same page now. How's this?

tomviz/AboutDialog.cxx
```diff
@@ -20,27 +20,96 @@
...
+  AddItem(tree, "Qt Version", QString(QT_VERSION_STR));
+
+  AddPythonAttributeItem(tree, "Python Path", "sys", "prefix");
+  AddPythonAttributeItem(tree, "Python Version", "sys", "version");
+  AddPythonAttributeItem(tree, "NumPy Version", "numpy", "__version__");
+  AddPythonAttributeItem(tree, "NumPy Path", "numpy", "__file__");
+  AddPythonAttributeItem(tree, "SciPy Version", "scipy", "__version__");
+  AddPythonAttributeItem(tree, "SciPy Path", "scipy", "__file__");
+
+#if QT_VERSION >= 0x050000
```
cryos: We don't need this version check, we have a hard requirement of Qt 5

tomviz/AboutDialog.cxx
```diff
@@ -20,27 +20,96 @@
...
+                                          const char* module,
+                                          const char* attribute)
+{
+  Python::initialize();
+
+  bool itemAdded = false;
+  Python py;
+  Python::Module mod = py.import(QString(module));
+  if (mod.isValid()) {
+    Python::Function func = mod.findFunction(QString(attribute));
```
cjh1: Glad you didn't have to add anything to PythonUtilities, looks a little odd the naming thought, looking up an attribute with findFunction(...), I wonder if we need to do some renaming, but I can let it go :snowflake:

cquammen: But... I did add a toString() function to object. I agree the naming is funny - I could add a `findAttribute()` function.

cryos:
This looks great, although I tried with Python 2, my one concern is whether the Python, NumPy, and SciPy paths are useful or might end up being confusing. Maybe moving them all to the bottom would be better as advanced information (keeping the versions at the top).

tomviz/AboutDialog.cxx
cquammen: But that might require a `Python::Attribute` class. Not sure if it is worth it.

cjh1: Sorry, yes, I didn't notice that. I am happy for you to leave as is.

cjh1:
For some request my PYTHONPATH get reported as /usr, strange?

screen

cryos:
Mine does too, but looking at Cory's screenshot it is reporting the prefix path. This is why I think putting the paths at the end makes sense, or possibly removing them if they become a point of confusion. It is great debug information if we have issues. I think ideally the whole thing would be available in a JSON/XML file that could be saved, and displayed like this.

cryos:
Can we show the full Python path, or rename this to Python Prefix perhaps?

cquammen:
Honest question: what would the full Python path? It wouldn't be the python executable as that is Tomviz itself. The shared library file?

I'll change the name to "Python Prefix" for now.

cryos:
For me it would be the path to the Python executable, on Linux you often have different python binaries living in the /usr prefix, commonly a python, python2 and python3. All you are showing is the prefix, so changing the label seems to make sense. Chris also thought it might have been PYTHONPATH environment variable, which I hadn't thought of. Thanks for changing the label, I think certain names evoke certain expectations that may be influenced by the OS you use (macOS path is much more useful than /usr on Linux which is the common prefix for all system installed binaries).

cryos:
I am seeing this link error with my Python 3 build:

[ 96%] Linking CXX executable ../bin/tomviz
libtomviz.a(PythonUtilities.cxx.o): In function `tomviz::Python::Object::toString() const':
/home/marcus/src/tomviz/tomviz/PythonUtilities.cxx:116: undefined reference to `PyString_AsString'
libtomviz.a(PythonUtilities.cxx.o): In function `tomviz::Python::Dict::toString()':
/home/marcus/src/tomviz/tomviz/PythonUtilities.cxx:211: undefined reference to `PyString_AsString'
libtomviz.a(PythonUtilities.cxx.o): In function `tomviz::Python::toPyObject(QString const&)':
/home/marcus/src/tomviz/tomviz/PythonUtilities.cxx:361: undefined reference to `PyUnicodeUCS4_DecodeUTF16'
libtomviz.a(PythonUtilities.cxx.o): In function `tomviz::Python::toPyObject(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)':
/home/marcus/src/tomviz/tomviz/PythonUtilities.cxx:366: undefined reference to `PyString_FromStringAndSize'
libtomviz.a(PythonUtilities.cxx.o): In function `tomviz::Python::toPyObject(tomviz::Variant const&)':
/home/marcus/src/tomviz/tomviz/PythonUtilities.cxx:378: undefined reference to `_Py_ZeroStruct'
libtomviz.a(PythonUtilities.cxx.o): In function `tomviz::Python::toPyObject(long)':
/home/marcus/src/tomviz/tomviz/PythonUtilities.cxx:408: undefined reference to `PyInt_FromLong'
collect2: error: ld returned 1 exit status

cquammen:
@cryos Does master compile with your Python 3? A number of the linker errors you report stem from code not introduced by this PR. Also, Chris is able to compile with Python 3.5.2 (as helpfully reported in his screenshot).

cryos:
Current master does compile, but the tip of this branch does not. This is Python 3.6.0 on Arch Linux, my laptop using Python 3.5.2 on Ubuntu does compile and link with this branch, so perhaps it is a change in 3.6. I found https://docs.python.org/2/c-api/string.html stating it is moving to PyBytes but should be aliased in Python 3.

cquammen:
This branch adds `QString Python::Object::toString() const`. Have you any idea where the other linker failures are coming from? Also, if there is a change in Python 3.6, why isn't there a failure when compiling `PythonUtilities.cxx`. If those functions are gone from 3.6, I would expect an earlier compilation failure, not a linker failure.

cryos:
I am just digging into it too, I am surprised. I have confirmed that /usr/include/python3.6m is being used, which is the correct include path, and /usr/lib64/libpython3.6m.so is being linked to. Switching back to master gives me a successful build, switching back does not. I rebased your branch on master, and still see the same link failure.

cryos:
Commenting out your ITK lines makes everything work/link fine for me. I think something in there is causing issues with my include path, adding the 
    include_directories(${ITK_INCLUDE_DIRS})
back causing link failures. I then went and inspected the ITK build and it had reverted to Python 2 at some point. Changing that things build, so I think this is a local configuration issue, but we should add a CMake check. ITK also adds lots of include directories to the global build path (much as ParaView does), which is a shame especially as it is only used to get the version information.

cryos:
We can do the above in a follow up commit, long and short is everything seems to work OK but verification of consistent Python versions in ParaView (which we use to set our Python version) and ITK (which we do not) is more important if we are adding it to our include path. Bad things would have happened later when running ITK code, error at CMake time is preferable.

cryos:
That would be good, as I feared this branch is causing build failures on the buildbot machines, e.g. https://open.cdash.org/viewBuildError.php?buildid=4809771 for one. I don't think they necessarily made ITK available during Tomviz configure, and simply packaged it after. We will need to get that figured out, or back off on the ITK stuff being required.

cquammen:
Drat, I'll push a branch to clean things up.

cryos:
No worries, I wanted to see if it would work or not. It might not be too hard to ensure Tomviz can see the ITK build at configure time.

PR #1044, opened 2017-03-15
Update PYBIND11_PYTHON_VERSION for Python 3

cjh1:
Merge only when we pull the trigger ...

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,10 +46,6 @@ if(APPLE)
   set(tomviz_data_install_dir "Applications/tomviz.app/Contents/share/tomviz")
 endif()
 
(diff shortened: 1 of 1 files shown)


cjh1:
@cryos We will need to merge this for python 3. Can you think of a better way than hard coding it?

cryos:
No, and we let our users modify it at least. Maybe we will need to improve the docs a little. We don't *need* 3.6, I would prefer to pin it to 3.x

cjh1:
@cryos I will try using find_package(PythonInterp ...) to get the version and uses those to pass to pybind11.

cjh1:
@cryos I think this is better.

cryos:
Yeah, I think we can leave this alone for now, stick with 2.7 on the CI, and revisit the simplest solution once we have the release out. I will update the build instructions.

PR #1045, opened 2017-03-15
More web fix

jourdain:


jourdain:
@cjh1 can you comment on https://github.com/OpenChemistry/tomviz/pull/1045/commits/aa8779f064177e3619ab1e20f2f154ff5641d192

I'm not sure how to properly deal with `include` and `Python::Dict kwargs` as argument of C++ methods.

tomviz/WebExportWidget.h
```diff
@@ -18,6 +18,9 @@
 
 #include <QDialog>
 
+#include "PythonUtilities.h"
+#include "Utilities.h"
+
```
jourdain: This is the section I don't like @cjh1 but I don't know how to write it in a better way.

tomviz/SaveWebReaction.h
```diff
@@ -17,6 +17,7 @@
 #define tomvizSaveWebReaction_h
 
 #include <pqReaction.h>
+#include "PythonUtilities.h"
```
jourdain: This is the section I don't like @cjh1 but I don't know how to write it in a better way.

tomviz/WebExportWidget.h
```diff
@@ -39,10 +42,7 @@ class WebExportWidget : public QDialog
   WebExportWidget(QWidget* parent = nullptr);
   ~WebExportWidget();
 
-  QString getOutputPath();
-  int getExportType();
-  int getNumberOfPhi();
-  int getNumberOfTheta();
+  Python::Dict getKeywordArguments();
```
jourdain: Should there be a & somewhere? @cjh1

cjh1: Python::Dict is a light weight object that can be passed around by value without a problem.

tomviz/WebExportWidget.h
cjh1: Not sure I understand?

jourdain: I though it was bad to include `.h` files in header files.

tomviz/SaveWebReaction.h
jourdain: Same here: I though it was bad to include .h files in header files.

tomviz/WebExportWidget.h
jourdain: Great so no worries on your end.

tomviz/SaveWebReaction.h
cjh1: I think here, its not a big delay as its only included into two cxx files.

tomviz/python/tomviz/web.py
```diff
@@ -264,6 +300,7 @@ def export_volume_exploration_images(destinationPath, camera):
 def export_contour_exploration_images(destinationPath, camera):
     view = simple.GetRenderView()
     contour = get_contour()
+    origianlValues = [v for v in contour.Value]
```
cjh1: typo?

tomviz/python/tomviz/web.py
```diff
@@ -15,7 +15,16 @@
 HTML_WITH_DATA_FILENAME = 'tomviz_data.html'
 
 
-def web_export(executionPath, destPath, exportType, nbPhi, nbTheta):
+def web_export(*args, **kwargs):
```
cjh1: Why the change to **kwargs?

cjh1:
Volume export seems to segfault for me on linux, I do a little debugging.

tomviz/python/tomviz/web.py
jourdain: yep thx

tomviz/python/tomviz/web.py
jourdain: To allow additional option that could be specific to the export type.

cjh1: Makes sense

jourdain:
@cryos this tackle a lot of what we discussed.

Next I'll add more control in the C++/UI (contour values for exploration, down sample size for volume, ...)
Then we should probably create a pure web-viewer for the tomviz data export so we can reference it from a cdn. Moreover, you may not want any UI control with some automatic rotation/animation? Anyway, that may need more talk for it.

cryos:
Just rebuilding my ParaView and it is taking a while... I see there are a few minor clang-format issues, but the code all looks reasonable to me. I just wanted to give it a quick run through before we merge it.

cryos:
With Tomviz using Python 3 on my laptop I see this when trying to export a volume:

  File "/home/marcus/build/tomvizpy3/lib/site-packages/tomviz/web.py", line 69, in web_export
    export_volume(dest, **kwargs)
  File "/home/marcus/build/tomvizpy3/lib/site-packages/tomviz/web.py", line 495, in export_volume
    f.write(buffer(scalars))
NameError: name 'buffer' is not defined

Just recompiling my Python 2 Paraiew... I can try this in the morning where I have both on my desktop.

cjh1:
@jourdain Looks like the segv fault was caused by not have the GIL locked,  add ```Python py;``` to the method should resolve this, if you haven't already fixed it?

jourdain:
I've fixed it by using QMap and only the kwargs in the method calling python.

jourdain:
@cjh1 but I'm not sure how to fix the python3 issue with buffer...

cryos:
When using Python 2, and the "Volume" option I get the following error in the browser console:

tomviz_data.html:53 Uncaught TypeError: Cannot read property 'getListeners' of undefined
    at e.render (tomviz_data.html:53)
    at l._renderValidatedComponentWithoutOwnerOrContext (tomviz_data.html:130)
    at l._renderValidatedComponent (tomviz_data.html:130)
    at l.performInitialMount (tomviz_data.html:130)
    at l.mountComponent (tomviz_data.html:130)
    at Object.mountComponent (tomviz_data.html:26)
    at l.performInitialMount (tomviz_data.html:130)
    at l.mountComponent (tomviz_data.html:130)
    at Object.mountComponent (tomviz_data.html:26)
    at l.performInitialMount (tomviz_data.html:130)

cjh1:
@jourdain I will take a look at the python3 issue

jourdain:
Thanks @cjh1 for Python3.

For @cryos, I'm wondering if I messed up something in the latest html file. I'll double check.

jourdain:
@cryos do you have `TOMVIZ_DOWNLOAD_WEB` `ON`? 
Moreover this is the one I get
```
MD5 (web/tomviz.html) = 65507a0930da37bf82d84d45736262d9
```

tomviz/python/tomviz/web.py
```diff
@@ -492,7 +492,7 @@ def export_volume(destinationPath, **kwargs):
     # Write data field
     fieldDataPath = os.path.join(destinationPath, 'data', 'fieldData')
     with open(fieldDataPath, 'wb') as f:
-        f.write(buffer(scalars))
+        f.write(py2to3.buffer(scalars))
```
jourdain: Interessting, did you do something similar with ParaView/Web/Python/paraview/web/dataset_builder.py ?

cjh1: I don't think so

cryos:
@jourdain sorry that was totally my fault, I had a stale file in my build tree. Removed it, ran CMake and things look good.

tomviz/python/tomviz/web.py
jourdain: Then maybe the other export geometry may not work on python 3

PR #1046, opened 2017-03-15
Ensure ITK and ParaView are built against the same Python at CMake Configure time

cquammen:


CMakeLists.txt
```diff
@@ -74,13 +74,37 @@ if(NOT PARAVIEW_QT_VERSION STREQUAL "5")
...
+  PYTHON_INCLUDE_DIR PYTHON_LIBRARY
+)
+load_cache(${ParaView_DIR}
+  READ_WITH_PREFIX ParaView_
+  PYTHON_INCLUDE_DIR PYTHON_LIBRARY
+)
+
+if (NOT ITK_PYTHON_INCLUDE_DIR STREQUAL ParaView_PYTHON_INCLUDE_DIR)
+  message(FATAL_ERROR
+    "Python include directories are not the same for both ParaView and ITK."
```
cryos: This needs a space at the end, or a newline so that the period doesn't run into the first word of the new sentence. I would also like to see it print out the two include directories for easier debugging, as shown in the failing CI builds.

CMakeLists.txt
```diff
@@ -74,13 +74,37 @@ if(NOT PARAVIEW_QT_VERSION STREQUAL "5")
...
+    "Please ensure ParaView and ITK are being built against the same Python."
+  )
+endif()
+
+if (NOT ITK_PYTHON_LIBRARY STREQUAL ParaView_PYTHON_LIBRARY)
+  message(FATAL_ERROR
+    "Python libraries are not the same for both ParaView and ITK."
```
cryos: Same as above.

thewtex:
LGTM.

cryos:
So I forgot to hit comment last time, writing out my take on this again. We only use ITK in the Python world, so our CI doesn't necessarily have it, and our buildbots actually insert it as a packaging step. This all means that (for now at least) it would be preferable to leave ITK as a non-required dependency.

Ideal case we use ITK_FOUND to do some checking, and use CMake variables for the version number so that our buildbots can inject it into the build. This would make the CMake code a little more complex, but it isn't worth the disruption at this point to do it any other way. This may be something we just push back a little, to Tomviz we see ITK as a Python only dependency, and have for quite some time.

cquammen:
How's this, @cryos ?

cryos:
Good, one final suggestion so that we can make the CIs work without too much modification, can we add an advanced option to skip the checks for CI, people who think they know what they are doing, etc? We can merge, and make some quick modifications to the CI scripts then.

PR #1047, opened 2017-03-16
Fixup itk

cquammen:
This should fix up the nightlies.

Adding ITK version info can be done in another PR.


cquammen:
Oops, the change to `DataTransformMenu.cxx` was meant for another PR and snuck into this one. It was needed, nevertheless.

cryos:
It was fine, I wanted to see if the nightly builds sprang back into life.

PR #1048, opened 2017-03-16
Updated icon for move objects

cryos:


cryos:
Thanks for taking a look Cory.

PR #1049, opened 2017-03-16
Create input UI for more control in Web export

jourdain:


jourdain:
Find an issue with vtk-js... 

Trying to fixing that before merging that topic even if they are totally independent issues.

jourdain:
@cryos @cjh1 Some big volume down sampled with a stride of 5.

https://drive.google.com/open?id=0B0nURBixaI44N3BNQ1JnbVZrSnM

jourdain:
Should be good to merge when green...

cjh1:
If I select "Contour exploration" and don't change anything, then export, I get the following exception:

```python
  File "/home/cjh/work/build/tomviz/lib/site-packages/tomviz/web.py", line 60, in web_export
    export_contour_exploration_images(dest, camera, **kwargs)
  File "/home/cjh/work/build/tomviz/lib/site-packages/tomviz/web.py", line 339, in export_contour_exploration_images
    originalValues = [v for v in contour.Value]
AttributeError: 'NoneType' object has no attribute 'Value'
Failed to execute the script. (/home/cjh/work/source/tomviz/tomviz/SaveWebReaction.cxx:90, bool tomviz::SaveWebReaction::saveWeb(QMap<QString, QVariant>*))
```

jourdain:
That export expect to find a contour filter in the pipeline. The same apply for Volume exploration.

I thought, I gracefully handle to produce nothing when the expected filter was not available. But I may have broke that part. Maybe we want to create them in case they are missing? Any thoughts @cryos @cjh1 ?

cryos:
The ideal for me would be to pop up an info box saying that they will not work with a contour or volume, and then only offer a cancel button or similar to make it clear that nothing would be produced. We can open an issue to track some of the desired polish, I am OK with it being a little less intuitive when we first merge and then tracking this in an issue

jourdain:
Ok let me fix at least the graceful handle and not produce any error if a contour is missing.

jourdain:
Should be pretty nice now, and it should not fail anymore with the error @cjh1 was seeing.

jourdain:
You right, I'm not getting the colors values. But ParaViewWeb lookup table does only RGB color interpolation. So to truly support it, that would require more effort. What I was initially thinking was to add the Tomviz preset into our web application. That way you could get something similar.

cryos:
That sounds reasonable, although what outside of RGB is needed? I think we only use RGB color interpolation at least in the Tomviz stuff we use by default. Having something that looks similar would be nice.

PR #1050, opened 2017-03-16
Add support for copying tomviz.html from source

cjh1:


cjh1:
@cryos I think this provides the behavior you were expecting.

CMakeLists.txt
```diff
@@ -94,15 +94,18 @@ endif()
 # Add web application
 # -----------------------------------------------------------------------------
 option(TOMVIZ_DOWNLOAD_WEB  "Enable downloading web application." OFF)
-set(tomviz_html_path "${CMAKE_BINARY_DIR}/web/tomviz.html")
+set(tomviz_html_source_path "${CMAKE_SOURCE_DIR}/web/tomviz.html")
```
jourdain: I think the actual path is `${CMAKE_BINARY_DIR}/web/www/tomviz.html`

cjh1: Does that matter isn't it the install location that is important? I am happy to change it but would like to understand why?

jourdain: The path I gave you, is the one that is used when building/generating that file. 
I can probably change it to be the root directory of the web, but conceptually, I'd rather keep it separate.

cjh1: ok, will update

Issue #621, opened 2016-09-27
Add Paraview Ruler Tool for Quantitative Analysis of Data


Hovden:
Ruler tool in Paraview would let us measure distances in the sample. It is a nice feature we should consider adding.

<img width="1044" alt="screen shot 2016-09-27 at 1 56 48 pm" src="https://cloud.githubusercontent.com/assets/6197997/18885634/7987c6a2-84ba-11e6-9845-1ceacf768795.png">

cryos:
We were just talking about this, @mathturtle it would be a nice addition and I think you said you were working on this feature in ParaView.

Hovden:
Bumping this. Hoping it would be an easy quantitative tool to add.

ercius:
I will bump your bump and think this would be a great tool to add.

cryos:
@mathturtle did some of the work on adding this to ParaView, we will see if we can get this in soon.

cjh1:
@mathturtle Can this now be closed?

mathturtle:
We have the ruler added as a visualization module, so probably.

Hovden:
It would be nice to note what the units are, in this case length = pixels?

Hovden:
-If the ruler reported the voxel values at the start and end points then we could knock off issue #311

Cosmetic comments: 
-Turning the "eye" off does not make the arrow disappear. Thus users must delete all their measurements for visualization.
-There is a large lag between the ruler annotation and arrow when moving points around. If its easy to make the refresh time faster, that would help.

mathturtle:
> -If the ruler reported the voxel values at the start and end points then we could knock off issue #311

Sure I can make it show the end points in the properties panel (and let you manually set them if you want).

> 
> Cosmetic comments:
> -Turning the "eye" off does not make the arrow disappear. Thus users must delete all their measurements for visualization.

This is a ParaView thing, there is a separate toggle for the arrow.  I'll link the two so that the arrow is forced off when the module is hidden.

> -There is a large lag between the ruler annotation and arrow when moving points around. If its easy to make the refresh time faster, that would help.

I'll look at this, but I don't expect there will be anything easy.


Units: whatever the units on the data are (assuming the data spacing is set appropriately).  I'll see if I can set the text on the ruler to include units... I don't know if it is possible or not.

Hovden:
Ruler "Show Line" box becomes re-checked when highlighted in the pipeline.

cquammen:
[...]

PR #1052, opened 2017-03-17
Ruler show line state

cquammen:
Fixes problem with the Ruler "Show Line" checkbox becoming re-checked when highlighted in the pipeline.

Also fixes a crash when loading more than one state file during a single run of Tomviz.

Addresses remaining issue identified in #621.

### DIFF
--- a/tomviz/Module.h
+++ b/tomviz/Module.h
@@ -127,6 +127,11 @@ public slots:
   /// properties to the panel.
   virtual void addToPanel(QWidget* panel);
 
+  /// This method is called just prior to removing this Module's properties from
+  /// the panel. Subclasses can use it to perform any necessary cleanup such as
+  /// disconnecting some signals and slots.
+  virtual void prepareToRemoveFromPanel(QWidget* panel);
+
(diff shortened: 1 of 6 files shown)


tomviz/Module.h
```diff
@@ -127,6 +127,11 @@ public slots:
   /// properties to the panel.
   virtual void addToPanel(QWidget* panel);
 
+  /// This method is called just prior to removing this Module's properties from
+  /// the panel. Subclasses can use it to perform any necessary cleanup such as
+  /// disconnecting some signals and slots.
+  virtual void prepareToRemoveFromPanel(QWidget* panel);
+
```
cjh1: Any reason this isn't just a signal?

cquammen: I thought about going that route, making this a signal that is emitted prior to the removal of the `Module` items from the panel, but I'd still need a slot to connect it to. No reason not to, and that might be more generally useful, but some other PR...

cjh1: Fair enough

PR #1053, opened 2017-03-20
Avoid segfault due to vtk_image_data loss

thewtex:
This avoids segfault's with newer ITK that is using the Python Buffer-Protocol
NumPy array view in code that uses tomviz.itkutils.convert_vtk_to_itk_image.
A reference needs to be kept to the source vtk_image_data that is responsible
for the pixel buffer memory management.

This is an alternative to #971 that works for additional ITK-based transforms / operators and is to address the case where there is a cast of the input `vtkImageData`.

I have confirmed that it fixes segfault when using ITK master with tomviz and the PeronaMalikAnisotropicDiffusion operator on the test dataset.


cryos:
It looks quite innocuous to me, but @cjh1 has dug into the memory management in VTK with Tomviz pretty deeply so I would appreciate him taking a quick look when he has a chance.

PR #1054, opened 2017-03-20
Decrease anisotropic diffusion time step default to 0.0625

thewtex:
This avoids the warnings:

WARNING: In /home/matt/src/ITK2/Modules/Filtering/AnisotropicSmoothing/include/itkAnisotropicDiffusionImageFilter.hxx, line 82
GradientAnisotropicDiffusionImageFilter (0x7fc762d60600): Anisotropic diffusion unstable time step: 0.13
Stable time step for this image must be smaller than 0.0625

A smaller (more conservative) default that works well for 3D images.

### DIFF
--- a/tomviz/python/Operators.md
+++ b/tomviz/python/Operators.md
@@ -114,6 +114,8 @@ parameter is an array of one or more ints or doubles.
 values specifies the component-wise minimum of a multi-element parameter.
 * `maximum` - Like the `minimum`, but sets the maximum value that a parameter
 may be.
+* `precision` - Optional number of digits past the decimal for `double`
+parameters.
 * `options` - An array of JSON objects, each of which contains a single key/value
 pair where the key is the name of the option and the value is an integer index
 of the options.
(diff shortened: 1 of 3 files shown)


cquammen:
I think there still may some issues when the number passes through the double-spinbox in the UI. We may need to increase the number of digits allowed in that UI widget to fully support this - correct me if I am wrong.

thewtex:
@cquammen Yes, the number is rounded to two digits past the decimal point.

The code that generates the double-spinbox should look at the number of digits in the `default` field of the operator `.json` file?

cquammen:
Actually, you can just add the `precision` attribute to the timestep entry, e.g.,

```
    {
      "name" : "timestep",
      "label" : "Time Step",
      "type" : "double",
      "default" : 0.125,
      "minimum" : 0.0,
      "precision" : 4
    }
```

For this operator, maybe set it to 4.

thewtex:
@cquammen Cool! Adding...

Issue #875, opened 2017-02-08
Histogram doesn't seem to restore correctly from state file


cjh1:
![histo](https://cloud.githubusercontent.com/assets/1031053/22742021/4cdfbf82-ede3-11e6-9d6b-df12f0b47980.png)

cryos:
Resolved.

Issue #828, opened 2017-01-18
vtkColorTransferFunctionItem doesn't resize after being hidden


cjh1:
Steps to recreate:

1. Load SMALL_PtCu_NanoParticle.tif
2. Load MEDIUM_CoP_NanoParticle.tif
3. Remove MEDIUM_CoP_NanoParticle.tif
4. Remove SMALL_PtCu_NanoParticle.tif
5. Load SMALL_PtCu_NanoParticle.tif

The histogram will look like the following screenshot:

![screen](https://cloud.githubusercontent.com/assets/1031053/22073068/4bbe58f6-dd72-11e6-83e4-970c4f0ee0a8.png)

cryos:
Resolved.

Issue #1057, opened 2017-03-20
Color bar in HistogramWidget shrinks when data range changes.


alvarosan:
1. Load MEDIUM_CoP dataset.
2. File Menu -> Reset
3. Load SMALL_PtCu.

![bug_color_bar](https://cloud.githubusercontent.com/assets/767484/24124450/2bcb713e-0d9a-11e7-94af-c3c664950392.png)

cryos:
Resolved.

PR #1058, opened 2017-03-21
Fix a number of bugs with poorly sized color bar

cryos:
The chart containing the color bar wasn't always seeing that its plot
transform was invalid when being made visible again. Calling this
ensures that the bounds and the plot transform will be recalculated.

### DIFF
--- a/tomviz/vtkChartHistogramColorOpacityEditor.cxx
+++ b/tomviz/vtkChartHistogramColorOpacityEditor.cxx
@@ -160,6 +160,7 @@ void vtkChartHistogramColorOpacityEditor::SetColorTransferFunction(
   this->HistogramChart->SetLookupTable(ctf);
   this->ColorTransferFunctionItem->SetColorTransferFunction(ctf);
   this->ColorTransferControlPointsItem->SetColorTransferFunction(ctf);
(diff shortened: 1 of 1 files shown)


cryos:
This should address #828, #875, and #1057. It was a hard won issue a little deep within the charts and how they calculate transforms.

alvarosan:
@cryos , tested this and seems to fix the reported issue (#1057).  It still happens however in some cases, I could reproduce the issue by doing:
1. Load any dataset.
2. File Menu -> Reset.
3. Load MEDIUM dataset.
4. Load SMALL dataset.

screenshot from 2017-03-21 10 54 36

cjh1:
I can confirm is resolves #828

cjh1:
also confirmed is resolves #875,

cryos:
Thanks @alvarosan, I thought I had caught them all. There are some situations when changing the color transfer function that required the same call. It isn't that expensive, and should ideally be triggered by the chart. This should fix up even more cases.

cryos:
Thanks for the confirmation @cjh1

alvarosan:
Nice @cryos , that resolved it.

cryos:
Great, glad we could squash some of these irritating corner cases.

Issue #1060, opened 2017-03-21
Multiple box widgets can be created for Crop operator


cquammen:
* Add a Crop operator to a data set. Apply it.
* Double-click the pipeline entry to edit it. Do not apply the edited settings.
* Double-click the pipeline entry again. A second box widget will appear.

You can get as many box widgets as you want by double-clicking the Crop operator as many times as you wish.

cquammen:
You also get multiple dialogs

Hovden:
I would just emphasize that the crop tool is one of the most used features for our analysis in tomviz.

cquammen:
This actually is true for any Operator that has a custom editor widget, not just the Crop editor.

cquammen:
Fixed in #1062.

PR #1062, opened 2017-03-21
Limit one EditOperatorDialog per Operator

cquammen:
Before this change, it was possible to open many EditOperatorDialogs
by double-clicking as many Operators in the PipelineView that have
custom UI widgets as you wanted, including opening different dialogs
for the same Operator!  With this change, if an EditOperatorDialog
is open for an Operator and the Operator is double-clicked in the
PipelineView, the existing dialog will be raised to the top instead
of a new EditOperatorDialg being created.

Addresses #1060.

### DIFF
--- a/tomviz/Operator.cxx
+++ b/tomviz/Operator.cxx
@@ -31,6 +31,8 @@ Operator::Operator(QObject* parentObject) : QObject(parentObject)
 Operator::~Operator()
 {
   setNumberOfResults(0);
+
+  emit aboutToBeDestroyed(this);
 }
 
 DataSource* Operator::dataSource()
--- a/tomviz/Operator.h
+++ b/tomviz/Operator.h
@@ -211,6 +211,9 @@ signals:
   /// Emitted when a child data source is create by this operator.
   void newChildDataSource(DataSource*);
 
+  /// Emitted just prior to this object's destruction.
+  void aboutToBeDestroyed(Operator* op);
+
 public slots:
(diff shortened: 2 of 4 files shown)


mathturtle:
Are you sure we want to call `reject()` on the existing dialog?  It will close immediately and any changes not applied will be lost.  This will also prevent being able to edit multiple operators at a time (which I am not sure we want or not).  Would it be better to find the dialog for the operator they clicked on and give it focus?

cquammen:
For Operators that don't have custom widgets, we show only one set of properties for the selected Operator in the pipeline view. Doing it this way is therefore consistent with how other operators work. I also think having more than one dialog open at a time is confusing and I posit that it is not a likely use case.

cquammen:
@Hovden What's your opinion on having several operator property dialogs open at once? Useful or no?

cryos:
I think if we are going to call reject on a dialog we should warn the user and give them a chance to change their mind as they could lose changes they were working on. I tend to work on one thing at a time, but hadn't seen an issue allowing more to be open.

I don't think there is any case for opening multiple widgets for the same operation, but the freedom to edit several operations I am not as certain of. If they open multiple widgets and it causes issues it may be reasonable to introduce this new limitation. My thoughts...

cquammen:
@cryos I updated this PR to support up to one EditOperatorDialog per Operator, with support for more than one EditOperatorDialog being open at the same time.

PR #1063, opened 2017-03-21
Remove web

jourdain:


jourdain:
@cryos @cjh1 some cleanup in repo to fully rely on the published [tomvizweb](https://github.com/OpenChemistry/tomvizweb) project and published files.

jourdain:
Only commit https://github.com/OpenChemistry/tomviz/pull/1063/commits/e8c53dd358fcdc51022f14d3ec86995c0a7e5912 is the one that matter as the other one is just removing the `web` directory.

jourdain:
What do you mean by broken hash?

BTW, I was planning to push an updated version of the client... soon

jourdain:
Yes the client is not ready yet for the color handling. Is it what you meant?

cryos:
MD5 and SHA1 are known to have collisions, and as CMake has hashes without any collisions that are still secure it is worth using them (we used SHA512). Sorry for any confusion, not a big deal, just a small effort to upgrade the hashes.

I was just about to look at the client code again now, this certainly works well and it was a very minor comment (Girder uses SHA512 internally and so it seemed like a good choice).

jourdain:
Ah ok. Do you know the command line on OS X to compute a SHA512 of a file?

jourdain:
Although in that specific usage is to validate that the file was not corrupted.

cryos:
A quick look tells me openssl dgst -sha512 /path/to/file should work (Linux has sha512sum). I think it is, but we also want to unambiguously know it is the same file ideally. You could probably alias that command to sha512sum :-)

jourdain:
Once I get the new client built, I'll update the cmake code with sha512. Thx!

PR #1065, opened 2017-03-21
Grid Axis refactor

### LINKED ISSUE #834: Visualization does not update automatically when changing the values in "Unit and Size"
1. Load a dataset. Click View Menu -> Show Axis Grid
screen shot 2017-01-20 at 11 21 22 am

2. Edit X/Y/Z values in the "Unit and Size" panel. The outline and axis grid do not update automatically.
screen shot 2017-01-20 at 11 22 41 am

3. I have to redraw outline and axis grid to see the changes.

cjh1:
Fixes #834

This change removes the "Show Axes Grid" from the view menu and now makes it an option on the Outline. Also fixes the updating when the lengths and units change.

### DIFF
--- a/tomviz/DataPropertiesPanel.cxx
+++ b/tomviz/DataPropertiesPanel.cxx
@@ -275,6 +275,11 @@ void DataPropertiesPanel::updateXLength()
   }
   updateSpacing(0, newLength);
   updateData();
+  DataSource* dsource = m_currentDataSource;
+  if (!dsource) {
+    return;
+  }
+  emit dsource->dataPropertiesChanged();
 }
 
 void DataPropertiesPanel::updateYLength()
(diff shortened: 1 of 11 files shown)


cryos:
This behavior is looking a lot better, and more intuitive. If we can ensure the units work well, and only show the axes I think this will be a much better solution.

cquammen:
The grid appears only on three of the cube sides. Is that intended?

cquammen:
Otherwise interaction is nice and intuitive.

cjh1:
@cquammen What is not like that before ?  :smile_cat:

cquammen:
Quite possibly. Just mentioning it if it wasn't noticed.

cryos:
Yeah it is a new addition, which if people really didn't like we could consider removing. Spoke with @utkarshayachit about some of the ParaView default settings that might make this shift labels to the axis that is not occluded too, so there may be a few more tweaks down the line.

cjh1:
@cryos I have enable front face culling ...

cjh1:
@cryos Try that :crossed_fingers:

cryos:
I think that is the one - it looks good to me, I think @cquammen might appreciate the grid behavior more now too. I am going to merge this, and we can iterate if others see issues.

cquammen:
:1st_place_medal: 👏

PR #1069, opened 2017-03-22
Fix connection of DataSource to volume mapper

cquammen:
We can simply connect the DataSource producer, which doesn't change,
to the volume mapper directly. This requires less work to ensure the
volume is updated correctly.

In addition, this fixes a bug where a Volume module would not update
when the source data was changed.


cryos:
@cjh1 mentioned he had seen some issues with the orthogonal slice and volume in the units/axes branch he was working on.

cryos:
I am guessing they are OK with it, don't see any downsides, so I will merge it and we can always apply further fixes later.

Issue #878, opened 2017-02-08
Integrate tiff reader data spacing into "data properties panel"


Hovden:
Integrate tiff reader data spacing into "data properties panel".

Marcus had a good suggestion: The current popup to specify the data spacing gets annoying. We could integrate these parameters under a new section in the "data properties" panel. Perhaps call it "Voxel spacing (aspect ratios)"

cryos:
Yes, this sounds good, and then we would stop showing the panel for loading TIFF data. It should go above or below the data scale user interface, and would need similar interaction (we need to fix that too so that we rescale the view when the scale or spacing is changed).

Hovden:
Because the Units and Size is really just a voxel scale setting, we still are unable to provide scale in tomviz. It will be great to have a voxel scale and a separate dimensions parameter.

cryos:
You might have to elaborate on that a little, I am not sure I get what you are saying. @tjcorona is doing quite a bit of work on scale, and we talked last week about him adding the voxel spacing parameter underneath the dimensions/units. He had some nice ideas about integration, and view versus data scale I am hoping that we can build in.

Hovden:
Right now the "Units and Size" is doing the same thing as voxel spacing (i.e. aspect ratio). The two should be separate. That is I should be able to specify a voxel aspect ratio that is 1x1x10 and a actual distances that are arbitrary in x,y,z (e.g. 20nm x 12nm x 5nm) and does not change the scale or aspect ratio of the object.

cryos:
I see what you mean, and totally agree. Thanks for the clarification.

cryos:
After looking at this more closely we can improve the experience by suppressing the TIFF reader dialog as it serves no useful purpose. The setting of units using spacing is the only way that we can viably support units in the 3D visualization. After looking at this further I am not sure that we want to support anything different, at least without more thought.

We can support non-square voxels, but they must correspond the units that will be shown without distortion. If you don't have units we could mark the units as pixels, and you can set the spacing to whatever makes sense. Support for squashing or stretching in one dimension is tricky to tie in and we would need to add support in the measurement tools to use an additional property.

Hope that makes sense, maybe we can discuss it more if not.

cryos:
I am going to mark this as resolved, I think we can open new issues if we need to push further into distorted rendering scenarios where units in different directions have non-uniform scaling. There is an easy path to abuse the units and set them to the stretch you wish but you must remember any measurements along those axes will be incorrect.

PR #1070, opened 2017-03-22
Don't open the TIFF configuration dialog

cryos:
We can already set the spacing for a data source, and the defaults
always work as expected.


cryos:
This is related to #878, and once your pull request is merged I think changing the spacing/units/size will all be a little smoother.

cryos:
Thanks for confirming the TIFF series stuff still works.

Issue #1005, opened 2017-03-08
File->Reset doesn't close dialogs


yijiang1:
File->Reset only deletes everything in the pipeline.
Shouldn't it also close all active dialogs and terminate ongoing transforms?
tomviz crashes if I reset the pipeline when a transform is running...

mathturtle:
You seem to have found another way to delete a running operator.  I'll look into it.

mathturtle:
@cryos Would it be a good enough solution to disable the reset action when there is an operator running similar to how we disable the delete actions?

cryos:
Yes, I think that would be reasonable.

cryos:
This should be resolved now.

PR #1072, opened 2017-03-22
Disable File -> Reset when operators are running

mathturtle:
For #1005.


mathturtle:
I wanted to make it tell you *why* it is disabled, but it turns out to be tricky to do that.

Issue #872, opened 2017-02-08
Tilt axis alignment (manual): change 'enter' and 'return' keys default action


ercius:
Im working with the tilt axis alignment (manual) GUI in the newest release. If you change a number then the GUI does not update until you click somewhere. (This is a known QT issue).

However, my first instinct is to press the 'enter' key to update the GUI. By default this exits the GUI and starts the shift/rotate functions. I constantly have to start this process over due to my own reflexes.

Can you change the default behavior such that 'enter' or 'return' do not accept the current conditions of the GUI?

cryos:
That is a great suggestion, I was looking at the interaction in this dialog today and spotted a few issues. We will see if we can improve it, and make this more intuitive.

ercius:
> I was looking at the interaction in this dialog today and spotted a few issues
Regarding this: It is also not possible to zoom or pan in the left or center reconstruction window. Only the right reconstruction window can be zoomed and panned with the middle mouse button. That is worth fixing if possible.

mathturtle:
I'll look at why those reconstruction slices can't be zoomed.  It is probably something simple.  I'm not sure about the 'enter' thing... I looked at it before and it gets tricky.  You have to grab they key events yourself, and I don't remember where you have to grab them to make it work.

PR #1075, opened 2017-03-22
Manual tilt align updates

mathturtle:
This fixes the issues mentioned in #872 with the 'enter' key and with only one view having interaction.  I don't know what else @cryos was planning to fix, but here are the fixes for those.  Pressing 'enter' will now clear the focus from the widget, forcing its editingFinished event to fire. (Pressing enter twice will do that and then close the dialog, so we may want a better indicator that the dialog is working on displaying updates.  This dialog could probably do with some backgrounding of long-running operations.)

### DIFF
--- a/tomviz/RotateAlignWidget.cxx
+++ b/tomviz/RotateAlignWidget.cxx
@@ -58,6 +58,7 @@
 #include <QDoubleSpinBox>
 #include <QGridLayout>
 #include <QHBoxLayout>
+#include <QKeyEvent>
 #include <QLabel>
 #include <QLineEdit>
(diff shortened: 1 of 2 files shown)


cryos:
This works, a similar change is needed in the image alignment dialog. I was planning on reworking the UI in both a little to support resizing elements but more so for the image alignment. This looks reasonable, hopefully the event filters will be reasonably well behaved - the interaction feels improved.

mathturtle:
> This works, a similar change is needed in the image alignment dialog.

Ok, I'll make another PR for that.

PR #1076, opened 2017-03-22
Gradient opacity defaults #1025

### LINKED ISSUE #1025: Improve Gradient Opacity Defaults / UX
The gradient opacity feature is an interesting was to visualize volumetric data. Right now it is fairly opaque to it's inner workings.

-There is no clear explanation of what the gradient opacity is doing. My understanding is that it is the gradient magnitude (not the laplacian, or a weighted Fourier filter right?) with a transparency map that is multiplied with the volumetric visualization. If so, shouldn't the x-axis on the histogram be in units of inverse length? Shouldn't the histogram contain values?

-The default check box for gradient opacity should be off in the volume visualization.
-Right now, when the check box is clicked on or off, nothing changes in the visualization because the gradient opacity defaults to a constant. If possible, we should create a default opacity curve that will work for many datasets.

alvarosan:
Resolves #1025  and other issues with the gradient opacity chart.

### DIFF
--- a/tomviz/GradientOpacityWidget.cxx
+++ b/tomviz/GradientOpacityWidget.cxx
@@ -42,8 +42,8 @@
 namespace tomviz {
 
 GradientOpacityWidget::GradientOpacityWidget(QWidget* parent_)
-  : QWidget(parent_), m_qvtk(new QVTKOpenGLWidget(this))
-  , m_adjustedTable(nullptr)
+  : QWidget(parent_), m_qvtk(new QVTKOpenGLWidget(this)),
+    m_adjustedTable(nullptr)
 {
   // Set up our little chart.
   vtkNew<vtkGenericOpenGLRenderWindow> window_;
@@ -100,7 +100,8 @@ void GradientOpacityWidget::setInputData(vtkTable* table, const char* x_,
   m_histogramView->Render();
 }
 
-void GradientOpacityWidget::prepareAdjustedTable(vtkTable* table, const char* x_)
+void GradientOpacityWidget::prepareAdjustedTable(vtkTable* table,
+                                                 const char* x_)
 {
   m_adjustedTable = vtkSmartPointer<vtkTable>::New();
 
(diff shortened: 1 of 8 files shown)


alvarosan:
Still tracking some issues related to the chart not updating/initializing properly.   One can be reproduced as follows:
1. Load data.
2. Add volume.
3. Show the gradient opacity chart.
4. Drag the edge of the window to resize.

The plot does not resize to the axis as expected,  this does not occur if the gradient opacity chart is visible before adding the volume.

tomviz/GradientOpacityWidget.cxx
```diff
@@ -102,11 +89,58 @@ void GradientOpacityWidget::setLUT(vtkPiecewiseFunction* gradientOpac)
...
+    extentsData[i] = i * step;
+
+    // Add two extra steps in the last table value. This is done to adjust the
+    // range of the chart so that the last point is not occluded.
+    if (i == numBins - 1) {
+      extentsData[i] = (i + 2) * step;
+    }
+  }
+
+  vtkIntArray* pops = vtkIntArray::New();
```
cryos: vtkNew<vtkIntArray> pops instead?

tomviz/GradientOpacityWidget.cxx
```diff
@@ -102,11 +89,58 @@ void GradientOpacityWidget::setLUT(vtkPiecewiseFunction* gradientOpac)
...
+void GradientOpacityWidget::prepareAdjustedTable(vtkTable* table,
+                                                 const char* x_)
+{
+  m_adjustedTable = vtkSmartPointer<vtkTable>::New();
+
+  vtkDataArray* array = vtkDataArray::SafeDownCast(table->GetColumnByName(x_));
+  double* range = array->GetRange();
+  const vtkIdType numBins = array->GetNumberOfTuples();
+
+  vtkFloatArray* extents = vtkFloatArray::New();
```
cryos: vtkNew here too?

tomviz/GradientOpacityWidget.cxx
```diff
@@ -102,11 +89,58 @@ void GradientOpacityWidget::setLUT(vtkPiecewiseFunction* gradientOpac)
...
+  pops->SetName(vtkStdString("image_pops").c_str());
+  pops->SetNumberOfComponents(1);
+  pops->SetNumberOfTuples(numBins);
+  // Initialize with a value > 1.0 so that the y-axis range displays correctly
+  memset(pops->GetVoidPointer(0), 10,
+         sizeof(int) * static_cast<size_t>(numBins));
+
+  m_adjustedTable->AddColumn(extents);
+  m_adjustedTable->AddColumn(pops);
+  extents->Delete();
```
cryos: Could avoid need for delete calls with vtkNew

alvarosan: Sounds good, will do that

Issue #1074, opened 2017-03-22
Color bar does not resize with window


alvarosan:
1. Load dataset.
2. Click on the edge of the window and drag to resize.

![screenshot from 2017-03-22 17 30 57](https://cloud.githubusercontent.com/assets/767484/24221631/79421364-0f25-11e7-884a-515bf4ec2390.png)

cryos:
Confirmed, and it looks like a new regression, I will try to track this down later. I was actually testing a branch and couldn't reproduce, but this issue is present in master...

cryos:
Should be working as expected now.

PR #1078, opened 2017-03-23
Ensure the color bar recalculates its bounds on resize

cryos:
This shouldn't be necessary, but it fixes issues when the length of
the display area changes and the color bar was not updating as
expected.


cryos:
Fixes #1074, this consistently happens due to the recalculate bounds in the set transfer function method, I will try and circle back to the chart classes in VTK to see why these recalculations are not always being triggered as expected for the color bar. This fixes the immediate issue, at least in my testing.

PR #1082, opened 2017-03-23
Update python paths to 3.6

cjh1:


mathturtle:
You may want to check that those are the right paths and that you don't need anything else (like under `Contents/Python/*`).  But this seems OK.

PR #1087, opened 2017-03-23
Set python home macos

cjh1:


cjh1:
@cquammen This should fix the final warning you are seeing

cquammen:
Umm, sure? I don't have Python 3 set up to test, so this doesn't compile. Based on the documentation for PYTHONHOME, this seems reasonable, and I'm happy to try another nightly.

cjh1:
@cquammen Yep, this change will now peg us to python3, @cryos Is that OK? If not I can add some guards. The CI will be a mess until I can fix them.

cryos:
Sorry, I missed this, and now created a conflict as I was seeing if my sledge hammer solution to PYTHONPATH would work, question, happy to try this, or we could use qputenv to add this to the application's environment variables.

Issue #890, opened 2017-02-13
Tilt series sample data missing angles


ercius:
I loaded the sample data  to test some things out with the manual alignment widget. However, the tilt angles are not loaded with this dataset in 0.9.3. This makes it difficult to use as a test data set.

yijiang1:
Are you able to set tilt angles manually?

ercius:
Yes. But I don't know the correct start/stop/step. Its probably 2 degrees per tilt, but the correct values should be loaded automatically.

yijiang1:
I don't think tomviz ever gets tilt angles when loading a data set.
Are tilt angles stored in typical tilt series files?

ercius:
>I don't think tomviz ever gets tilt angles when loading a data set.

This is true.

>Are tilt angles stored in typical tilt series files?

Some data sets store the tilt angles. Sometimes they are stored in a separate file (i.e. filename.rawtlt in FEI's case). 

>For the star nanoparticle tilt series, the tilt angle increment is 2 degree, so the default parameters should work.

I suggest that the test tilt series data set from the pull down menu should load the data and set the tilt angles.  Otherwise the user has to guess the angles which is not ideal for a test dataset.

cryos:
I can look into improving this.

mathturtle:
It shouldn't be too hard to add to the example data if we just want something custom for that dataset (which is what we have to do for now I think).  If someone can find out the correct angles, I can add that.  But I have no idea what the right values should be.

cryos:
This is now resolved in the latest binaries, and works well in my testing.

PR #1088, opened 2017-03-23
Set tilt angles for tilt series sample data

mathturtle:
The load sample data for the tilt series now effectively loads the data and sets the tilt angles on it (with the default values from the set tilt angles dialog accepted).  According to @cryos, these values should be right for this dataset.


mathturtle:
This should address #890.

cryos:
@mathturtle there is a small clang-format issue that needs fixing (placement of *)

mathturtle:
Oops, fixed.

PR #1091, opened 2017-03-24
Try setting PYTHONHOME instead of PYTHONPATH

cryos:


cryos:
Sorry, got distracted, back to :book: writing...

Issue #494, opened 2016-07-14
Negative numbers in tilt series data sets introduces issues during alignment


ercius:
MRC data from FEI microscopes uses a non-standard MRC format and unsigned integers. tomviz imports this data as shorts (signed integers) which does not lose any data (even though this is unphysical) but creates problems later during tilt series alignment. For example, when doing shift alignments, areas shifted from outside of the data are set to 0. This is not the minimum value of a data set with negative values. This causes the reconstruction algorithm to introduce a "block" of density around the edges of the data.
![image](https://cloud.githubusercontent.com/assets/15657108/16824037/4876019c-491d-11e6-9d2e-b6923a4481e4.png)
- Notice large number of 0 values in histogram, and "gray" values at the bottom of the shifted image.

Here are some suggestions for resolving this automatically:
1) Import data as unsigned integers (aka unsigned shorts) to match how the data was written (0 - 65535). I am not sure if tomviz allows for unsigned integers though.

2) import the data as signed integers, cast to float, set minimum to zero

3) Use the array's minimum value as the "cval" option in ndimage.interpolation.shift()

mathturtle:
> uses a non-standard MRC format

Ok, we already have a problem here.  I'm not sure how much we can do to help this in an automated way.  I don't know if changing the reader to handle this will break the reader for some other data file.  If you have any more details on what their non-standard format is, can you link to them on here?  Is there a way to tell if a file is in this format versus a normal MRC file?

> 1) Import data as unsigned integers (aka unsigned shorts) to match how the data was written (0 - 65535). I am not sure if tomviz allows for unsigned integers though.

Like I mentioned above changing the reader to support this will be tricky.

> 2) import the data as signed integers, cast to float, set minimum to zero

Are the negative values bad data?  I think you will be losing valuable data here.  For signed/unsigned short the negative values are the ones that were originally greater than 32768.

> 3) Use the array's minimum value as the "cval" option in ndimage.interpolation.shift()

Hmmm. This might work.  You would need to shift everything less than 0 by +65536 while leaving the zeroes and positive values alone.  You would probably need to convert the data type first too so the values would fit within the range of the new numpy type.  I'm not particularly familiar with scipy, but according to the documentation it might work.  Can you try it?  It would be fairly simple to add a python operator in the data transforms menu to fix this.

ercius:
[...]

PR #1094, opened 2017-03-24
Add "Reinterpret Signed to Unsigned" data transform

cquammen:
Useful for converting data read from non-standard files that incorrectly advertise their integral data types as signed when they should advertise it as unsigned.

Addresses #494.


ercius:
Ill test this out on the binary when it gets incorporated. Thanks for implementing this!

PR #1096, opened 2017-03-24
Various fixes for things found when testing save/load state

mathturtle:
So far:
* toggling outline visibility back and forth would force the axes visibility to on.
* the slice module wasn't updating the panel when it was loaded back in, so the panel had garbage data.

### DIFF
--- a/tomviz/ModuleOutline.h
+++ b/tomviz/ModuleOutline.h
@@ -72,6 +72,7 @@ private:
   vtkWeakPointer<vtkPVRenderView> m_view;
   vtkNew<vtkGridAxes3DActor> m_gridAxes;
   pqPropertyLinks m_links;
+  bool m_AxesVisibility = false;
 };
 }
 
--- a/tomviz/ModuleRuler.cxx
+++ b/tomviz/ModuleRuler.cxx
@@ -150,8 +150,10 @@ bool ModuleRuler::setVisibility(bool val)
 {
   vtkSMPropertyHelper(m_Representation, "Visibility").Set(val ? 1 : 0);
   m_Representation->UpdateVTKObjects();
-  if (m_Widget) {
+  if (!val || m_showLine) {
+    bool oldValue = m_showLine;
     m_Widget->setWidgetVisible(val);
+    m_showLine = oldValue;
   }
   return true;
 }
--- a/tomviz/ModuleScaleCube.h
+++ b/tomviz/ModuleScaleCube.h
@@ -73,6 +73,7 @@ private:
   ModuleScaleCubeWidget* m_controllers = nullptr;
   unsigned long m_observedPositionId;
   unsigned long m_observedSideLengthId;
+  bool m_annotationVisibility = true;
 
 signals:
   /**
--- a/tomviz/ModuleContour.cxx
+++ b/tomviz/ModuleContour.cxx
@@ -237,7 +237,6 @@ void ModuleContour::addToPanel(QWidget* panel)
 
   if (panel->layout()) {
     delete panel->layout();
-    m_controllers = nullptr;
   }
 
   QVBoxLayout* layout = new QVBoxLayout;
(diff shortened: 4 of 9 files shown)


mathturtle:
* toggling the scale cube visibility back and forth would force the annotation visibility on even if it was off before

tomviz/ModuleScaleCube.h
```diff
@@ -73,6 +73,7 @@ protected slots:
   ModuleScaleCubeWidget* m_controllers = nullptr;
   unsigned long m_observedPositionId;
   unsigned long m_observedSideLengthId;
+  bool m_annotationVisibility;
```
cryos: You can use C++11 initialization inline here, i.e.

    bool m_annotationVisibility = false;

tomviz/ModuleOutline.h
```diff
@@ -72,6 +72,7 @@ private slots:
   vtkWeakPointer<vtkPVRenderView> m_view;
   vtkNew<vtkGridAxes3DActor> m_gridAxes;
   pqPropertyLinks m_links;
+  bool m_AxesVisibility;
```
cryos: Same here with initialization with declaration.

cryos:
@mathturtle there is some trailing whitespace clang-format is complaining about, minor suggestion on C++ initialization inline.

tomviz/ModuleScaleCube.h
mathturtle: I'd forgotten I could do that now.  Yay C++11!

mathturtle:
* The ruler arrow visibility was flipping to on when toggling the module visibility.

mathturtle:
* Contour saves properties like Specular Power that were recently exposed (not done testing contour yet)

mathturtle:
... And I found your segfault @cryos.  In the contour if you are coloring by a label map when you save the state, something with the categorical colors UI is segfaulting when you load it.  Looking into how to fix it now.

To be fair, if someone who knows how you are supposed to use the categorical colors and contours could test this more I'd appreciate it.  I got a label map and tried a few things, but I don't know a typical workflow for this part of tomviz so I may have missed something important.  @cquammen @tjcorona

mathturtle:
Updated to fix the segfault.  It still loads the wrong color by array....

mathturtle:
I'm removing the WIP from this.  Contour still loads the wrong color by (always colors by "This Data"), but the segfault is fixed and it would be good to get all these changes in for further testing.

Hovden:
Thank you for thoroughly testing the state files! This is an important part of our mission for reproducible science.

tomviz/ModuleRuler.cxx
```diff
@@ -150,8 +150,10 @@ bool ModuleRuler::setVisibility(bool val)
 {
   vtkSMPropertyHelper(m_Representation, "Visibility").Set(val ? 1 : 0);
   m_Representation->UpdateVTKObjects();
-  if (m_Widget) {
+  if (!val || m_showLine) {
```
cryos: I think this warrants a comment explaining what it is doing, I think I see that you are storing the member variable's value as it will be changed when you toggle the widget visibility. I don't understand why the null check for m_Widget was removed though.

cquammen:
> Updated to fix the segfault. It still loads the wrong color by array....

I suspect that this is caused by the child label map dataset not being created yet when the UI for the Contour is set up, but I don't know that for sure. The state of the Contour UI would need to be updated *after* the child dataset is created when the desired "Color By" dataset is fully available.

tomviz/ModuleRuler.cxx
mathturtle: Yeah, calling setWidgetVisible causes the widget to emit a signal and our signal handler overrides the m_showLine variable.  In this case we don't want the member variable being overridden, so we have to cache it to a local variable and restore it.  I'll add a comment.

The null check... I guess I can put it back.  It didn't matter in any of the cases I tested (and I tried to be thorough since I was removing a null check).

cryos: I am a fan of defensive programming, and would rather check for null even when you are fairly sure it won't happen. A comment would help a lot in this case.

PR #1098, opened 2017-03-24
Warn user if data sources aren't saved

cjh1:
This still needs some working, including:
- [x] Preventing Modules being serialized that are attached to modified data sources.
- [x] Check wording of the warning message
- [ ] Test all the various combinations of snapshoting/cloning/saving.
- [x] Autosave issue

### DIFF
--- a/tomviz/ModuleManager.cxx
+++ b/tomviz/ModuleManager.cxx
@@ -284,7 +284,8 @@ bool ModuleManager::serialize(pugi::xml_node& ns, const QDir& saveDir) const
   // Build a list of unique original data sources. These are the data readers.
   foreach (const QPointer<DataSource>& ds, this->Internals->DataSources) {
     if (ds == nullptr ||
-        uniqueOriginalSources.contains(ds->originalDataSource())) {
+        uniqueOriginalSources.contains(ds->originalDataSource()) ||
+        ds->persistenceState() == DataSource::PersistenceState::Modified) {
       continue;
     }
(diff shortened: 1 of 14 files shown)


tomviz/ModuleManager.cxx
```diff
@@ -232,6 +242,53 @@ bool ModuleManager::serialize(pugi::xml_node& ns, const QDir& saveDir) const
...
+  }
+
+  if (modified > 0) {
+
+    QMessageBox modifiedMessageBox;
+    modifiedMessageBox.setIcon(QMessageBox::Warning);
+    QString text = "The pipeline contains an unsaved data source.";
+    QString infoText = "The unsaved data source is shown in the pipeline in "
+                       "italics with an asterisk. "
+                       "You may continue to save that state and this data "
```
cryos: I think I would tweak the text a little,

    text = "Warning: unsaved data - %1 data source";

then if > 1 append an 's' and it will keep me happy :-)

    infoText = "Unsaved data is marked in the pipeline italic text with an asterisk. You may continue to save the state, and any unsaved data (along with operators/modules) will be skipped.";

cryos:
Added an inline suggestion on tweaks to the message box text, the pipeline is already looking nicer, I think the italics and * help a lot in knowing what is and isn't saved. It warned me in the situations where I thought it should, and didn't when things were all OK.

cryos:
Just noticed it has a weird interaction with the autosave which reuses the state file, that should maybe get added to the list :-)

Issue #1033, opened 2017-03-13
Screen shots are 2x requested size in each dimension on Retina macs


cquammen:
Same behavior is observed in ParaView when high-resolution capable display is enabled.

cryos:
I am going to call this close enough, as the screen shot sizes are passed in as an integer, dividing by 2 leads to a rounding, meaning that any odd number will be rounded down on retinal Macs. Who really wants a 269x100 screenshot? If they do I think a more elegant fix will be required, but this rounds any odd pixel down by one pixel at worst in my testing.

cquammen:
Confirmed, it works reasonably well.

PR #1102, opened 2017-03-25
Attempt to workaround macOS retinal bug

cryos:
Screen shots are being saved at twice the requested resolution due to a
bug in ParaView/VTK on retinal displays at present.


cryos:
This attempts to address #1033

cryos:
Just in the office, and want to test binaries, I will give this a try on Bob's retinal iMac with all the other fixes to verify the binaries are looking good.

PR #1110, opened 2017-03-27
Ensure the plot transform is recalculated

### LINKED ISSUE #1086: Chart in GradientOpacityWidget not updating/initializing properly.
The chart is not updating/initializing properly when the transfer function is set before the chart is shown at least once.  The plot does not resize to the axis as expected, this does not occur if the gradient opacity chart is visible before adding the volume.

The issue can be reproduced as follows:
1. Load data.
2. Add volume representation.
3. Show the gradient opacity chart.
4. Drag the edge of the window to resize.

screenshot from 2017-03-22 19 10 03

cryos:
I see why this misses sometimes, marking an axis as modified is enough
to ensure a recalculation will be triggered correctly, fixes #1086.


cquammen:
FWIW, I didn't see the reported issue in master following the reproduction steps in #1086, but I also don't see the issue after the commit.

cryos:
I could reproduce it here, this is a fairly safe change. I think I will just merge it, it effectively just ensures that on the next paint the plot transform will be recalculated (and in some situations this seems to not work as expected).

PR #1111, opened 2017-03-27
Fixed image size for image export

jourdain:


jourdain:
Need to push new web app before merging...

cryos:
Why does this need the webapp to be merged? We should merge it when ready.

jourdain:
Because It is one less merge request. ;-)
I'm almost there too.. ;-)

cryos:
That is fine, I just couldn't understand from the diff :-) We would like to harmonize the file dialog too, hence my slight push but @mathturtle can base it on this PR tip.

jourdain:
Sounds good to me too. If need be, we can merge it as well. 
Updating the JS  sha, won't be an issue anyway.

jourdain:
Should be ready to merge.  Need to synch-up with @mathturtle  changes...

mathturtle:
I don't think there should be any conflicts...

Issue #801, opened 2016-12-15
Cannot save child table datasets to CSV files


cquammen:
My guess would be that folks will want this capability.

cryos:
This was resolved in #966

cquammen:
Hmm, #966 was closed because it was a duplicate of this issue. I just tested in recent master whether it was possible to save the table data out to a CSV file and it was not. One can save only the Point, Cell, or Field data from the current DataSource image.

cryos:
Thanks @cquammen, I obviously misread as I was working through stale bugs.

cryos:
This is now resolved.

PR #1116, opened 2017-03-27
Enable saving of OperatorResults

cquammen:
If an OperatorResult, such as the table data produced by Label Object
Attributes, is selected when "File -> Save Data" is run, save that
data rather than the active DataSource. This works by keeping track of
the active OperatorResult in ActiveObjects, then when writing data
writing the OperatorResult if it is non-null, rather than the active
DataSource.

### DIFF
--- a/tomviz/ActiveObjects.cxx
+++ b/tomviz/ActiveObjects.cxx
@@ -138,6 +138,19 @@ void ActiveObjects::setActiveOperator(Operator* op)
   emit operatorActivated(op);
 }
 
+void ActiveObjects::setActiveOperatorResult(OperatorResult* result)
+{
+  m_activeOperatorResult = result;
+
+  if (result) {
+    auto op = qobject_cast<Operator*>(result->parent());
+    if (op) {
+      setActiveOperator(op);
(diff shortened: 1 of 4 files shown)


cquammen:
Addresses #801.

cquammen:
To test:

* Load SMALL_PtCu_Particles
* Segmentation -> Binary Threshold
* Segmentation -> Label Object Attributes
* Click on the Component Statistics result in the pipeline view
* File -> Save Data. Save as .csv file.
* CSV file should contain columns, SurfaceArea, Volume, and SurfaceAreaToVolumeRatio

cryos:
Cool! Just giving it a try now.

cryos:
Note for future, it would also be good to improve the context menu in the pipeline to save in this context.

cquammen:
Thanks for the quick review. I thought about providing only sensible file types in the writer, but opted not to for this PR.

cryos:
Agreed, this was reasonable, it would be good to override the types offered, and add the context menu support. Thanks for getting this done, I think it is an important feature and we want to ensure it is as easy as possible to get data into and out of Tomviz.

PR #1117, opened 2017-03-27
Fix axes grid show/hide functionality

cjh1:


cjh1:
Oops already fixed ...

PR #1120, opened 2017-03-27
Allow loading of state contain EMD data sources ( and others )

cjh1:


tomviz/LoadDataReaction.cxx
```diff
@@ -275,6 +282,23 @@ void LoadDataReaction::dataSourceAdded(DataSource* dataSource)
     view = ActiveObjects::instance().activeView();
   }
 
+  ActiveObjects::instance().setMoveObjectsMode(oldMoveObjectsEnabled);
+
+  if (defaultModules) {
+    addDefaultModules(dataSource);
+  } else {
+    auto pqview = tomviz::convert<pqView*>(view);
+    pqview->resetDisplay();
```
cryos: Where did this come from? Have we always reset the view?

cryos:
This looks good in my testing, I wasn't able to trip it up. i don't understand the need for the else block if nothing graphical is added. We have done quite a bit of work to only reset the display when the first module is added to a view, and so it would be nice to avoid that if it is reasonable to.

cryos:
When loading from the recent files menu this now fails to add the default modules as I would expect it to for EMD files (other files seem to work as expected).

tomviz/LoadDataReaction.cxx
cjh1: Yes, dataSourceAdd(...) alway reset the view.

cjh1:
Looks like I need to add a addDefaultModules(...) for the EMD case in RecentFilesMenu.

tomviz/LoadDataReaction.cxx
cryos: I see now, but still don't understand why we would reset the display/render when nothing that can be rendered is being added. I don't see anything but a data source being added, and so reset display/render should not be needed, and would have to be called later when we add a module.

cjh1:
@cryos I have refactored this to simplify things further, I have also remove the unnecessary reset you where concerned about.

mathturtle:
So looking in more detail at the recent files menu... data files from within the state file will be added to the recent files menu with this change.  Do we want to do that?  We only save about the 10 most recent, so loading a state file that loads 4 data files will bump 5 of the recent files off the recent files menu where before it would only add the state file.

cjh1:
I would say probably not.

cjh1:
I can probably add another option to loadData(...)

cjh1:
@mathturtle That should fix it

PR #1121, opened 2017-03-27
Use QFileDialog for web export

mathturtle:
@cryos I don't think this actually conflicted, but best to be safe.  This is based on @jourdain's branch.


jourdain:
I haven't tried it but looks good to me. If need be, you can merge it. I can do another MR once I have the Web component regenerated.

Issue #1122, opened 2017-03-27
Histogram font is tiny on Mac retina display


ElliotPadgett:
Almost all other text and icons are fine, but the text labels on the histogram are very small on my Mac retina display.  Update: Looks fine on a Mac with regular displays.

cquammen:
Confirmed.

ElliotPadgett:
Actually I found that the outline axes labels are impacted by this as well.

Hovden:
I don't mind it. But I was the person that cranked the CRT pixel density to an obscene value.

cquammen:
#1126 was merged, so this should be fixed now.

PR #1126, opened 2017-03-28
Fix too-small fonts on macOS with Retina displays

cquammen:
Set the DPI of the charts so that the font sizes are consistent across different display DPIs.

**On Retina display**

Before:

image

With this MR:

image

**On non-Retina display**

Before:

image

With this MR:

image

### DIFF
--- a/tomviz/GradientOpacityWidget.cxx
+++ b/tomviz/GradientOpacityWidget.cxx
@@ -66,6 +66,13 @@ GradientOpacityWidget::GradientOpacityWidget(QWidget* parent_)
   hLayout->setContentsMargins(0, 0, 35, 0);
 
   setLayout(hLayout);
+
+  // Delay setting of the device pixel ratio until after the QVTKOpenGLWidget
+  // widget has been set up.
+  QTimer::singleShot(0, [=] {
+    int dpi = m_qvtk->physicalDpiX() * m_qvtk->devicePixelRatioF();
+    m_histogramColorOpacityEditor->SetDPI(dpi);
+  });
 }
 
 GradientOpacityWidget::~GradientOpacityWidget() = default;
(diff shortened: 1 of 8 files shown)


cquammen:
@cjh1 @mathturtle @cryos It would be great to see if this causes any issues on Windows or Linux.

cquammen:
Addresses #1122.

mathturtle:
I don't see any issues on non-retina Linux, but I don't have a retina display to test with...

tomviz/GradientOpacityWidget.cxx
```diff
@@ -66,6 +66,13 @@ GradientOpacityWidget::GradientOpacityWidget(QWidget* parent_)
   hLayout->setContentsMargins(0, 0, 35, 0);
 
   setLayout(hLayout);
+
+  // Delay setting of the device pixel ratio until after the QVTKOpenGLWidget
+  // widget has been set up.
+  QTimer::singleShot(0, [=] {
+    int dpi = m_qvtk->physicalDpiX() * m_qvtk->devicePixelRatioF();
```
cryos: Why are you using the float version of device pixel ratio, and then turning it back into an int? This can only be 1 or 2, so I generally use the int version. Everything else makes sense to me.

cryos:
I have a 4k monitor hooked up to my Linux machine, and things look pretty good here
image

cryos:
There are some formatting issues I think pointed out by Travis, this looks like an improvement on my high resolution Linux machine.

tomviz/GradientOpacityWidget.cxx
cquammen: Ah, leftovers from testing different things. Thanks for catching.

cryos:
For some reason the tooltip text comes out super big, maybe there are some remaining issues.
image

cquammen:
Yuck, clang-format insists on an indentation after a macro with no subsequent semicolon is invoked. @cryos do you prefer odd indentation or one of those poor, unfairly-maligned semicolons following the macro invocation? I know which I prefer :-)

cryos:
I will just merge, and tell the bot to be quiet :-)

PR #1127, opened 2017-03-28
Allow saved cloned data source to be included in state file

cjh1:


cryos:
This worked well locally, saving as an EMD misses something as it didn't update the filename or mark it as saved (I thought I might have been pushing it). Saving to TIFF let me save/restore with a cloned transformed data set successfully.

tomviz/SaveDataReaction.cxx
```diff
@@ -185,6 +186,12 @@ bool SaveDataReaction::saveData(const QString& filename)
   }
   writer->UpdateVTKObjects();
   writer->UpdatePipeline();
+
+  if (!ModuleManager::instance().isChild(source)) {
```
cryos: My EMD was special, I exited based upon success or failure.

cjh1:
@cryos OK, EMD should work now.

PR #1128, opened 2017-03-28
Tweaks to the file menu

cryos:
image


cryos:
Thanks for taking a look, I will merge and we can of course make further tweaks in the future.

PR #1147, opened 2017-03-30
Prefer finding Python 3 when building ITK

thewtex:
Python_ADDITIONAL_VERSIONS is used by PythonInterp.cmake to specify
preferred versions of Python. Prefer Python 3 on systems like Linux
where /usr/bin/python may be Python 2.


thewtex:
This will give a gentle nudge, as opposed to forcing. :peace_symbol: Setting PYTHON_EXECUTABLE, PYTHON_LIBRARY, PYTHON_INCLUDE_DIR variables is a good idea to be explicit in specifying desired variables.

PR #1148, opened 2017-03-30
Add acquisition dev mode

cjh1:


acquisition/devops/README.md
```diff
@@ -0,0 +1,34 @@
...
+Where ```<log>``` is one of:
+
+- ```stderr``` - Redirected stderr ( server must be started with ```-r``` options ). Bottle writes it output here.
+- ```stdout``` - Redirected stdout ( server must be started with ```-r``` options ).
+- ```debug``` - The servers debug log.
+
+
+# Deploying a source adapter
+
+If there server is started in dev model source adapter can be deployed remotely
```
cryos: I think you mean "If the server is started in dev mode a source adapter can be deployed"

cryos:
Merging, we can fix the docs up later.

PR #1149, opened 2017-04-02
Ensure out of range angle produces blank image

cjh1:


cryos:
I need to look at my local set up, master and this branch both behave in the same way (and master was working yesterdya) - no response. The changes all look reasonable like I said, but I couldn't test.

PR #1155, opened 2017-04-04
Partial integration 2D Transfer Functions

alvarosan:
This changeset sets the base for further integration of 2D transfer functions into tomviz.  The feature is currently enabled (as a proof of concept), but it should be hidden/disabled before merging given that it is incomplete.

It requires a custom [VTK branch](https://gitlab.kitware.com/alvaro.sanchez/vtk/commits/2d-transfer-func) to run.

window_tf2d

### DIFF
--- a/tomviz/CMakeLists.txt
+++ b/tomviz/CMakeLists.txt
@@ -80,6 +80,8 @@ set(SOURCES
   GradientOpacityWidget.cxx
   HistogramWidget.h
   HistogramWidget.cxx
+  Histogram2DWidget.h
+  Histogram2DWidget.cxx
   InterfaceBuilder.h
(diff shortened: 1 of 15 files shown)


alvarosan:
@cryos,  I left it in a working state in case you want to try it out (the rendered image updates as the box item is manipulated), there are a few bugs, so make sure not to hit the edges of the chart.   Before merging, make sure the involved widgets are hidden and the connection to compute the 2D histogram is commented.

alvarosan:
@cryos @cquammen , could you reopen this merge request?  I think you could begin reviewing the current workflow so that we can start to get it ready for the next release.  There are still a few open issues on my list but I would like to start getting your feedback.

PR #1156, opened 2017-04-04
Add some extra calls for disconnect, acquisition params

cryos:


cryos:
I am not sure if the key values are correct, but this is a snapshot of what I thought we had this morning.

cjh1:
Disconnect doesn't seem to be wired up?

cryos:
No, I had forgotten, added now and seems to work as expected.

PR #1157, opened 2017-04-05
Try scaling our effective DPI a little

cryos:
This is intended to reduce the font size a little on high DPI screens
where it seems to grow too quickly. Admittedly a little hackish, and
this should be followed up with some work in VTK to verify that the
DPI is scaling the font pixel sizes appropriately.


cryos:
This gives something that looks more reasonable on Linux,
image

PR #1166, opened 2017-05-12
Improve building documentation

jabl:


jabl:
These were issues I ran into when trying to follow BUILDING.md to build tomviz on Ubuntu 16.04.

cryos:
What was the issue with needing system protobuf? This is meant to be the simplest way to get up and running on all platforms, so I would like to avoid system libraries for the general case. I use Ubuntu 16.04 and haven't had issues with protobuf. I would like to update the instructions to consistently use Python 3.

jabl:
The build itself went through fine, but when I tried to run tomviz, it aborted with an error complaining about version conflicts.  Sorry I didn't copy the exact error message..

Then I tried to remove the vendored protobuf libs from the paraview-build/lib directory, but then the result was another similar error message  (not that surprising...). But this one I did save:

<pre><code>
[libprotobuf FATAL google/protobuf/stubs/common.cc:72] This program was compiled against version 2.3.0 of the Protocol Buffer runtime library, which is not compatible with the installed version (2.6.1).  Contact the program author for an update.  If you compiled the program yourself, make sure that your headers are from the same version of Protocol Buffers as your link-time library.  (Version verification failed in "ParaViewCore/ServerImplementation/Core/vtkPVMessage.pb.cc".)
terminate called after throwing an instance of 'google::protobuf::FatalException'
  what():  This program was compiled against version 2.3.0 of the Protocol Buffer runtime library, which is not compatible with the installed version (2.6.1).  Contact the program author for an update.  If you compiled the program yourself, make sure that your headers are from the same version of Protocol Buffers as your link-time library.  (Version verification failed in "ParaViewCore/ServerImplementation/Core/vtkPVMessage.pb.cc".)
zsh: abort      bin/tomviz
</code></pre>

ldd bin/tomviz showed that tomviz was using the system protobuf. Rebuilding paraview with the system protobuf allowed it to work as then paraview and tomviz were both using the same version.

But now that I try to recreate my original error I'm not able to. Huh? Well, let me revert that commit and squash it..

jabl:
Now I'm able to recreate my failure using the vendored protobuf. The error message is:

<pre><code>
libprotobuf FATAL /l/jblomqvi/paraview/ThirdParty/protobuf/vtkprotobuf/src/google/protobuf/stubs/common.cc:62] This program requires version 2.6.0 of the Protocol Buffer runtime library, but the installed version is 2.3.0.  Please update your library.  If you compiled the program yourself, make sure that your headers are from the same version of Protocol Buffers as your link-time library.  (Version verification failed in "/build/mir-pkdHET/mir-0.21.0+16.04.20160330/obj-x86_64-linux-gnu/src/protobuf/mir_protobuf.pb.cc".)
Aborted
</code></pre>

I'm not AFAIK(?) using Mir, but there is a package libmirprotobuf3 installed, which links against the system libprotobuf-lite (which is version 2.6).. Don't know if that is relevant.

jabl:
This seems to be the same error message as reported here: https://cmake.org/pipermail/paraview/2017-March/039751.html

Unfortunately it seems that discussion died out without any resolution.

PR #1191, opened 2017-06-27
First iteration 2D Transfer Functions

alvarosan:
First iteration of 2D transfer functions in tomviz.

Supersedes #1155 

tomviz_2dtf

### DIFF
--- a/tomviz/Module.h
+++ b/tomviz/Module.h
@@ -43,6 +43,15 @@ public:
   Module(QObject* parent = nullptr);
   ~Module() override;
 
+  /// Transfer function mode (1D or 2D). This enum needs to be synchronized with
+  /// the order of the tabs in ui->tabWidget1D2DTransfer.
+  enum class TransferMode : unsigned char {
+    TF_1D = 0,
+    TF_2D
+  };
+  void setTransferMode(const int mode);
+  int getTransferMode() const;
+
   /// Returns a  label for this module.
   virtual QString label() const = 0;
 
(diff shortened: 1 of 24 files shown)


tomviz/Module.h
```diff
@@ -42,6 +43,16 @@ class Module : public QObject
...
   ~Module() override;
 
+  /// Transfer function mode (1D or 2D). This enum needs to be synchronized with
+  /// the order of the tabs in ui->tabWidget1D2DTransfer.
+  enum TransferMode
+  {
+    TF_1D = 0,
+    TF_2D
+  };
+  void setTransferMode(const int mode);
```
alvarosan: Good catch

cjh1:
@alvarosan CI should have the correct version of ParaView now.

alvarosan:
Thanks @cjh1 .   I pushed my latest fixes.

alvarosan:
@cryos , should we merge this?

mathturtle:
@alvarosan If I open tomviz, load in a dataset and exit tomviz it leaks a vtkImageData (According to vtkDebugLeaks).  `git bisect` points at this branch, but I don't want to recompile ParaView so I can't bisect further...

cryos:
I think it would be best to report an issue and track it there rather than here (reference this PR). It is new functionality, with new API that was added to VTK. It would be good to see if @alvarosan can reproduce the leak and track it down, thanks for catching this.

alvarosan:
Good catch @mathturtle , I created an issue for it.

PR #1206, opened 2017-07-21
Update to new name of ParaView property

mathturtle:
LockScalarRange property is replaced by AutomaticRescaleRangeMode in
ParaView 5.5, tomviz needs to update to use the new name.


mathturtle:
There's still a vtkDebugLeaks issue I need to track down too.

PR #1210, opened 2017-07-26
Use vtkSmartPointer to manage members

cryos:
This simplifies memory management logic for members that are assigned
from the outside.


cryos:
I think this should solve the issue from #1209 while removing manual memory management, it would be good to go through and check we aren't doing this in other places. I don't have a build with debug leaks to confirm, but am pretty certain it will still be leak free.

cryos:
Thanks @mathturtle, debug mode has compile errors, I will get that fixed up.

cryos:
@mathturtle this compiles in debug mode locally.

alvarosan:
+1 for using smart pointers, I should have done that from the beginning.  I will take another pass on these new classes to make sure they are not managing memory manually.

PR #1212, opened 2017-07-26
Tiff double

cryos:


cjh1:
Not sure quite how to test this. Saving out tiff works. Is there an easy way to validate the the tiff file contains floats?

cryos:
Thanks, it is very quiet - easiest way is to simply open the file you saved, and it should show as a float/have the data expected. I think this is the right way of handling it, I am in two minds about popping up a warning that TIFF requires the data be converted to 32 bit floating point.

PR #1214, opened 2017-07-27
Remove modules from sys.module before importing them

cjh1:
Otherwise we end up with functions from the previous module.


cjh1:
Fixes #1214

Issue #1196, opened 2017-06-30
No suitable VTK data type for np.sqrt() transform


ercius:
tomviz 1.0.0, windows 10, 64-bit

A strange issue occurs when running the default custom transform np.sqrt() on the sample tilt series. Apparently, applying np.sqrt() to an uint8 dtype produces a float16 dtype. VTK does not have such a dtype (not surprisingly).

This error might appear for other datasets in the future with the same dtype. Adding data type checking in tomviz.utils.set_array() could fix this.

To reproduce:
1) Load sample tilt series
2) Click Data Transforms --> Custom Transform
3) Press OK

The error:
```
Traceback (most recent call last):
  File "Transform Data", line 24, in transform_scalars
  File "C:\Program Files\tomviz\lib\tomviz\site-packages\tomviz\utils.py", line 101, in set_array
    vtkarray = np_s.numpy_to_vtk(arr)
  File "C:\Program Files\tomviz\lib\paraview\site-packages\vtk\util\numpy_support.py", line 144, in numpy_to_vtk
    vtk_typecode = get_vtk_array_type(z.dtype)
  File "C:\Program Files\tomviz\lib\paraview\site-packages\vtk\util\numpy_support.py", line 70, in get_vtk_array_type
    'Could not find a suitable VTK type for %s' % (str(numpy_array_type)))
TypeError: Could not find a suitable VTK type for float16
Failed to execute the script. (:0, )
```

mathturtle:
We should probably handle unsupported types (I doubt this is the only one) by converting them into some default.  Maybe float32 as the default?

Do we also want to handle datatypes that tomviz doesn't handle well like this?  I know some parts of tomviz don't like double/float64 data (I think that is only the tiff writer though).

cryos:
I think double is only really an issue for the TIFF writer, as it only supports single precision. In that case we should warn the user it is being converted to float, and save it as such. I can't think of any other parts that have issues with double precision numbers.

I think I remember @cquammen did some work mapping types for ITK in the Python world. We should probably extend that/mirror it for VTK. I think we support the integer types, but if it is integer type and we need to convert it should be a more accurate integer type. If it is float and we need to convert a more accurate float type would work - in this case float32.

ercius:
It would be good to at least change the custom transform to work properly "as is" if the datatype conversions are not finished for 1.1.0. It looks bad when default transform fails.

cryos:
[...]

PR #1216, opened 2017-07-28
Restructured VTK to NumPy mediation a little

cryos:
The set_scalars function checked, but would raise an exception, and the
set_array function wasn't checking. Handle the ValueType exception, and
then convert to float32 when VTK doesn't have an existing mapping for a
NumPy type.

### DIFF
--- a/tomviz/python/tomviz/utils.py
+++ b/tomviz/python/tomviz/utils.py
@@ -31,18 +31,25 @@ def get_scalars(dataobject):
     return vtkarray
 
 
+def is_numpy_vtk_type(newscalars):
+    # Indicate whether the type is known/supported by VTK to NumPy routines.
+    try:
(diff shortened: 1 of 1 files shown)


cryos:
This fixes issue #1196, as far as I can tell the previous approach wasn't doing as expected as a ValueType exception is raised. I am not entirely familiar with Python's exception mechanism, so would appreciate you verifying the logic is correct. In limited tests it fixed the problem with the custom transform and the SMALL_ ame data set.

cjh1:
Logic looks good. What makes float32 a good default conversion?

cryos:
We were using float64, but discussed it and felt it was overkill as a default but float32 should be a good choice for unknowns - VTK maps most types quite well, float16 was the cause of the bug I was fixing.

Issue #1199, opened 2017-06-30
Suggested transform


ercius:
Most reconstructions end up having artifacts at the edges surrounding the volume. This is due to edge artifacts introduced during the alignment process. I think it would be useful to introduce a clipping transform:
The only user input is the variable clipNum
```
def transform_scalars(dataset):
    """Set values outside a cirular range to minimum(dataset) to remove reconstruction artifacts"""
    
    from tomviz import utils
    import numpy as np
    
    #Input variable
    clipNum = 5 #number of pixels to clip at the edge of the volume
        
    array = utils.get_array(dataset)
    
    # Constant values
    numX = array.shape[1]
    numY = array.shape[2]
    minVal = array.min()
    
    #Find the values to be clipped
    maxR = min([numX,numY])/2-clipNum
    XX, YY = np.mgrid[0:numX,0:numY]
    RR = np.sqrt((XX-numX/2)**2 + (YY-numY/2)**2)
    clipThese = np.where(RR>=maxR)
    
    #Apply the mask along the original tilt axis direction (always the first direction in the array)
    for ii in range(0,array.shape[0]):
        curImage = array[ii,:,:] #points to the relevant 2D part of the array
        curImage[clipThese] = minVal

    # Return to tomviz
    utils.set_array(dataset, array)

mathturtle:
@ercius Is this meant to be used with tilt series or the final dataset?  I can restrict it to the one that it is meant to work with if you want.  It looks like it operates in 2D, so my guess is tilt series before the reconstruction, but I could be wrong.

@Hovden Do you have any input for which menu this would belong under?  I'm putting it in `Data Tranforms` to test it, but does it make more sense to you somewhere else?

cryos:
It removes reconstruction artifacts, and so I think it would operate on the final data sets, and it would be appropriate as a data transform in my opinion, so putting it there for now would work well I think.

mathturtle:
So set it to be applied only to reconstructed volumes?

ercius:
Yes. This should only be applied to reconstructions along the original tilt
axis. That is why it operates in 2D at every slice along one direction. It
removes artifacts outside the reconstructible volume. Thanks for including
it!

cryos:
This was merged in #1220, thanks @ercius for the additional operator, let us know how it looks if you get a chance.

ercius:
Just tested in 1.0.1-68 from yesterday. This works as expected. Thanks!

PR #1220, opened 2017-07-31
Add Clip Edges operator

mathturtle:
This branch adds it to the data transforms menu and restricts it to work for volumes only (since it seems it is supposed to be applied to reconstructed volumes).  For #1199.


cryos:
We really need to figure out how to get the text to wrap in the generated interface - very stretched. That said, it works as expected and the UI is a separate meta-issue.

PR #1221, opened 2017-08-01
Update for changes in ITK build options code in Python

cquammen:
In commit f86c9bd5bbd5c93e95aed3f0b1598629c30ec554,
WrapITKBuildOptionsPython.py was renamed to itkBuildOptions. Addressed
that change and also worked around a bug in ITK that occurs when
importing itkBuildOptions.


cquammen:
Latest commit changes the incorrect `itk.BuildOptions` to `itkBuildOptions`.

Issue #1195, opened 2017-06-30
Snapshot data and state files


ercius:
tomviz 1.0.0 on windows 10 64-bit.

If you use the  snapshot data function the new data has an asterisk next to it indicating it is not saved to disk. Saving the data (File --> Save Data) does not remove the asterisk. Also, saving a state file complains that the snapshot data and subsequent transforms can not be saved. Thus, snapshotting can not be used as it was intended.

Note: maybe this was fixed in 1.0.1? Sorry if I'm posting an already fixed issue.

cryos:
@cjh1 could you look into this please, and verify where master is, it would be good to resolve this for the next release if possible.

cjh1:
I can verify the same behavior in master. Currently we do not save the child data associated with the snapshot operator.

cryos:
This is now merged, we should hopefully have updated binaries within a few hours. Please test, this worked well in my test builds.

ercius:
I checked this today. Saving the data now works properly. Saving a child dataset now removes the asterisk and allows the state file to be saved without a warning. The data is also loaded properly with the state file.

However, the full pipeline is run when loading a state file. The reason to snapshot a dataset is to avoid running the full pipeline. Im not sure if this can be easily changed. I suggest adding a feature to the state file loading sequence to ask whether the pipe line should be run or not.

cryos:
This is a bigger change, we will see what we can do in the short term, but it may need to come with some of the bigger pipeline changes. Marking this issue as resolved, we will see what we can do about the larger issue for a future release.

PR #1222, opened 2017-08-01
Allow snapshot data to be saved a restored

cjh1:


cjh1:
Fixes #1195

cjh1:
@cryos @cquammen @mathturtle PTAL this should allow snapshot data to be save and restored. I would appreciate some QA as the change is pretty invasive, particularly any operators with child datasets.

cryos:
I will merge this tomorrow morning if no one else comments, as far as I can see this is an improvement in my testing.

PR #1232, opened 2017-08-07
Show full output by default

cquammen:
ParaView's default is to not show full output. This commit changes the
default to show the full output.

You'll have to clear out preferences to see this working in action. On Mac, delete the `tomviz1.X.Y.ini` file in `~/.config/tomviz/`. Restart tomviz and do some action that causes the Output Window to appear. The "Show full output" checkbox should be checked.

### DIFF
--- a/tomviz/MainWindow.cxx
+++ b/tomviz/MainWindow.cxx
@@ -109,6 +109,14 @@ namespace tomviz {
 MainWindow::MainWindow(QWidget* parent, Qt::WindowFlags flags)
   : QMainWindow(parent, flags), m_ui(new Ui::MainWindow)
 {
+
+  // Override the default setting for showing full messages. This needs to be
+  // done prior to calling m_ui->setupUi(this) which sets the default to false.
+  pqSettings* qtSettings = pqApplicationCore::instance()->settings();
+  if (!qtSettings->contains("pqOutputWidget.ShowFullMessages")) {
+    qtSettings->setValue("pqOutputWidget.ShowFullMessages", true);
+  }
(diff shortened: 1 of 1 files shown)


tomviz/MainWindow.cxx
```diff
@@ -109,6 +109,15 @@ namespace tomviz {
 MainWindow::MainWindow(QWidget* parent, Qt::WindowFlags flags)
   : QMainWindow(parent, flags), m_ui(new Ui::MainWindow)
 {
+
+  // Override the default setting for showing full messages. This needs to be
+  // done prior to calling m_ui-setupUi(this)
```
cryos: Missing > in the m_ui I think.

tomviz/MainWindow.cxx
```diff
@@ -109,6 +109,15 @@ namespace tomviz {
...
+  pqSettings* qt_settings = pqApplicationCore::instance()->settings();
+  if (!qt_settings->contains("pqOutputWidget.ShowFullMessages")) {
+    qt_settings->setValue("pqOutputWidget.ShowFullMessages", true);
+    std::cout << "ShowFullMessages true" << std::endl;
```
cryos: I don't think we should keep the output, the result should be obvious (I guess you wanted it for debugging)

cquammen: Gah... Thanks for catching.

cquammen:
Requested changes complete, @cryos .

cryos:
That's great, thanks @cquammen

Issue #1203, opened 2017-07-10
Crash when loading state file in new directory


ercius:
tomviz 1.0.0 windows 10

The state file contains an absolute path to the data to load. tomviz crashes without any error message if the data and state file are moved to another folder. Ideally, tomviz should look for the absolute path, then for any local data with the same file name, and then give an error message if the data cant be found.

This error makes it impossible to share state files with collaborators. I seem to remember this working previously with the data in the same directory as the state file, but now its not working.

Here is how to reproduce the error:
1. Load the sample tilt series.
2. Make some modification to the scene to get a unique state (just to make sure the state file loads)
3. Save the data (I tested XMF/HDF5 and TIFF) to a specific location (e.x. Desktop)
4. Save the state file to the same location (e.x. Desktop)
5. Load state file to ensure it works properly
6. Create a new folder (e.x. \Desktop\temp\) and move data and state file to this new directory
7. Load the state file from the new location
--> tomviz crashes

I am able to load the state file and data by manually setting the path in the state file XML code:
Original crashes:
`
      <Annotation key="tomviz.filename" value="C:/Users/Peter/Desktop/temp1.xmf" />
`
Manually updated working:
`
      <Annotation key="tomviz.filename" value="C:/Users/Peter/Desktop/temp/temp1.xmf" />
`

cryos:
We will have to dig into this, we did some work to make paths relative but it would seem this is not working here. Thanks for reporting the issue, the default was intended to be to state file paths relative so that we could enable relocation/sharing.

ercius:
I just contacted a collaborator. He used 0.9.3 to open the state file I had trouble with in 1.0.0. Seems to a new issue. Thanks for looking into it!

mathturtle:
@ercius Can you send me the state file you were having trouble with?  I have been trying and failing to reproduce this by saving state files and moving them.  Was your state file created in a different version of tomviz than you loaded it in?

ercius:
@mathturtle @cryos Data and state files sent via email separately.

I also reproduced this using my steps above and was able to do so. Im using tomviz 1.0.0 on Windows 10. You may be skipping a step above. You need to move BOTH the data and the state file. If you move ONLY the state file then tomviz does not crash because the location of the data file remains fixed.

Moving both the state file and the data file "simulates" sending data to a collaborator who has a different file structure.

ercius:
I just tested this and it seems to work as expected now as of 1.0.1-100

PR #1233, opened 2017-08-07
Modify the paths to saved data in the state file

mathturtle:
If the user uses the Save Data function, the path to the saved file is
preferred when loading the state file and reading the data back in.
Make the state file save the relative path to the saved data so that the
state + data files can be moved.

### DIFF
--- a/tomviz/Utilities.cxx
+++ b/tomviz/Utilities.cxx
@@ -56,12 +56,34 @@
 #include <QMessageBox>
 #include <QString>
 
+namespace tomviz {
+
+const char* Attributes::TYPE = "tomviz.Type";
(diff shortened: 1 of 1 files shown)


mathturtle:
This is for #1203.  It won't help with the state file he has already saved though since they are supposed to be saved with relative paths.

mathturtle:
So I did find an issue with this "Load output from Save Data rather than the original" behavior.  If you ran operations before saving the data, it treats the file as if it were the original and runs those operations on it again...

Load dataset.
Square root
Save Data
Save State
Reset/close & reopen tomviz
Load State --> data will have been square rooted twice.

mathturtle:
I guess I'll make a new issue for this

PR #1238, opened 2017-08-09
Ensure child data is removed with operators

cryos:
Previously child data of operators could be missed when the top-level
data source was removed. The use of deleteLater also meant that the
rendered image might have some modules of child data sources due to
signal propagation.


cryos:
I think this should address the issues we found earlier, and appears to work locally.

cjh1:
@cryos There is a formatting issue

cryos:
Thanks for testing, fixed the formatting issue.

PR #1239, opened 2017-08-10
Only display remaining time to 1 decimal place

cryos:
The C++ reconstruction was using quite a large accuracy, this makes the
layout jitter, and is also not particularly useful. Just show seconds
estimated to the nearest single decimal place and the dialog is more
stable with an estimate that is easier to read.


cryos:
I think the dialog still needs a little work as it jitters a little when we move from double to single digit seconds in my tests, but it is far more stable than before.

PR #1241, opened 2017-08-16
Gradient ui fixes

alvarosan:


alvarosan:
@cryos ,   this includes most of the changes we discussed.  The missing piece is separating the 1D-TF from the one represented by TransferFunctionBoxItem in 2D mode.

alvarosan:
You might be interested @Hovden.

cryos:
Just taking a look right now, some initial thoughts on the interaction. I think the combo box is a much better way of doing this, the border around the charts is big, and we should reduce that significantly. When switching to other module types it should return to the standard scalar opacity editor, and then switch back to whatever is appropriate for the volume module when selected.

I like the new layout a lot better in general, and the interaction/layout feels more intuitive, and just discoverable enough.

alvarosan:
Fixed the margin and switch-back issues.  Also made sure HistogramWidget opacity changes are updated in the 2D-tf (in case it is active in ModuleVolume).

PR #1242, opened 2017-08-17
Export module data

mathturtle:
@cryos This is working for the contour and volume modules, the orthogonal slice works as far as my testing can tell (I'm 99% sure the math for which slice to export is right and I've tested the xy slices with the OME-TIFF files that have the slice number written in the image, but if you have some test data with xz and yz slices where I can be sure which slice got exported I'd be happy to do more testing).  Visually they look right, but that doesn't mean much for most of our data since adjacent slices are always similar.

I just threw in some formats for the 'Mesh' and 'image slice' output types.  If there's something we want to support but I missed let me know.  I'm pretty sure we don't want the full list of writers for that data type from ParaView.

Depends on this ParaView branch (just merged): https://gitlab.kitware.com/paraview/paraview/merge_requests/1853

### DIFF
--- a/tomviz/ExportDataReaction.cxx
+++ b/tomviz/ExportDataReaction.cxx
@@ -19,6 +19,7 @@
 #include <pqActiveObjects.h>
 #include <pqCoreUtilities.h>
 #include <pqProxyWidgetDialog.h>
+#include <vtkArrayCalculator.h>
 #include <vtkDataArray.h>
 #include <vtkImageData.h>
 #include <vtkNew.h>
@@ -216,9 +217,25 @@ bool ExportDataReaction::exportData(const QString& filename)
               << std::endl;
     vtkNew<vtkImageData> newImage;
     newImage->DeepCopy(imageData);
-    vtkDataArray* scalars = imageData->GetPointData()->GetScalars();
+    vtkSmartPointer<vtkDataArray> scalars =
+      imageData->GetPointData()->GetScalars();
     double range[2];
     scalars->GetRange(range);
+
+    if ((imageType == VTK_FLOAT || imageType == VTK_DOUBLE) &&
+        (range[0] >= 0 && range[1] <= 1)) {
+      std::cout << "Converting normalized floating point values to integers in "
+                   "the range 0-255."
+                << std::endl;
+      vtkNew<vtkArrayCalculator> calc;
+      calc->AddScalarVariable("scalars", scalars->GetName());
+      calc->SetFunction("floor(scalars*255 + 0.5)");
+      calc->SetResultArrayName("result");
+      calc->SetInputData(imageData);
+      calc->Update();
+      scalars = calc->GetDataSetOutput()->GetPointData()->GetArray("result");
+      scalars->GetRange(range);
+    }
     vtkNew<vtkUnsignedCharArray> charArray;
     charArray->SetNumberOfComponents(scalars->GetNumberOfComponents());
     charArray->SetNumberOfTuples(scalars->GetNumberOfTuples());
--- a/tomviz/ExportDataReaction.cxx
+++ b/tomviz/ExportDataReaction.cxx
@@ -29,6 +29,7 @@
 #include <vtkSmartPointer.h>
 #include <vtkTIFFWriter.h>
 #include <vtkTrivialProducer.h>
+#include <vtkUnsignedCharArray.h>
 
 #include "ActiveObjects.h"
 #include "ConvertToFloatOperator.h"
@@ -122,6 +123,26 @@ void ExportDataReaction::onTriggered()
   }
 }
 
+namespace {
+
+template <typename FromType, typename ToType>
+void convert(vtkDataArray* outArray, int nComps, int nTuples, void* data)
+{
+  FromType* d = static_cast<FromType*>(data);
+  ToType* a = static_cast<ToType*>(outArray->GetVoidPointer(0));
+  for (int i = 0; i < nComps * nTuples; ++i) {
+    a[i] = static_cast<ToType>(d[i]);
+  }
+}
+
+template <typename FromType>
+void convertToUnsignedChar(vtkDataArray* outArray, int nComps, int nTuples,
+                           void* data)
+{
+  convert<FromType, unsigned char>(outArray, nComps, nTuples, data);
+}
+}
+
 bool ExportDataReaction::exportData(const QString& filename)
 {
   auto server = pqActiveObjects::instance().activeServer();
@@ -153,7 +174,7 @@ bool ExportDataReaction::exportData(const QString& filename)
   vtkSmartPointer<vtkSMSourceProxy> producer;
   producer.TakeReference(vtkSMSourceProxy::SafeDownCast(
     pxm->NewProxy("sources", "TrivialProducer")));
-  auto trivialProducer =
+  vtkSmartPointer<vtkTrivialProducer> trivialProducer =
     vtkTrivialProducer::SafeDownCast(producer->GetClientSideObject());
   trivialProducer->SetOutput(data);
   trivialProducer->UpdateInformation();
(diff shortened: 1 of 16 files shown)


ercius:
Can this be extended to the non-ortho slice module?

mathturtle:
@ercius I plan to extend it to the non-ortho slice.  I just haven't finished that part of it.  The image from a non-ortho slice is no longer rectangular... so I have to do some extra stuff to it to make it rectangular again.  I can get the data, but then what do I default the area outside the slice to? (If I can set that, I have to see what vtkImageReslice is doing)

mathturtle:
So it turns out you can get the texture image from the non-orthogonal slice and I've made it export that for now.  This image is *much* bigger than it needs to be since it is sized for the texture plane and the rounded up to the next power of 2.  It looks like it leaves the uninitialized space as zeros.

Remaining issues:
* Some formats only support certain data types (PNG only unsigned 8 and 16 bit int for example).  Do we auto-convert or block other data?
* Is the list of formats complete (probably not)?  What other formats do we want to support exporting to?

mathturtle:
And rebased for merge conflicts...

ercius:
I have not tried this out yet. However, there are two things I would want to export. Either an image 8 or 16 bit is ok (tiff, png, etc.). I would also like to export the raw data.

If you put zeros for pixels outside the object then that assumes that the minimum value is 0.  This is not always true.

I was able to implement this capability this using VTK in python external to tomviz using
vtk.vtkImageReslice() and vtk.vtkImageSlabReslice(). If you are using similar functions then I can suggest how to set these functions and the write up.

mathturtle:
> If you put zeros for pixels outside the object then that assumes that the minimum value is 0. This is not always true.
>
> I was able to implement this capability this using VTK in python external to tomviz using
> vtk.vtkImageReslice() and vtk.vtkImageSlabReslice(). If you are using similar functions then I can suggest how to set these functions and the write up.

I am using `vtkImageReslice` to get the slices to export.  I'm not sure what its behavior is with points outside the original volume.  It may take the nearest point, but it may leave them 0.  If you have any pointers to offer on `vtkImageReslice` then please let me know.

> I have not tried this out yet. However, there are two things I would want to export. Either an image 8 or 16 bit is ok (tiff, png, etc.). I would also like to export the raw data.

You can export the raw data if the format supports it... which some like VTK Image Data will.  And I copied the code that auto-converts double to float for TIFF files.  I was just wondering if I needed to do something similar for other image formats like PNG or JPEG which don't support floating point at all.

ercius:
> If you have any pointers to offer on vtkImageReslice then please let me know.
 I use `vtkImageReslice.SetResliceAxesDirectionCosines()` to set the location of the slice based on the point and normal reported by tomivz's Slice visualization module. I also use vktImageReslice.AutoCropOutputOn() to keep the output to only within the confine of the data extent. It would be very useful to use the data minimum rather than just 0 as the pixel value outside the volume. I can share the code if you want to see it all.

>  I was just wondering if I needed to do something similar for other image formats like PNG or JPEG which don't support floating point at all.
Yes. converting the data to fit in the 8-bit range allowed by other image formats is necessary. Otherwise, NOT(WYSIWYG).

mathturtle:
> Yes. converting the data to fit in the 8-bit range allowed by other image formats is necessary. Otherwise, NOT(WYSIWYG).

I agree.  That is what I have implemented right now: image formats that don't support the current type will convert the data to a type they support (I've put in all the conversions I have run across but there may be more).  But if the format supports the type, then the data will be written as is.

There is still an issue I've run into: if the image is normalized floating point (range 0-1), then the conversion will end up with an image of all zeros (black).  Do we want to try to detect this automatically and rescale before the conversion?  Pop up and warn/ask the user?

cryos:
Floating point normalized should likely be scaled automatically to the 0-255 range for unsigned char based formats.

mathturtle:
An example of the non-orthogonal slice producing an image that is too big:

toobigimage

According to the documentation, vtkImageReslice::AutoCropOutputOn() does this:
```
Turn this on if you want to guarantee that the extent of the output will be large enough to ensure that none of the data will be cropped (default: Off). 
```

I have observed that it behaves exactly as described.  If you set the output extent, it will use whatever you set, but if you leave the output extent blank and turn this on it will size the image appropriately so that nothing is cropped.

cryos:
Trying the simplest thing, I went to export from the orthogonal slice, using the default (was PNG here), and got the error message:
```
~/src/ParaView/ParaViewCore/ServerManager/Core/vtkSMWriterFactory.cxx, line 330
vtkSMWriterFactory (0x559283cce8e0): No matching writer found for extension: png
```
It seemed to hit this for all types, I guess this is the ParaView version bump you needed for the XML? It is on 100/15663 objects, but just wanted to confirm the SHA was recent enough in the superbuild for this change now?

mathturtle:
> I guess this is the ParaView version bump you needed for the XML?

Exactly.  The error message is essentially 'No writer that can be connected to this proxy is available for extension .png'  ParaView didn't like connecting a trivial producer to writers so I had to remove the restrictions that were blocking this.  And I'm not sure if the SHA in the superbuild is new enough.

mathturtle:
Digging around in git, the SHA in the superbuild is too old.  It was updated for @cjh1's deadlock fix to a ParaView commit merged August 2 and my branch to fix the writers was merged August 17.

cquammen:
@mathturtle with the ParaView SHA update, this should be ready to go. Is it ready for a final review?

cryos:
Testing this on a mesh, trying to export to the default STL format, using test as the name for the file, I get a segfault here (Linux):

#0  tomviz::ExportDataReaction::exportData (this=this@entry=0x55555a864e70, 
    filename=...) at /home/marcus/src/tomviz/tomviz/ExportDataReaction.cxx:198
#1  0x00005555556dd671 in tomviz::ExportDataReaction::onTriggered (
    this=0x55555a864e70)
    at /home/marcus/src/tomviz/tomviz/ExportDataReaction.cxx:123

It looks like imageData is null in this case.

mathturtle:
Apparently when I was adding code to handle the special cases for images (the format not supporting a data type autoconversions) I broke the non-image exports.  Fixed now and rebased so I didn't have to rebuild a really old ParaView.

Note: I did have to rebuild an older ParaView, I was getting some strange crashes when I tried this with my existing ParaView master build.  It would print `No such property 'Visibility'` and then crash when loading data.

tomviz/ExportDataReaction.cxx
```diff
@@ -0,0 +1,280 @@
...
+      trivialProducer->UpdateInformation();
+      trivialProducer->Update();
+      producer->UpdatePipeline();
+    }
+
+    if ((strcmp(writerName, "vtkPNGWriter") == 0 &&
+         (imageType != VTK_UNSIGNED_CHAR || imageType != VTK_UNSIGNED_SHORT)) ||
+        (strcmp(writerName, "vtkJPEGWriter") == 0 &&
+         imageType != VTK_UNSIGNED_CHAR)) {
+      std::cout << "File type does not support the current data type, converting "
```
cryos: We can leave this for now, but I think it is much better to pop up a message box, with an option to silence forever that write to the standard out. If you want to write to terminal qDebug/qWarning would be more consistent here.

tomviz/ExportDataReaction.cxx
```diff
@@ -0,0 +1,280 @@
...
+{
+  convert<FromType, unsigned char>(outArray, nComps, nTuples, data);
+}
+}
+
+bool ExportDataReaction::exportData(const QString& filename)
+{
+  auto server = pqActiveObjects::instance().activeServer();
+
+  vtkSmartPointer<vtkDataObject> data = this->m_module->getDataToExport();
```
cryos: Don't use this->, could just use auto for return type here too.

PR #1252, opened 2017-08-22
Add Mac OS X builds on Travis

cjh1:
This PR adds supposes for running Mac OS X build on Travis. It works, however, there is a lot of contention for the Mac virtual machines and builds can end up waiting for 30+ minutes. I have contacted circleci about getting access to their Mac infrastructure as an alternative.

Note we now have three separate Travis jobs:

- python 2.7, acquisition test, flake8, clang-format (linux)
- python 3.4, acquisition test, flake8 (linux)
- mac os x

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ matrix:
     - os: linux
       dist: precise
       python: 3.4
-      env: TASKA="pytest(acquisition), flake8"
+      env: TASKS="pytest(acquisition), flake8"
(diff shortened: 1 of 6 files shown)


cjh1:
So I investigated CircleCI, but with their service its an either or thing, so we wouldn't be able to running Mac and Linux builds. So for now we are going to go with Travis and remove the hard requirement that the Travis build completes before a PR can be merged.

PR #1253, opened 2017-08-23
Improve location of "Separate Color Map" checkbox

cquammen:
It was stuck in the upper left corner without a margin and didn't follow the capitalization scheme of the other labels. This patch centers it to be in line with other widgets and sets margins to be more consistent with the other widgets.

This has been bothering me.

Old **Volume** Panel

image

New **Volume** Panel

image

New **Orthogonal Slice** Panel

image

New **Slice** Panel

image

New **Threshold** Panel

image

### DIFF
--- a/tomviz/ModulePropertiesPanel.cxx
+++ b/tomviz/ModulePropertiesPanel.cxx
@@ -83,11 +83,11 @@ void ModulePropertiesPanel::setModule(Module* module)
 
   deleteLayoutContents(ui.PropertiesWidget->layout());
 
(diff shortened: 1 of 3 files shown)


cryos:
This looks better, but it doesn't compile for me locally, nor on the CI builds, ui_ModulePropertiesPanel.h:76:31: error: no member named 'setText' in 'QWidget'

cquammen:
Sorry about that, I'm not sure how the `text` property got set for the `QWidget`. I removed it and it should compile with the most recent change, which is an amended commit.

Issue #1247, opened 2017-08-18
Apply/Ok buttons can cause multiple instances of the same pipeline running


mathturtle:
Load a tilt series.  Mark as tilt series.  Reconstruct it with backprojection C++.

Double click on the "Set Tilt Angles" operator to open its editor window.  Click Apply (the reconstruction should start re-running).  While the reconstruction is running go over to the operator dialog and click apply again.  There will now be two reconstructions running (you can see this in the UI, the current slice will jump around as they fight over which is running).

mathturtle:
What is the correct behavior here?  Cancel the currently running pipeline in favor of the updated one?

cryos:
Yes

mathturtle:
Ok, I'll work on this then.

ercius:
This could be tricky. I would not want to accidentally stop an hour long
reconstruction by pressing ok/apply by accident. Maybe a warning box should
be thrown up?

mathturtle:
Good point @ercius.  That should be doable.

mathturtle:
I think we can close this now.

mathturtle:
ping

cryos:
Agreed, closing.

PR #1261, opened 2017-09-08
Edit operator in running pipeline

mathturtle:
Warns the user, cancels the currently running pipeline and then applies the edits for #1247.  Except that the cancel sets a state variable that is reset when the new pipeline starts and the background thread doesn't get it unless it checks in a very narrow window (so still broken, thus WIP).

@cjh1 Pushed so you can think about it and experiment... and something you should consider when designing the new pipeline.  You can even roll this branch in with that one if it will take too much time to work this out now.

### DIFF
--- a/tomviz/DataSource.cxx
+++ b/tomviz/DataSource.cxx
@@ -1082,7 +1082,8 @@ void DataSource::cancelPipeline(std::function<void()> canceled)
 {
   if (this->Internals->Future) {
     if (canceled) {
-      connect(this->Internals->Future, &PipelineWorker::Future::canceled, canceled);
+      connect(this->Internals->Future, &PipelineWorker::Future::canceled,
+              canceled);
     }
     this->Internals->Future->cancel();
   }
--- a/tomviz/DataSource.cxx
+++ b/tomviz/DataSource.cxx
@@ -1076,6 +1076,11 @@ void DataSource::resumePipeline()
   executeOperators();
 }
 
+void DataSource::cancelPipeline()
+{
+  this->Internals->Future->cancel();
+}
+
 void DataSource::setPersistenceState(DataSource::PersistenceState state)
 {
   this->Internals->PersistState = state;
--- a/tomviz/DataSource.cxx
+++ b/tomviz/DataSource.cxx
@@ -1070,10 +1070,12 @@ void DataSource::pausePipeline()
   this->Internals->PipelinePaused = true;
 }
 
-void DataSource::resumePipeline()
+void DataSource::resumePipeline(bool execute)
 {
   this->Internals->PipelinePaused = false;
-  executeOperators();
+  if (execute) {
+    executeOperators();
+  }
 }
 
 void DataSource::cancelPipeline()
--- a/tomviz/DataSource.cxx
+++ b/tomviz/DataSource.cxx
@@ -1078,9 +1078,12 @@ void DataSource::resumePipeline(bool execute)
   }
 }
 
-void DataSource::cancelPipeline()
+void DataSource::cancelPipeline(std::function<void()> canceled)
 {
   if (this->Internals->Future) {
+    if (canceled) {
+      connect(this->Internals->Future, &PipelineWorker::Future::canceled, canceled);
+    }
     this->Internals->Future->cancel();
   }
 }
--- a/tomviz/DataSource.cxx
+++ b/tomviz/DataSource.cxx
@@ -1078,7 +1078,9 @@ void DataSource::resumePipeline()
 
 void DataSource::cancelPipeline()
 {
-  this->Internals->Future->cancel();
+  if (this->Internals->Future) {
+    this->Internals->Future->cancel();
+  }
 }
 
 void DataSource::setPersistenceState(DataSource::PersistenceState state)
--- a/tomviz/DataSource.h
+++ b/tomviz/DataSource.h
@@ -167,7 +167,7 @@ public:
 
   // Resume the automatic execution of the operator pipeline, will execution the
   // existing pipeline. If execute is true the entire pipeline will be executed.
-  void resumePipeline(bool execute=true);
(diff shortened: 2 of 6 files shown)


cjh1:
Thanks @mathturtle I will take a look.

cjh1:
OK, so I have fixed up the issue to do with two pipeline executions being run at the same time. We now essential wait for the first execution to cancel before re-executing the pipeline.

mathturtle:
So I just tested this out and all the dialogs seem to come up and act the way I expect them to.  What was the case you said needed looking at @cjh1?

cjh1:
@mathturtle Not sure what I was seeing, but works for me now. @cquammen Could you review this branch as an independent third party? Thanks.

cquammen:
Looking now.

cquammen:
Hmm, I am trying to update the **Set Tilt Angles** operator with new angles after the **Reconstruction** operator has finished. When I click **Apply** or **OK**, the **Reconstruction** operator does not re-execute as I would expect.

cquammen:
Editing the tilt angles while the **Reconstruction** operator is executing brings up a dialog to either stop the pipeline or not. When I stop the pipeline, I see green checkmarks next to each operator and no indication that things need to be updated. Going back and changing the tilt angles again does not trigger a re-execution of the **Reconstruction** operator.

cquammen:
Alternative test for the issue at hand:

* Sample Data -> Generate Random Particles
* Data Transforms -> Add a Constant. Leave it at 0.0.
* Data Transforms -> Perona-Malik Anisotropic Diffusion. 100 iterations will run for a while.
* While the Perona-Malik Anisotropic Diffusion operator is running, click on the **Add a Constant** operator in the **Pipelines** panel. Change the constant and click **Apply**. No cancellation dialog appears.
* Two dialogs with the progress of Perona-Malik operator appear, and then there is a crash. The relevant portion of the stack trace is:

```
2  libvtkpqWidgets-pv5.4.1.dylib 	0x000000010a6c0e1e pqConsoleWidget::printString(QString const&) + 126 (pqConsoleWidget.cxx:500)
13  libvtkpqCore-pv5.4.1.dylib    	0x000000010a40e736 pqOutputWidget::pqInternals::displayMessageInConsole(QString const&, pqOutputWidget::MessageTypes) + 214 (pqOutputWidget.cxx:264)
14  libvtkpqCore-pv5.4.1.dylib    	0x000000010a40dc02 pqOutputWidget::displayMessage(QString const&, pqOutputWidget::MessageTypes) + 146 (pqOutputWidget.cxx:431)
15  libvtkpqCore-pv5.4.1.dylib    	0x000000010a40f580 OutputWidgetInternals::OutputWindow::DisplayText(char const*) + 112 (pqOutputWidget.cxx:84)
16  libvtkCommonCore-pv5.4.1.dylib	0x00000001157e6786 vtkOutputWindow::DisplayWarningText(char const*) + 54 (vtkOutputWindow.cxx:119)
```

mathturtle:
I'll take a look.

mathturtle:
> Going back and changing the tilt angles again does not trigger a re-execution of the Reconstruction operator.

Are you sure you have the right branch built @cquammen?  I tested this just now and the re-execution happened exactly like I expected.

Your second test case... I think I see what I missed.  I had forgotten that we also supported editing operators via the properties panel.

cquammen:
Yes, the SHA-1 in the title bar matched the last commit in this branch. Perhaps I misunderstood the testing directions? Does changing one or more of the tilt angles trigger a pipeline update for you?

mathturtle:
That depends how I change them (I think I see the confusion now).  If I double click on the 'Set Tilt Angles' operator, edit its properties and click Apply or Ok on the dialog then it does trigger an update.  If I click on the dataset and scroll down in its properties to the tilt angles and edit them there, then it just appends a new 'Set Tilt Angles' to the end of the pipeline (which is executed after the current pipeline finishes).

cquammen:
Even editing them in the dialog and clicking Apply or OK doesn't seem to trigger a pipeline update for me.

mathturtle:
I just tried this out on my mac and I still can't reproduce what you're seeing Cory.  If someone else wants to take a look at this, go ahead.

cquammen:
I tried it again with my alternative test listed above with the most recent commit and it works as expected now.

cquammen:
However, I tried the initial steps (loading in tilt series, Set Tilt Angles, Simple Back Projection (C++)). If I let the pipeline finish, then go back and edit the tilt angles, the pipeline re-executes as expected. If I edit the tilt angles while the pipeline is running, tomviz now asks to cancel the pipeline. However, when I do, further changes to Set Tilt Angles do not trigger a re-execution of the pipeline. I have to delete the Reconstruction pipeline item, then re-add it for the pipeline to re-run when the tilt angles are modified.

mathturtle:
How are you managing that?  I'm still getting the expected behavior: it cancels the pipeline and immediately triggers a re-execution with the new data.  And changing the Set Tilt Angles again still works as far as I can tell.  Maybe you're trying something that isn't occurring to me to test, but I think I was pretty thorough just now and I couldn't find any cases that failed to re-execute when they should.

cquammen:
Shawn and I spoke offline. My test case differs from his tess in that it involves the cancellation dialog coming up and me clicking "Yes" after the pipeline has already finished executing.

mathturtle:
@cjh1 I fixed the obvious issue of the operator finishing before they hit yes on the dialog to cancel and apply changes.

I think there may still be a threading issue there if the two happen at the same time.  The root cause of the bug is that we are pausing the pipeline unconditionally and only unpausing it in the `whenCanceled()` callback.  So if that callback is never called, the pipeline never gets unpaused.

I think we need to be listening for the operator finishing/canceling before we show the 'Do you want to cancel...' dialog in order to avoid the threading issue.

Issue #1258, opened 2017-08-30
Direct coloring of floating-point scalars does not work in Slice, OrthogonalSliceModules


cquammen:
`vtkTexture` did not support direct coloring (i.e., not color mapping) of floating-point scalars. A fix is posted https://gitlab.kitware.com/vtk/vtk/merge_requests/3220. Once that modification to VTK bubbles up to the version of ParaView tomviz uses, this issue will be fixed.

cquammen:
This has bubbled up to Tomviz.

PR #1263, opened 2017-09-13
Direct scalar coloring in non-orthogonal plane

cquammen:
Enable direct scalar coloring in the Slice module for floats and doubles.

Partially address #1258.

To test:

* Load an image
* Run **Sample Data -> Generate Random Particles**
* Add **Slice** module
* Uncheck **Color Map Data** in the Properties panel
* The data array should be rendered as intensity, black to white.

### DIFF
--- a/tomviz/vtkNonOrthoImagePlaneWidget.cxx
+++ b/tomviz/vtkNonOrthoImagePlaneWidget.cxx
@@ -46,6 +46,7 @@
 #include "vtkScalarsToColors.h"
 #include "vtkSphereSource.h"
 #include "vtkStreamingDemandDrivenPipeline.h"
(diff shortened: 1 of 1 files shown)


cquammen:
Hmm, circleci seems to be failing with

`/build.sh: line 4: pip3: command not found`

cjh1:
@cquammen I am looking at it

cjh1:
@cquammen Circle is happy now.

cquammen:
Happy Circle, happy life, isn't that the expression? Travis Xcode doesn't seem too happy, but we aren't expecting that to work, are we?

cjh1:
@cquammen Yes, Travis is having trouble with their infrastructure, like they need to get more! Ignore Xcode for now.

PR #1266, opened 2017-09-19
Initial support for multicolor datasets

cquammen:
This topic enables loading multicomponent (e.g., RGB) datasets into tomviz. It removes some checks that restricted such datasets from being loaded.

In addition, it modifies the histogram computation to handle multicomponent datasets by computing the histogram over the magnitudes of the data values at each voxel. A future enhancement could enable selection of the component to use for computing the histogram.

### DIFF
--- a/tomviz/CentralWidget.cxx
+++ b/tomviz/CentralWidget.cxx
@@ -97,8 +97,8 @@ void PopulateHistogram(vtkImageData* input, vtkTable* output)
   switch (arrayPtr->GetDataType()) {
     vtkTemplateMacro(tomviz::CalculateHistogram(
       reinterpret_cast<VTK_TT*>(arrayPtr->GetVoidPointer(0)),
-      arrayPtr->GetNumberOfTuples(), minmax[0], pops, inc, numberOfBins,
-      invalid));
+      arrayPtr->GetNumberOfTuples(), arrayPtr->GetNumberOfComponents(),
+      -1 /* Magnitude */, minmax[0], pops, inc, numberOfBins, invalid));
     default:
       cout << "UpdateFromFile: Unknown data type" << endl;
   }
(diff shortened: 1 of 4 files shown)


Hovden:
I'm working with a new student to get this pull request compiled and tested. It should also help with more testing and development on our side.

mathturtle:
@cquammen Jonathan is testing with Paraview master (I didn't think of telling him to check out the version we use before he started the build). Hopefully that doesn't break anything.

cquammen:
ParaView master should be fine.

cquammen:
Please review.

cquammen:
Scratch that, let me resolve the conflict first.

jtschwar:
Hi all, 
I've come across two crashes when an RGB dataset is implemented. First, when I select the contour option I get the following error: "Source scalars have more than one component! Cannot categorize!". Second, if I check on the "separate color map" square in volume properties the program will crash.

cquammen:
Hi @jtschwar. Thanks for your review! I'll work on better handling of these cases.

cryos:
I think we should fix some of the more obvious crashes before merging, the code proposed all looks reasonable to me. Testing with existing data doesn't appear to have caused any regressions in functionality either. There is no rush to merge this, we can let it mature a little.

jtschwar:
Hi Cory, 
Your branch now won't let me open tomviz. If I check it out, the program crashes during startup whenever I have the branch attached.

cryos:
As this introduces no regressions to existing capabilities I would like to merge it, and see if there are any unexpected side effects. Above looks like a local build issue to me, and we can help resolve that separately.

cquammen:
Thanks, @cryos. I am away this week and when I get back will work on addressing the issues raised in this MR.

cryos:
Sounds good @cquammen, I would like to be able to test some of this in the main development branch, but it would be great to circle back and make this more robust soon.

PR #1272, opened 2017-09-22
Remove space that is breaking ITK configure command

mathturtle:


mathturtle:
I found some issues with the build instructions while helping someone get started on the project.  The extra space in the ITK configure command make copy-pasting the command fail.

BUILDING.md
```diff
@@ -79,7 +79,7 @@ you will need to specify the correct generator for the installed compiler.
       -DITKGroup_Segmentation:BOOL=ON \
       -DITKGroup_Registration:BOOL=ON \
       -DITKGroup_Nonunit:BOOL=ON \
-      -DPython_ADDITIONAL_VERSIONS:STRING=3 \ 
+      -DPython_ADDITIONAL_VERSIONS:STRING=3 \
```
cquammen: For ITK master, this variable isn't even used. I'm not sure about the latest release, though.

mathturtle: Oh, I almost forgot, we use ITK 4.12.0 now and not a random commit off master.  Updated that too.

mathturtle: But it would be good to go through the ITK build arguments and see if we can clean any of them up.

cryos: This is how @thewtex told me to get Python 3 for ITK, it seemed to work when I tried it, if it is incorrect it would be good to change it to something that will work.

thewtex: +1 for keeping it to ensure that CMake finds Python 3.

Thanks for removing the extra space.

mathturtle: @thewtex When I was helping the new student get ITK configured, it was finding python2 libraries from the system and the python3 executable from macports.  We had to manually set the python 3 libs and include dirs

cryos: Linux it seemed to find Python 3, I will retest later with this command line. Was this just for ITK, or ParaView too @mathturtle?

mathturtle: I think we had the same problem with ParaView.  ParaView was nice enough to warn that the python lib version and executable version were different though.  ITK was silent about that.

cryos: It is a longstanding weakness of separate modules for lib and interpreter... We can add some notes, it would be good for ITK to add the warning when they get a chance. The bigger task is to ensure we are always working with the same Python library and interpreter, but documenting the issue for now works.

thewtex: ITK will emit a warning with Python libraries and the interpreter are for different versions.

`Python_ADDITIONAL_VERSIONS` is a variable used by upstream CMake `FindPython*`.

thewtex: See also:

  https://github.com/InsightSoftwareConsortium/ITK/blob/e1b9228ef6c3e1bfcf5e20009848333848fc3d02/Wrapping/Generators/Python/CMakeLists.txt#L11-L15

PR #1276, opened 2017-09-27
Support for merging arrays from different images into a single image

cquammen:
This MR adds to tomviz the ability to select two or more images in the Pipeline view and merge them into an entirely new image in one of two ways. The first way is to merge all arrays from each image into the new image. Arrays with the same name are handled by appending the array's input number to the array name. This is a useful option when the data arrays should maintain some distinction from each other.

The second way to merge image data is to merge all data arrays into a single array by appending all components in the input arrays into a single array in the new image. This can be used to create an RGB image from three separately loaded images.

To test:

* Run tomviz
* Load an image
* Load a second image with the same extent as the first
* Using Ctrl-left mouse button, click on the two loaded images in the Pipeline. It's okay if operators and modules are also selected, only the DataSources will be considered in the selection.
* Right-click on one of the selected DataSources and select **Merge Images**.
* A dialog box will appear that looks like:

image
Choose either option. 
* Click on the new DataSource. The arrays in the new data set will be listed in the **Original Dimensions & Range** and **Transformed Dimensions and Range**.

Caveat: A progress dialog pops up when **Merge array components** is chosen. It does not repaint properly nor does it go away after the operation is complete. I haven't figured out what is going on there yet, hence a WIP.


cquammen:
@cjh1 I could use a hand figuring out why the progress dialog appears when the **Merge array components** option appears and does not repaint or go away. I'm wondering if it is a thread ownership issue? If that's the case, though, I'm not sure why it doesn't appear for the **Merge image arrays** option.

cjh1:
@cquammen I will take a look.

cjh1:
@cquammen I am not seeing the processing dialog?

cjh1:
@cquammen OK I see it if I do a "Merge array components"

cquammen:
Yes, that's when I see it two on linux and Mac.

cjh1:
@cquammen I am seeing this, not sure is a progress dialog, as I think the title would have "progress" in it?
screen

cquammen:
It is actually the  progress dialog from ParaView that is showing up. If I commented out lines 30-33 of tomviz's `ProgressBehavior.cxx`, the dialog does not appear.

cjh1:
OK, I'm not familiar with that one, I will take a look.

cjh1:
I guess the question is do we need pqProgressManager?

cjh1:
@mathturtle and I have seen this mystery dialog before in other situations :smile_cat:

cjh1:
I am tempted to comment out [this line](https://github.com/OpenChemistry/tomviz/blob/master/tomviz/Behaviors.cxx#L96). @utkarshayachit What does this behavior do for us?

cquammen:
My guess is that it was the initial progress dialog used to listen for progress updates from ParaView proxies. Tomviz has its own that works directly with operators, so my guess is we can comment out / remove the line you cite @cjh1 and be okay.

cryos:
Give it a try and see what happens.

utkarshayachit:
ParaView shows a status bar progress when readers are updating or when filters are executing. In `tomviz`, we wanted to show a popup dailog instead. So the easiest way was to hookup to ParaVIew' progress events and popup a QProgressDialog to show that work was happening.

cquammen:
I think it is safe to remove it now that operators have their own progress reporting dialog. I will do so on this PR.

cquammen:
Ready for review.

cquammen:
The phantom progress dialog has been removed and no longer appears mysteriously on linux. Ready for review.

PR #1277, opened 2017-09-27
Hide the Python console and animation dock by default

cquammen:
These docks are not used much but take up a lot of screen
real estate, so hide them.

Closes #1277.

### DIFF
--- a/tomviz/MainWindow.cxx
+++ b/tomviz/MainWindow.cxx
@@ -125,13 +125,16 @@ MainWindow::MainWindow(QWidget* parent, Qt::WindowFlags flags)
   setWindowIcon(icon);
 
   // tabify output messages widget.
-  tabifyDockWidget(m_ui->dockWidget_3, m_ui->dockWidgetMessages);
-  tabifyDockWidget(m_ui->dockWidget_3, m_ui->dockWidgetPythonConsole);
+  tabifyDockWidget(m_ui->dockWidgetAnimation, m_ui->dockWidgetMessages);
+  tabifyDockWidget(m_ui->dockWidgetAnimation, m_ui->dockWidgetPythonConsole);
   m_ui->dockWidgetMessages->hide();
+  m_ui->dockWidgetPythonConsole->hide();
 
   // don't think tomviz should import ParaView modules by default in Python
   // shell.
   pqPythonShell::setPreamble(QStringList());
+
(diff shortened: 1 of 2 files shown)


tomviz/MainWindow.cxx
```diff
@@ -125,14 +125,17 @@ MainWindow::MainWindow(QWidget* parent, Qt::WindowFlags flags)
   setWindowIcon(icon);
 
   // tabify output messages widget.
-  tabifyDockWidget(m_ui->dockWidget_3, m_ui->dockWidgetMessages);
-  tabifyDockWidget(m_ui->dockWidget_3, m_ui->dockWidgetPythonConsole);
+  tabifyDockWidget(m_ui->dockWidgetAnimation, m_ui->dockWidgetMessages);
+  tabifyDockWidget(m_ui->dockWidgetAnimation, m_ui->dockWidgetPythonConsole);
   m_ui->dockWidgetMessages->hide();
+  m_ui->dockWidgetPythonConsole->hide();
```
cryos: Won't this hide the widgets every time you open Tomviz? I don't have time to test, but I think that would also be a step backwards. It should default to being hidden, but if someone takes the time to show it then it should remain visible the next time it is opened. The converse is certainly true - it remembers I closed these widgets the next time I open Tomviz.

cquammen: Nope, this just sets the initial visibility. I can't find the docs that describe this behavior, but I believe the dock widget visibility is restored as part of the main window geometry restoration. Testing locally confirms the expected behavior.

cryos: Then that is great, but not obvious. It would be good to add a short comment to note that hide sets the default, but saved visibility will be restored.

cquammen: Found it. `pqPersistentMainWindowStateBehavior` invokes `restoreState()` in `pqSettings.cxx`, which restores the dock widget visibility, geometry, etc.. This all happens after the `hide()` calls added here, so the settings will be restored between tomviz invocations.

cryos: Sounds good, but very much not obvious, this can certainly be fixed with a short comment above the calls documenting the restore behavior taking place afterwards.

cquammen: Comment added.

PR #1279, opened 2017-09-29
Improve dialog for the RAW file reader to prevent misformed input

mathturtle:
For #1200, this improves the RAW reader dialog warning the user if the data dimensions/type doesn't match the file size (and blocking them from pressing OK until they fix it).

### DIFF
--- a/tomviz/RAWFileReaderDialog.cxx
+++ b/tomviz/RAWFileReaderDialog.cxx
@@ -160,11 +160,10 @@ void RAWFileReaderDialog::sanityCheckSize()
   }
   size_t selectedSize =
     dataSize * dims[0] * dims[1] * dims[2] * numComponents;
-  auto labelText =
-    QString("Selected parameters would read %1 of %3 bytes (%2% of the file)")
-      .arg(selectedSize)
-      .arg(static_cast<float>(selectedSize) / m_filesize * 100)
-      .arg(m_filesize);
+  auto labelText = QString("Reading %1 of %3 bytes (%2% of the file)")
+                     .arg(selectedSize)
+                     .arg(static_cast<float>(selectedSize) / m_filesize * 100)
+                     .arg(m_filesize);
   m_ui->statusLabel->setText(labelText);
   if (selectedSize != m_filesize) {
     m_ui->buttonBox->button(QDialogButtonBox::Ok)->setEnabled(false);
--- a/tomviz/RAWFileReaderDialog.cxx
+++ b/tomviz/RAWFileReaderDialog.cxx
@@ -1,0 +1,153 @@
+/******************************************************************************
+
+  This source file is part of the tomviz project.
+
+  Copyright Kitware, Inc.
+
+  This source code is released under the New BSD License, (the "License").
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+
+******************************************************************************/
+#include "RAWFileReaderDialog.h"
+#include "ui_RAWFileReaderDialog.h"
+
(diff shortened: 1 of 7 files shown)


tomviz/RAWFileReaderDialog.cxx
```diff
@@ -0,0 +1,153 @@
...
+namespace tomviz {
+
+RAWFileReaderDialog::RAWFileReaderDialog(vtkSMProxy* reader, size_t filesize,
+                                         QWidget* p)
+  : QDialog(p), m_ui(new Ui::RAWFileReaderDialog), m_reader(reader),
+    m_filesize(filesize)
+{
+  m_ui->setupUi(this);
+  connect(m_ui->dimensionX,
+          static_cast<void (QSpinBox::*)(int)>(&QSpinBox::valueChanged), this,
```
mathturtle: I would have liked to use `qOverride` here since it is less ugly, but I couldn't get the compiler to find the symbol, even with the `<QtGlobals>` header included.

cryos:
This looks great, there are a few tweaks I would like to make before we merge, but I think this will be very useful. First, I think everyone knows/expects it to be x, y, and z, so having Dimensions: and then the three boxes next to each other would be simpler/more compact. I think we can prepopulate the boxes or set tooltips, but it has been a while since I edited these.

For the data type I think that it may be more transparent for us to specify the number of bits (with the C type perhaps in brackets), and maybe have a tick box for signed versus unsigned, but I am on the fence about that. Lots of people in science are more aware of the number of bits/signedness than they are the names of the corresponding types in C.

I think we can shorten it to simply Components on the last box. The biggest change that I would find most helpful is in the text, simple saying it does not match is nowhere near as useful as telling me where we are, so if I select 8 bit integer (char) is it half of the file size, twice the file size because I got Z wrong, etc. I think I would actually put that information at the bottom of the dialog too, so I can make a few changes and see where I am.

ercius:
I have not seen this new dialog, but I have one request if it's not already implemented. Having a display of the raw file size in bytes and the requested file size to compare with are very useful. You can see if you are off by 2x, etc. and modify boxes accordingly.

cryos:
We were discussing it yesterday (and it was one of my points above in the last paragraph). One question, I wasn't sure whether raw byte size or percentage of the file to be read (or maybe both) would be most useful.

cryos:
The dialog layout still doesn't seem right, and I think we should be much more concise with the message. Let's make it span the whole width of the dialog at the bottom, and read:

Selected parameters would read xxx of yyy bytes (zz% of the file)

It is a fragment, so no need for a full stop/period either.

cryos:
Another thing, can we show the file name in the dialog title? It often contains metadata, and I want to use it to fill in the dialog.

cryos:
This is looking really nice, I think the minimums for the x, y, z entry boxes are off (one is bigger than the other two, they should all be equal, and reserving space for 4 digits should be great.

cryos:
Maybe it is the text box, and perhaps shortening it to "Reading xxx of yyy bytes (zz%)" would convey all that is needed.

cryos:
In case you want to preview it @ercius the proposed dialog is above, with a few more tweaks on the way.

mathturtle:
Here is the dialog with my latest tweaks applied.

new_raw_reader

cryos:
I still think the dimensions could be four digits rather than 7 or 8, but this looks good (and we can tweak it more). Just going through the code quickly...

mathturtle:
That might be the size of the dialog... they try to take up space by default.  I'm not sure we have that much space to shrink the dialog (although there is some).  I had to add a stretch in to force the components spin box to be small.  I could try to fix it the same way, but the minimum width is set in pixels rather than something useful like digits.

mathturtle:
It looks like the minimum width designer will give is based on the maximum for the spin box.  I didn't want to set an arbitrary maximum that will break as we feed it bigger data, so I used MAX_INT.  Thus the 10 digits of space they take up.  I think I can make them smaller if you tell me a more reasonable maximum size to use that noone will run up against.

Issue #1278, opened 2017-09-28
Incorrect dimension calculation from MRC data set


ercius:
I think there is a problem with the dimension size calculation in the data properties panel for MRC data sets.

The MRC file meta data contains pixel sizes in Angstrom. tomviz converts to nanometer and then multiplies by the number of pixels along a dimension to get the physical size of each dimension.
length = pixel_size * numPix

This is incorrect. The proper calculation should be:
length = pixel_size * (numPix -1)

A 512x512x11 data set I recently generated should be exactly 100nm across, but tomviz calculates it incorrectly as noted above:
![image](https://user-images.githubusercontent.com/15657108/30994827-2d64639e-a46b-11e7-97f8-fdfc8ffc2560.png)

cryos:
Closing this issue as the pull request has now been merged. Please reopen if the issue is not fully resolved.

PR #1281, opened 2017-10-03
Display point data based size rather than voxel based size

mathturtle:
@cryos Hopefully I got everywhere that was assuming voxel based size and had the extra +1 in the calculations.


cryos:
This fixes #1278

PR #1290, opened 2017-10-12
Allow reading a subset of a RAW file

mathturtle:


mathturtle:
@cryos This might be another small change worth getting in the release.

PR #1299, opened 2017-10-17
Recent files menu: delay exists check until user selects file

mathturtle:


tomviz/RecentFilesMenu.cxx
```diff
@@ -285,19 +281,19 @@ void RecentFilesMenu::stateTriggered()
...
+      return;
+    }
+  } else {
+    // // If the user tried to open a recent file that no longer exists, remove
+    // it from the recent files
     qWarning() << "Error: file '" << actn->iconText() << "' does not exist.";
-    return;
   }
 
-  QString filename = actn->data().toString();
```
cryos: Removing this causes a compile error below, where filename is used in the conditional but no longer exists.

mathturtle: It was moved inside an if statement above, I moved the declaration back out.

tomviz/RecentFilesMenu.cxx
```diff
@@ -238,6 +227,14 @@ void RecentFilesMenu::dataSourceTriggered(QAction* actn, bool stack)
   for (pugi::xml_node node = root.child("DataReader"); node;
        node = node.next_sibling("DataReader"), --index) {
     if (index == 0) {
+      if (!QFileInfo::exists(actn->iconText()) && !stack) {
+        // If the user tried to open a recent file that no longer exists, remove
+        // it from the recent files
+        qWarning() << "Error: file '" << actn->iconText() << "' does not exist.";
```
cryos: This would be better presented as a message box, but I think it can be left as a future improvement. This works as expected in my testing, and is an improvement over the previous behavior where items were silently remove.

Issue #1301, opened 2017-10-18
Constraint Based Direct Fourier Method does not return a child dataset


ercius:
The referenced recon method does not create a Reconstruction child data set. It instead updates the original data:
```
        # Set the result as the new scalars.
        utils.set_array(dataset, recon)

        # Mark dataset as volume
        utils.mark_as_volume(dataset)
```
The Weighted Backprojection does return as Reconstruction child dataset:
```
        # Set up the output dataset
        from vtk import vtkImageData
        recon_dataset = vtkImageData()
        recon_dataset.CopyStructure(dataset)
        utils.set_array(recon_dataset, recon)
        utils.mark_as_volume(recon_dataset)

        returnValues = {}
        returnValues["reconstruction"] = recon_dataset
        return returnValues
```

Hovden:
Good catch @ercius . This is also the case for the Constraint-based Direct Fourier
@yijiang1 may not be aware of this updated feature.

mathturtle:
So this is actually the legacy of a UI dispute between Yi and me. :(  If a python operator returns a child dataset, it has to use the new JSON specification for its parameters.  Yi disliked the JSON-generated UI for this reconstruction and wanted to keep the original custom UI that he wrote... so I didn't update this one when I changed all the reconstructions to produce child datasets.  See #764 and the associated PR for some of the discussion (some was at meetings I think, because I remember more than is in the issue).

I can't find the commit I had to fix this last reconstruction operator though.  So I'll have to redo it.

cryos:
Let's track this in this issue, and make sure we come to a conclusion on producing child data sets. If there are blockers we want to get them into issues, and we can decide how/when we want to address them. Thanks for bringing this up, I hadn't gone through all the reconstructions but had thought they were all producing child data sets now.

mathturtle:
After a bit of digging through git's more obscure commands, I found my work-in-progress commit for this operator.  I'm cherry-picking it now and I'll push it as a branch in a few minutes.

cryos:
This has been resolved, and was part of the 1.2.0 release.

PR #1304, opened 2017-10-19
Modify the recon_dft_constraint to use child dataset

mathturtle:
For #1301.


tomviz/AddPythonTransformReaction.cxx
```diff
@@ -358,7 +358,7 @@ OperatorPython* AddPythonTransformReaction::addExpression(DataSource* source)
       addPythonOperator(source, this->scriptLabel, this->scriptSource,
                         arguments);
     }
-  } else if (scriptLabel == "Reconstruct (Constraint-based Direct Fourier)") {
+  }/* else if (scriptLabel == "Reconstruct (Constraint-based Direct Fourier)") {
```
cryos: Why not just remove these lines?

mathturtle: I think this is one of the reasons this commit was still work in progress before I cherry-picked it...  updated.

cryos:
Any thoughts @ercius or @Hovden on whether this is ready to merge.

ercius:
I just downloaded the nightly (1.1.0-121) but #1301is not fixed. Running the Constraint-based Direct Fourier Method reconstruction does not create a child data set.

I think that all reconstruction methods should behave in the same way (i.e. producing a child data set).

cryos:
I want to document the old UI in my opinion does look better, but I think we need to work in improving possible generated layouts. I think the advantage of uniformity is worth the UI cost for now, but would like to work on improving the generated UI. On my system the original UI:
image
The new UI has a very large minimum size, and a lot of empty space:
image

cryos:
Added some line breaks, and will merge one that is a little improved for layout, we should force the reflow of text in future though.
image

PR #1309, opened 2017-10-20
Add more info about setting python variables in building instructions

mathturtle:
@cryos Yi and I figured out what he was doing.  His CMake was finding Anaconda python rather than the python3 installed via homebrew.  So this branch updates the build instructions to say not to use anaconda and also to check which python you found (which Jonathan ran into on OSX too, CMake doesn't find the right thing by default on OSX).

Feel free to reword...


cryos:
Let me reword this a little when I have a little more time, glad you got to the bottom of it.

mathturtle:
I added a commit (to avoid conflicts) that removes some options we no longer need to pass to ParaView as they are defaults in ParaView now (Qt5 and OpenGL2).

Only the python options are non-default ParaView.  The rest of the options are to reduce build time (or use minimal Qt without `qhelpgenerator`).

mathturtle:
+1

Issue #1314, opened 2017-10-26
Missing entries in view menu - orthographic, etc


cryos:
The view menu is missing orthographic/perspective and other entries. A quick grep shows the code is still in there, but this is not showing  up in the menu of the 1.2.0 release.

cryos:
This was fixed by #1316

PR #1316, opened 2017-10-27
View menu fixes

mathturtle:
For #1314.  The ParaView `pqViewMenuManager` changed its behavior and broke our `ViewMenuManager` due to virtual functions called in the constructor being undefined behavior... This restores the menu items and fixes a few other issues with the view menu manager that I found while looking into this.


mathturtle:
@utkarshayachit I figured out what broke.  `buildMenu` is virtual, but you are calling it in the constructor so it doesn't get the subclass's implementation.  You should remove the virtual keyword from it since it is now called in the constructor.

PR #1318, opened 2017-10-27
Paste into tilt angles table

mathturtle:
This adds support for pasting in tilt angles from excel/google sheets/etc.

Currently the operator dialog works fine, but the data properties panel is causing segfaults on the background thread after pasting in tilt angles....

### DIFF
--- a/tomviz/DataPropertiesPanel.cxx
+++ b/tomviz/DataPropertiesPanel.cxx
@@ -42,10 +42,11 @@
 
 #include <QClipboard>
 #include <QDebug>
-#include <QKeyEvent>
-#include <QMimeData>
 #include <QDoubleValidator>
+#include <QKeyEvent>
 #include <QMainWindow>
+#include <QMessageBox>
+#include <QMimeData>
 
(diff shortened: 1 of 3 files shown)


cquammen:
Works nicely and the feedback is great when you don't do something correctly. However, I wonder if it would be better to allow only one row to be selected at a time in the table. That would prevent the need to do something special for multiple selections and non-contiguous selections. Alternatively, in both of these case you could just pick the first index and paste from there and not ascribe any particular meaning to the multiple selections.

ricounet67:
I tried to use these feature with tomviz 1.2.1-62-g0e67e1e with angles values open with Excel 2016 Mac and when I paste the values, I have a message Error: pasted tilt angle is not a number. Do you have any idea of the origin of the problem? If I copy/paste a single value, it works.
Thank

mathturtle:
I'm not sure but I'll look into it.  I'm not sure we have Office for Mac on any of the computers here, but I'll try Office on Windows and see if I can reproduce.

PR #1321, opened 2017-11-01
Use a message box rather than the console error when file not found

mathturtle:


tomviz/RecentFilesMenu.cxx
```diff
@@ -230,7 +230,7 @@ void RecentFilesMenu::dataSourceTriggered(QAction* actn, bool stack)
       if (!QFileInfo::exists(actn->iconText()) && !stack) {
         // If the user tried to open a recent file that no longer exists, remove
         // it from the recent files
-        qWarning() << "Error: file '" << actn->iconText() << "' does not exist.";
+        QMessageBox::warning(nullptr, "Error", QString("The file '%1' does not exist").arg(actn->iconText()));
```
cryos: We should set the parent to ensure it shows up in the correct location over the parent window.

tomviz/RecentFilesMenu.cxx
```diff
@@ -293,7 +293,7 @@ void RecentFilesMenu::stateTriggered()
   } else {
     // // If the user tried to open a recent file that no longer exists, remove
     // it from the recent files
-    qWarning() << "Error: file '" << actn->iconText() << "' does not exist.";
+    QMessageBox::warning(nullptr, "Error", QString("The file '%1' does not exist").arg(actn->iconText()));
```
cryos: We should set the parent to ensure it shows up in the correct location over the parent window.

mathturtle:
Updated

PR #1323, opened 2017-11-03
Remove used variable

cjh1:
I think this was just left over from cut and paste.


cryos:
Do you mean **unused** variable here? If not, I would like to keep the used variable :smiley_cat:

cjh1:
Lets try again!

PR #1330, opened 2017-11-06
Remove an assert that no longer appears necessary

### LINKED ISSUE #1329: Volume module crashes when Separate Color Map option is enabled
Steps to reproduce

* Sample Data menu -> Generate Constant Dataset
* Add a **Volume** module
* Check the **Separate Color Map** box. Tomviz will crash

The problem is a Q_ASSERT in `Module::gradientOpacityMap()`:

`Q_ASSERT(!m_useDetachedColorMap);`

We need better handling of the gradient opacity map when a separate color map option is selected.

cquammen:
The code after the assert appears to handle the case that the assert
was meant to prevent.

Fixes #1329.

### DIFF
--- a/tomviz/Module.cxx
+++ b/tomviz/Module.cxx
@@ -163,7 +163,6 @@ vtkSMProxy* Module::opacityMap() const
 
 vtkPiecewiseFunction* Module::gradientOpacityMap() const
 {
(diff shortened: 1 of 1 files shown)


cquammen:
Thanks!

PR #1340, opened 2017-11-08
Fix setting of separate color map to preset

### LINKED ISSUE #1145: "Separate color map" feature broken for contour
The "separate color map" check box appears to be broken for contour renderings.  to replicate the problem:
1) open sample data (and leave up slice rendering)
2) add contour rendering
3) check "separate color map" and change the color map
When this is done the contour does not change color, but the pre-existing slice will change color!  Also, the colormap histograms change for the dataset and the slice, but if you click away and click back to the contour rendering, it keeps the default colormap.  It looks like changing the colormap impacts all objects except the selected one.

cquammen:
Remove unneeded severing of signal/slot connections between the
CentralWidget and the active Module when setActiveColorMapDataSource()
is invoked. One of the severed connections was responsible for setting
the current color map in the HistogramWidget. Because the separate
color map was never being set in the HistogramWidget, changes to the
color map were always being applied to the shared color map. Leaving
these connections in place fixes the problem.

Fixes #1145.

### DIFF
--- a/tomviz/CentralWidget.cxx
+++ b/tomviz/CentralWidget.cxx
@@ -327,10 +327,6 @@ CentralWidget::~CentralWidget()
 
 void CentralWidget::setActiveColorMapDataSource(DataSource* source)
 {
-  if (m_activeModule) {
-    m_activeModule->disconnect(this);
-    m_activeModule = nullptr;
-  }
   setColorMapDataSource(source);
 }
 


cquammen:
My guess is also copy🍝. For your general knowledge, it is possible to make a connection unique by passing `Qt::UniqueConnection` in as an additional argument at the end of `QObject::connect()`. From [1]:

> By default, a signal is emitted for every connection you make; two signals are emitted for duplicate connections. You can break all of these connections with a single disconnect() call. If you pass the Qt::UniqueConnection type, the connection will only be made if it is not a duplicate. If there is already a duplicate (exact same signal to the exact same slot on the same objects), the connection will fail and connect will return an invalid QMetaObject::Connection.
> 
> Note: Qt::UniqueConnections do not work for lambdas, non-member functions and functors; they only apply to connecting to member functions.

[1] http://doc.qt.io/qt-5/qobject.html

mathturtle:
That's neat.  I didn't know about `Qt::UniqueConnection`.  Thanks for the tip!

Issue #1342, opened 2017-11-09
Link to mailing list on tomviz web page


cquammen:
Should there be a link to the tomviz mailing list [1] on the tomviz.org web page?

[1] https://public.kitware.com/mailman/listinfo/tomviz

ercius:
There's a mailing list?

cquammen:
Yep.

cryos:
I think a link on the web page is more appropriate, I will get something added soon. I want to expand things out a little, and make it clear how people can communicate with us.

cryos:
Added https://www.tomviz.org/discuss/ which I think covers this well, and gives a few viable options.

PR #1343, opened 2017-11-09
Add information about the tomviz mailing list to the README

cquammen:
Partially addresses #1342.


cryos:
I think it belongs on the web site, but don't think it needs to go in the readme. I would rather we moved to something more modern, and so haven't been overly pushing the mailing list, but it should be more visible.

cquammen:
ok

PR #1344, opened 2017-11-10
Modified Volume Downsize Transform for RGB datasets.

jtschwar:
Hi, 
I've been trying to implement RGB datasets into the _BinVolumeByTwo.py_ script and I run into a memory error when trying to put the modified array back into the original data set in utils.py line 137. If anyone would like, I can provide the RGB data set I've been testing by email.

### DIFF
--- a/tomviz/ModuleManager.cxx
+++ b/tomviz/ModuleManager.cxx
@@ -144,6 +144,14 @@ void ModuleManager::removeDataSource(DataSource* dataSource)
   }
 }
 
+void ModuleManager::removeChildDataSource(DataSource* dataSource)
+{
+  if (this->Internals->ChildDataSources.removeOne(dataSource)) {
+    emit this->childDataSourceRemoved(dataSource);
+    dataSource->deleteLater();
+  }
+}
+
 void ModuleManager::removeAllDataSources()
 {
   foreach (DataSource* dataSource, this->Internals->DataSources) {
--- a/tomviz/ModuleManager.cxx
+++ b/tomviz/ModuleManager.cxx
@@ -144,14 +144,6 @@ void ModuleManager::removeDataSource(DataSource* dataSource)
   }
 }
 
-void ModuleManager::removeChildDataSource(DataSource* dataSource)
-{
-  if (this->Internals->ChildDataSources.removeOne(dataSource)) {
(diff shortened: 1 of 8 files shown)


mathturtle:
I can reproduce the error and I get
```
  File "Bin Volume x2", line 32, in transform_scalars
  File "/home/shawn/dev/tomviz/build/lib/site-packages/tomviz/utils.py", line 137, in set_array
    do.PointData.append(arr, arrayname) #Issue
  File "/home/shawn/dev/paraview/build-tomviz/lib/site-packages/vtk/numpy_interface/dataset_adapter.py", line 719, in append
    narray = narray.flatten() * numpy.ones((arrLength, components))
  File "/home/shawn/tools/tomviz-deps/lib/python3.6/site-packages/numpy/core/numeric.py", line 178, in ones
    a = empty(shape, dtype, order)
MemoryError
Failed to execute the script.

```

mathturtle:
@cquammen @cryos Do you have any ideas?

mathturtle:
So I've sort of figured it out with a bit more debugging.

append() expects the numpy array to have shape `(totalNumPoints, numComponents)` and we are passing in `(XDim, YDim, ZDim, numComponents)`.  So it is computing the number of components to be 6291456 and trying to allocate a 2097152 by 6291456 array (thus the out of memory).  You need to change the input into a 2D array where the three spatial dimensions are in fortran order.  Unfortunately I don't know numpy well enough to know how to do that.  If you look at the top of `set_array` where the variable `arr` is created, it looks like it is trying to make a 1D array out of it.  You would want a 2D with dimensions (128^3, 3).  Maybe one of the others can help you with the specific numpy calls you need.

cquammen:
I am traveling today but just wanted to chime in that we have not yet added any support for multicomponent images data transforms or reconstruction algorithms. We will be working toward that goal in the very near term, though.

jtschwar:
@mathturtle You're right that's the issue, I'll spend this week modifying the function so that set_array can support RGB datasets.

cryos:
This is what I mentioned at the FET meeting, our set/get functions will all assume that there is only one component. We can work together to make this more general.

jtschwar:
Hi, 
I updated the branch and now this transform is capable to process RGB datasets.

Issue #912, opened 2017-02-20
Display Camera lighting location


Hovden:
Reporting where the light is coming from would add transparency to understanding the final visualization. It is hard for me to know where the light is coming from—I think it is coming from underneath the object in surface render.

Hovden:
Bump.

It would be useful to move the lighting location.

cquammen:
Some enhancements to ParaView to greatly improve lighting controls have been merged to ParaView master, and plans to show widgets that represent lights are coming.

Hovden:
@jtschwar Please test out the new lighting features. I will as well.

Hovden:
Bumping this issue again. It would help with a figure I'm putting together.

mathturtle:
I have a branch that adds the ParaView "Lights Inspector" panel to tomviz.  It requires a ParaView update and the panel from ParaView is fairly complex.  This adds it sort of like the animation panel: it is there and you can use it if you want, but we will probably want to redo it at some point.

PR #1347, opened 2017-11-15
Enable ParaView lights inspector

mathturtle:
See #912.


cjh1:
Do we need to use a new version of ParaView on CI for these changes?

cryos:
Yes, it will need to be moved forward in the superbuild and the CI unfortunately. It is a shame there isn't a simple way to optionally use it with newer ParaView, it is pretty self-contained and won't be exercised by the CI in any way.

cjh1:
Ok, I can push things forward when the superbuild is updated.

cjh1:
CI build are updated to latest ParaView

PR #1351, opened 2017-11-17
Consider number of image components when listing available Modules

cquammen:
Some modules cannot be applied when the DataSource has multicomponent active scalars. This commit provides a method to query the number of components in the DataSource, and consults it when returning the list of applicable Module types. At the moment, the Contour, Volume, and Threshold Modules are not applicable to multicomponent images.

### DIFF
--- a/tomviz/DataSource.cxx
+++ b/tomviz/DataSource.cxx
@@ -27,6 +27,7 @@
 #include <vtkImageData.h>
 #include <vtkNew.h>
 #include <vtkPiecewiseFunction.h>
+#include <vtkPointData.h>
 #include <vtkSmartPointer.h>
 #include <vtkStringArray.h>
 #include <vtkTrivialProducer.h>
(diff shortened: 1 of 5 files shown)


cquammen:
**Module bar for single-component image data**
image

**Module bar for multicomponent image data**
image

cquammen:
@cryos @Hovden This currently removes the toolbuttons for the Modules that cannot be applied to the current data source. Another option would be to show the toolbuttons for the Modules that cannot be applied but disable them. Which option would you prefer?

jtschwar:
I tested this branch and it works as intended. However, we would prefer that you disable the tool buttons. This would keep the icons from translating left and right if a user were to switch between single and multi-component image data.

cryos:
Forgot about this, agreed on disabling rather than removing modules that can't be used. It seems more intuitive, less likely to confuse.

cquammen:
Agreed, I will change this to disable buttons instead.

cquammen:
Here's the new look:

At startup
image

Single component data set
image

Multicomponent data set
image

cquammen:
AppVeyor failure seems to be a configuration problem rather than a problem with this PR. cc: @cjh1

cquammen:
Thanks for the review!

Issue #1123, opened 2017-03-27
Scale cube features


ElliotPadgett:
I like the scale cube!  It would be even nicer with a few new features.  Here are some that I would like:
1) Set the color
2) Allow the user to type in the position
3) Allow the user to set the side lengths independently so that it is not necessarily a "cube".  This would allow you the flexibility to make something like a platform for the object to sit above.

Hovden:
Setting the color for the scalecube would be really valuable for me right now. I want to bump this issue with emphasis on the first issue. I think we could close this issue if that is addressed (@ElliotPadgett, I think there are ways we can accomplish the other two issues).

mathturtle:
@Hovden It was just pointed out to me that there are two scale cubes in tomviz.  There is the one in the visualization menu that you can move around within the dataset and the one in the view menu that is fixed in the lower right corner.  Which cube were you wanting to use with color right now?  @cryos thinks it is the visualization menu one.  (I'm working on setting color for both but I just figured I'd clarify).

Hovden:
The visualization menu scalecube.

mathturtle:
@cryos Should we close this and open a new issue for the global scale cube not being saved in the state file or use this issue?

PR #1352, opened 2017-11-17
Add option to set the color on the scale cube module

mathturtle:
The color will be saved/loaded as part of the state file too.

If we want to set the color of the one created by the view menu in the lower right, that is more complicated...


mathturtle:
For #1123.

jtschwar:
Works well on MacOS.

Issue #1356, opened 2017-11-27
Python operator with spin boxes fail with TypeError


mathturtle:
Load a dataset, try applying any python operator whose JSON-defined UI contains an integer or double with a spinbox as its UI.  (This was found with DataTransforms -> Gaussian Filter).  The transform fails to run and gets this error:
```
TypeError: transform_scalars() got an unexpected keyword argument 'qt_spinbox_lineedit'
Failed to execute the script.
```

Some initial diagnosis during the call with @jtschwar:
This is related to #1348, commenting out lines 136-139 of OperatorWidget.cxx "fix" it while breaking the file and directory types.  I think it is grabbing the SpinBox's internal LineEdit and trying to use that as if it were a parameter widget we created.  @cjh1

cjh1:
I will take a look.

cquammen:
#1357 appears to fix the problem.

PR #1357, opened 2017-11-27
Tag QLineEdit with the type

cjh1:
Tag the line edit with the type, so we can distinguish it from other line edit uses ( such as in a QSpinBox ).


cjh1:
Fixes #1356

jtschwar:
I tested this branch and it works as intended.

Issue #1181, opened 2017-06-20
Add Color Legend Button for Views


rqus:
Colorbar in the visualization view, which normally works in Paraview is not available in Tomviz. This feature would be extremely useful for optical diffraction tomography / tomographic phase microscopy reconstruction data.

cryos:
I think this is the issue tracking the #1360, it didn't come up when searching for color or legend, I will update the title a little so that they are easier to correlate.

cquammen:
Basic color legend functionality is in place. Please open new issues for requests to modify the appearance of the color legend.

PR #1360, opened 2017-12-05
Show color legend

cquammen:
This introduces a toolbutton at the bottom of the LUT controls in the HistogramWidget that toggles the color legend visibility in the active view.

image

The button controls the visibility of the color legend corresponding to the color map displayed in the HistogramWidget.

Note: State for the color legend (just visibility at the moment) is currently not saved to or reloaded from state files.

It would be great to get some feedback on the usability of this as I work on saving/loading color legend state to/from a state file.

Addresses #1181.

### DIFF
--- a/tomviz/CentralWidget.cxx
+++ b/tomviz/CentralWidget.cxx
@@ -30,12 +30,15 @@
 #include <vtkVector.h>
 
 #include <vtkPVDiscretizableColorTransferFunction.h>
+#include <vtkSMPropertyHelper.h>
+#include <vtkSMViewProxy.h>
 
 #include <QModelIndex>
 #include <QThread>
 #include <QTimer>
 
 #include "AbstractDataModel.h"
+#include "ActiveObjects.h"
 #include "ComputeHistogram.h"
 #include "DataSource.h"
 #include "Module.h"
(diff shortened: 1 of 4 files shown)


cryos:
Is there an issue for this feature?

cquammen:
Yes, I modified the description to reference it, which I should have done initially.

cryos:
I think it looks pretty good, the main gripes I would raise about the default are probably inherent to the scalar bar it is using - the main labels are simple, 20, 40, 60, but it is pretty jarring that the min is 0.0e+00, roughly stated 0, and 2.6e+02, which I think is rounding up from the max of 255. I am not sure if there is a simple way to improve this, or if the switch from numeric representations is intentional for the more general case.

cquammen:
Thanks for the review. It is currently using ParaView defaults for the range labels in scientific notation. We can either a). not show them, b), change the C format for them easily enough. Let me know which you would prefer. I've intentionally avoided any customizations until getting feedback on what folks would like to see in the default color legend.

cryos:
If it is reasonably easy my first suggestion would be to display range, but use the same notation as the other labels. In this case it would offer more accuracy on the max with less characters needed.

cquammen:
Sure, it should be a one-line change. I'll add it to this PR.

cquammen:
With the new range label format, the color legend looks like

image

Hovden:
Looks good.

Issue #543, opened 2016-08-03
Histogram Sig Figs


ercius:
Can we lower the number of significant digits on the histogram y-axis? Maybe we need only one decimal (X.Y or e.g. 1.0e6) at the most?
![clipboard01](https://cloud.githubusercontent.com/assets/15657108/17349904/a364bada-58d6-11e6-8a39-c1b0e0ee6e19.png)

cryos:
I will look into this, it should not be showing that many figures, in fact I set it to show none but it is possible there is a big in the tick generation code on Windows (I am assuming this is Windows - I just checked and don't see this issue on Linux).

ercius:
Yes. It's windows 10.

cquammen:
Behavior confirmed on Windows 10 with latest nightly build.

cryos:
I see this too, and it is a VTK bug in the charting code. It appears the API we are using fails to respect the requested significant figures. This is not intended behavior, we actually tried to make this look simpler/tweak the display of the chart.

Hovden:
Is this still an issue? Things look fine on OSX.

cryos:
I think it is, I will try and spend a little time on Windows digging into this.

Hovden:
Ok. Not a big priority for 1.0

ercius:
Its still an issue on 0.9.3-205-g1293187. But yes, low priority.

cryos:
Sorry, this won't make 1.0, but I will try and get to this as it is irritating on Windows.

ercius:
Thanks for the update. It's only an irritation as you say.

cryos:
Closing this issue, using one significant figure due to bug in MSVC, erring on the side of consistency over platforms.

PR #1365, opened 2017-12-08
Workaround Windows precision bug

cryos:
After investigating this issue, it turns out there is a bug in the MSVC
STL where using a precision of 0 with scientific notation results in the
display of 6 decimal places. I think we should favor consistency on all
platforms, using 1 decimal place for now. Fixes issue #543 (kinda).


cryos:
I think they moved the Qt packages @cjh1

cjh1:
I have raised #1366

PR #1368, opened 2017-12-12
View Code Button 

jtschwar:
Add a view code in the operator properties panel, to the thanks of Shawn Waldon.

@mathturtle

### DIFF
--- a/tomviz/EditOperatorDialog.h
+++ b/tomviz/EditOperatorDialog.h
@@ -39,7 +39,11 @@ public:
                      bool needToAddOperator, QWidget* parent);
   virtual ~EditOperatorDialog();
 
+  void setViewMode(const QString& mode);
+
   Operator* op();
+
+  static void showDialogForOperator(Operator* op, const QString& viewMode = "");
 
 private slots:
   void onApply();
--- a/tomviz/EditOperatorDialog.h
+++ b/tomviz/EditOperatorDialog.h
@@ -41,22 +41,16 @@ public:
 
   // Used to set the mode of the EditOperatorWidget in the dialog.  The mode
   // corresponds to dialog options like tabs and varies from operator to
-  // operator.
-  // If the requested mode is not recognized, or the widget does not support
-  // modes,
-  // this function does nothing.
+  // operator. If the requested mode is not recognized, or the widget does not
+  // support modes, this function does nothing.
   void setViewMode(const QString& mode);
 
   Operator* op();
 
   // If the given operator does not already have a dialog, this function creates
   // and shows a new dialog for that operator with the given mode (see comment
-  // above
-  // on setViewMode for details about modes).
-  //
-  // If the given operator has a dialog already, that dialog is set to the
-  // requested
-  // mode and given focus.
+  // above on setViewMode for details about modes).  If the given operator has a
+  // dialog already, that dialog is set to the requested mode and given focus.
   static void showDialogForOperator(Operator* op, const QString& viewMode = "");
 
 private slots:
--- a/tomviz/EditOperatorDialog.h
+++ b/tomviz/EditOperatorDialog.h
@@ -39,10 +39,24 @@ public:
                      bool needToAddOperator, QWidget* parent);
   virtual ~EditOperatorDialog();
 
+  // Used to set the mode of the EditOperatorWidget in the dialog.  The mode
+  // corresponds to dialog options like tabs and varies from operator to
+  // operator.
+  // If the requested mode is not recognized, or the widget does not support
+  // modes,
+  // this function does nothing.
   void setViewMode(const QString& mode);
 
   Operator* op();
 
+  // If the given operator does not already have a dialog, this function creates
+  // and shows a new dialog for that operator with the given mode (see comment
+  // above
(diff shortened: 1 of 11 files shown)


tomviz/OperatorPropertiesPanel.cxx
```diff
@@ -145,4 +158,38 @@ void OperatorPropertiesPanel::apply()
...
+void OperatorPropertiesPanel::viewCodePressed()
+{
+  if (!m_activeOperator) {
+    return;
+  }
+
+  if (m_activeOperator->hasCustomUI()) {
+    // See if we already have a dialog open for this operator
+    auto dialog =
+      ActiveObjects::instance().getDialogForOperator(m_activeOperator);
```
cryos: I would much rather call m_activeOperator->customDialog() or similar here, rather than use active objects as the operator should maintain that mapping internally (unless I am missing something).

mathturtle:
@jtschwar are you still wanting to look at this with me or should I just finish it?  It needs a rebase and master is kinda unstable with the pipeline changes.

mathturtle:
@cryos Updated based on your feedback above.  I've added comments on our latest round of changes.  This is ready for review now.

tomviz/EditOperatorDialog.h
```diff
@@ -39,8 +39,26 @@ class EditOperatorDialog : public QDialog
                      bool needToAddOperator, QWidget* parent);
   virtual ~EditOperatorDialog();
 
+  // Used to set the mode of the EditOperatorWidget in the dialog.  The mode
+  // corresponds to dialog options like tabs and varies from operator to
+  // operator.
```
cryos: Can we fix up the line wrapping in the comments please. Several others are the same, I am guessing an artifact from clang-format - enter as one long line and let it reflow accordingly, or use a static word wrap marker in your code editor at 80 characters to help.

tomviz/OperatorPropertiesPanel.cxx
```diff
@@ -17,15 +17,19 @@
 
 #include "ActiveObjects.h"
 #include "DataSource.h"
+#include "EditOperatorDialog.h"
 #include "Operator.h"
 #include "OperatorWidget.h"
 #include "Pipeline.h"
 #include "Utilities.h"
 
+#include "pqCoreUtilities.h"
```
cryos: Is this needed? If so it should use <> rather than ""

tomviz/OperatorPython.h
```diff
@@ -32,6 +32,13 @@ class OperatorPython : public Operator
   Q_OBJECT
 
 public:
+  // Pass this constant in to EditOperatorWidget::setViewMode to change the tab
+  // to the Script tab where the python operator's code is viewable.  This only
+  // applies to the custom EditOperatorWidget used by OperatorPython and
+  // returned
+  // from its getEditorContents method.
+  static const QString VIEW_CODE_MODE;
```
cryos: Why are we adding this to the public class interface? Is it accessed elsewhere?

mathturtle: Yes, we are using it in the OperatorPropertiesPanel when the "View Code" button is pressed.

tomviz/EditOperatorDialog.h
mathturtle: I've updated the comments to wrap better.

tomviz/OperatorPropertiesPanel.cxx
mathturtle: *tries taking it out*... apparently not.  We changed to use the new Utilities function (tomviz::mainWidget) from pqCorUtilities::mainWidget when we rebased onto the new pipeline changes and I guess we forgot to get rid of this include.  I've added a commit to remove this.

tomviz/OperatorPython.h
mathturtle: type/constant:  I thought about making it an `enum`, but since the valid values vary from widget to widget, it didn't seem to fit.  Given that an `enum` didn't make sense I went with string over integer since they are more descriptive.  I guess the constant may be overkill, but if some string in two parts of the program have to match I try to make it a constant somewhere rather than relying on it staying the same.

My thoughts on the name: I didn't really like it either... but I could see reusing this to do more than set the default tab (for example change styling based on if we are in a dialog or the properties panel).

cryos: Let's just keep it simple, and do string matching. I think this is a little over-engineered. Signals and slots are all based upon string matching at runtime, with some decoration. We can revisit this when/if it really needs more infrastructure, at that point I am more inclined to consider using JSON objects, and have been experimenting with that elsewhere.

cryos:
Remaining issues that I am aware of:
- Doesn't work on operators with no custom interface, i.e. invert
- Move to entering the string directly in the call to showDialogForOperator, consider [QStringLiteral](http://doc.qt.io/qt-5/qstring.html#QStringLiteral)
- Remove the static QString declaration from OperatorPython

There may be some layout tweaking too, but I think we can look at that once this is merged.

cryos:
Addressed the second two points I made, skipping the first for now.

cryos:
All green, merging.

PR #1370, opened 2017-12-15
WIP: Add active array concept for Modules

cquammen:
Adds a "Color Array" option to modules to enable selection of which data array to view.

image

### DIFF
--- a/tomviz/Module.cxx
+++ b/tomviz/Module.cxx
@@ -240,6 +240,20 @@ bool Module::deserialize(const pugi::xml_node& ns)
   return true;
 }
 
+void Module::setActiveArray(int index, const QString & arrayName)
+{
+  m_activeArrays.insert(index, arrayName);
+}
+
+QString Module::activeArray(int index)
+{
+  if (m_activeArrays.contains(index)) {
+    return m_activeArrays[index];
+  }
+
+  return QString();
+}
+
 void Module::onColorMapChanged()
 {
   emit colorMapChanged();
--- a/tomviz/Module.h
+++ b/tomviz/Module.h
@@ -18,6 +18,7 @@
 
 #include <QIcon>
 #include <QObject>
+#include <QMap>
 #include <QPointer>
 #include <QScopedPointer>
 
@@ -86,6 +87,11 @@ public:
   /// serialize the state of the module.
   virtual bool serialize(pugi::xml_node& ns) const;
   virtual bool deserialize(const pugi::xml_node& ns);
+
+  /// Set the active data array for the given index the array of the given
+  /// arrayName.
+  virtual void setActiveArray(int index, const QString &arrayName);
+  virtual QString activeArray(int index);
 
   /// Modules that use transfer functions should override this method to return
   /// true.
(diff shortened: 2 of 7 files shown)


cquammen:
@cryos Here is a first go at adding color array selection in Modules. So far, only the Orthogonal Slice module responds to the active array. I'm looking for early feedback - the API and UI may change, and I'll need to add serialization and deserialization as well.

cryos:
One thing that jumps out before doing any testing is color array implies something quite specific, but I thought you were working on a more general active array concept where I can choose the active array, for slice that will be used for coloring, but for contour it will be used for the isosurface determination. We also have a concept of color by for label maps, so this might be confusing and we want to maintain that option in the modules where it makes sense.

cquammen:
Thanks for the early feedback! It is very useful. Indeed, I plan to implement an active array concept for data sources. But I thought we discussed holding off on that until @cjh1's changes to the pipeline land in master to avoid merge conflicts.

With regards to the name "Color Array", yes, that is too specific. I will change that to "Active Array". For Slice, this will change the coloring and for Contour this will change the isosurface array. I would like to expose the UI for selecting the active array in each Module - I think that will provide the most flexibility for visualizing different data arrays in different Modules. Let me know if that isn't what you had in mind.

Eventually when the data source-level active array option is available in a follow-up PR, I can add a "global" array option to each Module's "Active Array" combo box. The default active array when a Module is created would be this "global" array. Does that sound good?

cryos:
That sounds great Cory, and exactly mirrors what I thought we had discussed during our Michigan visit. I wasn't suggesting pushing the global piece forward, just generalizing the name in the modules. I am working on pushing the data source refactor forward, and hope to have something ready to merge (but perhaps broken in places) this week.

cquammen:
An associated [VTK bug fix](https://gitlab.kitware.com/vtk/vtk/merge_requests/3716) fixes a complaint about no active scalars that is exposed when loading a VTI file and deselecting the "Scalars" array designated in the file.

cryos:
It looks like it just forwards on the call to ComputeIncrements, why not just call that method directly?

cquammen:
`vtkFlyingEdges3D` is doing the calling in the case I found, and `vtkImageData::ComputeIncrements` is protected.

The higher-level issue is what to do if a reader does not designate an active scalars array, which is possible if we are selectively loading arrays in the reader. There is no built-in way to designate an active scalar array in ParaView. instead, we would need to expose `vtkAssignAttribute` as a filter proxy in tomviz that we can use to designate an active scalar for any loaded dataset that does not have one. This would probably be the fastest way to resolve the reader-not-setting-an-active-scalars-array issue, and would be a fast way to add the active array concept to data sources in tomviz. However, the concept of active scalars has lost favor in VTK, and things are moving towards using `SetInputArrayToProcess`.

I would slightly prefer to move forward with modern VTK usage, but setting an active scalar would likely be the faster implementation path. Later on we could modernize the designation of array. What do you think, @cryos ?

cryos:
I would prefer to ensure we always have an active array assigned in the image data before it even gets passed to the pipeline, and to favor the faster path as we need to demonstrate this approach is preferable over others before investing more time than necessary. We have the opportunity ensure there is always an active array (I am rewriting parts of the load data logic to make it simpler/reduce redundant code paths as part of the data source refactor).

cquammen:
Sounds good. Please let me know if you would like any help setting the active scalar array.

Reviewer note: this [VTK bug fix](https://gitlab.kitware.com/vtk/vtk/merge_requests/3716) is not necessary for this MR to work as long as there is a designated active scalar array.

cquammen:
Closing in favor of #1380.

PR #1372, opened 2017-12-19
Save transfer function 2d

mathturtle:
Add support for saving 2D transfer functions
    
This is still initial support where the color map and opacity map of the
box are assumed to be respectively the color map and gradient opacity map
of the datasource/module since there is currently no way to change them
independently of these and save/load should not break the ability to
edit the color maps (currently they update as you update the other, though
there is no way to edit them directly from the 2D UI.  Save/loading them as
separate entities would break this.).


alvarosan:
Sorry @mathturtle , somehow this slipped my mind.  Looking at it now.

cryos:
The refactor has caused a number of conflicts, I think in light of #1373 it would be good to use this as a starting point. I want to ensure we only save the state for things that are actually being used, so if the user never used the 2D transfer function it shouldn't be initialized. I also think that the 2D transfer function will only ever have utility in the volume module, as with the gradient opacity, and so both should be kept in that module, or at the data source level.

I know we wanted to add support for multiple transfer functions, as we rewrite some of the state file code it would be good to understand what is needed. My impression is that we need to save/restore one or more shapes, along with one of more color transfer tables. It may be useful to sketch some of this out as I think the volume module is perhaps the most demanding visualization module for serialization/deserialization.

mathturtle:
@alvarosan I've been looking at this a bit.  It seems like the values are getting restored correctly and then (in my debug build) hitting an assertion in `vtkMath::ClampValue` at vtkTransferFunctionBoxItem.cxx line 456.  I'm guessing that in your build it is simply clamping to 0 silently (which is why you see the box stuck at the origin).  It looks like when tomviz is trying to restore the box, `GetValidBounds` is still returning uninitialized bounds and so it fails to restore the box... Where do those bounds come from?  the histogram?

alvarosan:
That is probably what is happening @mathturtle.  As far as I recall, the implementation of `vtkControlPointsItem::GetValidBounds` returns the bounds of the chart that contains it (in this case the histogram).  So the issue could be that the range of the histogram chart has not been set before the box parameters are loaded.  If the histogram is not being saved (which I think is currently the case), then the box should only be loaded once the histogram has been recomputed, or at least refresh the box position/dimensions.

cjh1:
@mathturtle Are you still looking at this? If not I am happy to take a look.

mathturtle:
This was a prototype of how saving the 2d transfer functions would work under the old state file format.  It still needs porting to the new state file and some restructuring (@cryos wanted a single transferfunction2d object, rather than a list of boxes).

No, I'm not looking at this right now so if you want you can take it over.

PR #1374, opened 2018-01-02
Fix various CI issues

cjh1:


cjh1:
@cryos This should clear up our backlog of CI issues.

Issue #1377, opened 2018-01-04
 Locate default modules such that they show pipeline output data


cjh1:

cjh1:
Sub issue of #1322

cjh1:
Fixed by #1378

PR #1378, opened 2018-01-09
Ensure default modules are moved to "Output" data source

cjh1:


cjh1:
Fixes #1377

cryos:
I think we will need to tweak this more in the future, but I think it moves us in the right direction.

PR #1380, opened 2018-01-10
Set active scalars in data sources

cquammen:
This branch adds initial support for selecting the "active" scalars in a data set. The active scalars are displayed in the histogram, used for coloring in different modules, and modified by transforms.

To select the active array, click on the data set. The list of arrays will appear as in the figure below:

image

Click on the row in the table of arrays to set that array as active. In this example, the array 'coordsX' is active.

We do not need to keep the selection behavior in the UI - opinions on ways to set the active array are welcome (combo box with array names, for instance?).

### DIFF
--- a/tomviz/DataSource.cxx
+++ b/tomviz/DataSource.cxx
@@ -497,14 +497,14 @@ void DataSource::setSpacing(const double spacing[3])
   emit dataPropertiesChanged();
 }
 
-void DataSource::setActiveScalars(const char* arrayName)
+void DataSource::setActiveScalars(const QString& arrayName)
 {
   vtkAlgorithm* alg = algorithm();
   if (alg) {
     vtkImageData* data =
       vtkImageData::SafeDownCast(alg->GetOutputDataObject(0));
     if (data) {
-      data->GetPointData()->SetActiveScalars(arrayName);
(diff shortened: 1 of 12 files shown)


tomviz/DataSource.cxx
```diff
@@ -497,6 +497,40 @@ void DataSource::setSpacing(const double spacing[3])
   emit dataPropertiesChanged();
 }
 
+void DataSource::setActiveScalars(const char* arrayName)
```
cryos: I feel like the API would fit better if it used QString, and then created the const char* for the VTK function.

tomviz/DataSource.cxx
```diff
@@ -497,6 +497,40 @@ void DataSource::setSpacing(const double spacing[3])
...
+    }
+  }
+
+  dataModified();
+
+  emit activeScalarsChanged();
+  emit dataPropertiesChanged();
+}
+
+const char* DataSource::activeScalars() const
```
cryos: Same here on the return type

tomviz/DataSource.cxx
cquammen: Agreed. Changed in latest revision.

cquammen:
Here's a test data file: https://data.kitware.com/#item/5a57b72d8d777f5e872f8380

PR #1384, opened 2018-01-11
Enable importing of Python operators via menu item

### LINKED ISSUE #1283: Add "install custom script" dialog
Allow the user to browse the file system for a custom script and then copy it to the correct location.

cquammen:
tomviz now always shows the "Custom Transforms" menu, which now features a new menu item "Import Custom Transform". When selected, this item brings up a file dialog one can use to select a Python (*.py) file to import a custom transform. The located Python file will be copied to <home>/tomviz. If a .json file with the same base file name is located in the same directory as the selected Python file, it will also be copied to the <home>/tomviz directory.
    
Imported custom transforms appear immediately in the "Custom Transforms" menu.

Fixes #1283.

### DIFF
--- a/tomviz/MainWindow.cxx
+++ b/tomviz/MainWindow.cxx
@@ -684,10 +684,11 @@ void MainWindow::registerCustomOperators()
 void MainWindow::registerCustomOperators(const QString& path)
 {
   std::vector<OperatorDescription> operators = findCustomOperators(path);
+
   // Sort so we get a consistent order each time we load
   std::sort(operators.begin(), operators.end(),
             [](const OperatorDescription& op1, const OperatorDescription& op2) {
-              return op1.label.compare(op2.label);
+              return  op1.label < op2.label;
             });
 
   if (!operators.empty()) {
--- a/tomviz/MainWindow.cxx
+++ b/tomviz/MainWindow.cxx
@@ -67,6 +67,7 @@
 #include <QDebug>
 #include <QDesktopServices>
 #include <QDir>
+#include <QFileDialog>
 #include <QFileInfo>
 #include <QIcon>
 #include <QMessageBox>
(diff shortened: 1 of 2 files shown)


tomviz/MainWindow.h
```diff
@@ -81,6 +86,7 @@ private slots:
   void registerCustomOperators();
 
   QScopedPointer<Ui::MainWindow> m_ui;
+  QMenu* m_customTransformsMenu;
```
cryos: I think we should prefer inline initialization to nullptr as below over doing so in the contructor

tomviz/MainWindow.cxx
```diff
@@ -684,22 +768,20 @@ void MainWindow::registerCustomOperators()
 void MainWindow::registerCustomOperators(const QString& path)
 {
   std::vector<OperatorDescription> operators = findCustomOperators(path);
+
   // Sort so we get a consistent order each time we load
   std::sort(operators.begin(), operators.end(),
             [](const OperatorDescription& op1, const OperatorDescription& op2) {
-              return op1.label.compare(op2.label);
+              return  op1.label < op2.label;
```
cjh1: Any reason for this change?

cquammen: Yep, explained in the first commit. Seems like `compare` should have worked, but it was always returning positive integers, so the sort result was the same even when swapping operands `op1` and `op2`. Plus, `std::sort` takes a comparison function that returns a `bool` rather than an `int` [1].

[1] http://en.cppreference.com/w/cpp/algorithm/sort

tomviz/MainWindow.h
cquammen: Sounds good. Still not in the swing of C++11...

cquammen: Fixed in latest revision.

tomviz/MainWindow.cxx
cjh1: Ok, thanks for the explanation. I have learn something :smile:

Issue #1322, opened 2017-11-02
Separate out "original" and "transformed" data sources


cjh1:
#1236

cjh1:
Here is an example of the current progress:

![pipe](https://user-images.githubusercontent.com/1031053/32345342-428491c8-bfe0-11e7-88f7-9af7642cbd2a.png)

I don't think the "[Transformed]" postfix works ...

cryos:
No, but it does summarize what I was describing well, and is described as transformed in the current data source dialog.

cjh1:
Here is another example, closer to the situation we were discussing with @Hovden in the meeting yesterday:

![example](https://user-images.githubusercontent.com/1031053/32391946-36831d70-c0aa-11e7-96ad-087a918b4d47.png)

cryos:
I intend to merge #1364 soon as this branch is already quite large, and spans quite some time. I think the general direction is good, and am happy to have sorted out many of the issues with the running pipeline. Major issues that need to be tracked here (and possibly broken out into a few follow up issues with trackers) are:

- [ ] Ensure serialization/deserialization fully support data sources, operators, and modules in the updated pipeline, supporting restoring saved copies of child data/reconstructions, etc
- [ ] Logic in the pipeline for removing operators, and keeping the "output" in the right place
- [ ] Default to appending operators to linear chain rather than creating more branches
- [ ] Develop appropriate pipeline execution animation to show pipeline as it runs
- [ ] Locate default modules such that they show pipeline output data

Serialization/deserialization needs to be rewritten, I have been experimenting with JSON, and in light of significant changes to the DataSource, which I think set us up well for the advanced pipeline we need, I am not sure there is value in porting the existing code. I started developing on this branch, but think serialization/deserialization would be better completed in a follow up branch. I see some good paths, but could use input on the pipeline, operators, ensuring we keep track of child data sets.

cjh1:
#1378

PR #1393, opened 2018-01-19
Add the operator and move the transformed data as one step

mathturtle:
This avoids the pipeline view flashing briefly to the state where the
new operator is added but the transformed data is still a child of the
previous operator in the pipeline.

### DIFF
--- a/tomviz/Pipeline.h
+++ b/tomviz/Pipeline.h
@@ -81,6 +81,11 @@ signals:
   /// This signal is fired the execution of the pipeline finishes.
   void finished();
 
+  /// This signal is fired when an operator is added.  The second argument
+  /// is the datasource that should be moved to become its output in the pipeline
+  /// view (or null if there isn't one).
+  void operatorAdded(Operator* op, DataSource* outputDS = nullptr);
+
 private:
   DataSource* findTransformedDataSource(DataSource* dataSource);
   Operator* findTransformedDataSourceOperator(DataSource* dataSource);
--- a/tomviz/PipelineModel.cxx
+++ b/tomviz/PipelineModel.cxx
@@ -21,6 +21,7 @@
 #include "ModuleManager.h"
 #include "Operator.h"
 #include "OperatorResult.h"
+#include "Pipeline.h"
 
 #include <QFileInfo>
 #include <QFont>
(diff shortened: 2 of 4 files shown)


cjh1:
@mathturtle Excellent, this works well.

tomviz/PipelineModel.cxx
```diff
@@ -769,6 +771,22 @@ void PipelineModel::operatorAdded(Operator* op)
...
+    auto transformedIndex = this->dataSourceIndex(transformedDataSource);
+    auto transformedDataSourceItem = this->treeItem(transformedIndex);
+    auto oldParent = transformedDataSourceItem->parent()->op();
+    auto oldParentIndex = this->operatorIndex(oldParent);
+
+    beginMoveRows(oldParentIndex, transformedIndex.row(), transformedIndex.row(), operatorIndex,
+                  operatorTreeItem->childCount());
+    transformedDataSourceItem = transformedDataSourceItem->detach();
+    operatorTreeItem->attach(transformedDataSourceItem);
+    endMoveRows();
```
cjh1: This is a duplicate of the dataSourceMoved(...) method, can we refactor so we only have it in one place?

tomviz/PipelineModel.cxx
```diff
@@ -685,8 +686,9 @@ void PipelineModel::dataSourceAdded(DataSource* dataSource)
   beginInsertRows(QModelIndex(), 0, 0);
   m_treeItems.append(treeItem);
   endInsertRows();
-  connect(dataSource, SIGNAL(operatorAdded(Operator*)),
-          SLOT(operatorAdded(Operator*)));
+  dataSource->pipeline()->disconnect(this);
```
cjh1: I think you can avoid this ```disconnect(...)``` by using ```Qt::UniqueConnection``` when connecting the signal, it will only make one connection.

tomviz/Pipeline.h
```diff
@@ -81,6 +81,11 @@ protected slots:
   /// This signal is fired the execution of the pipeline finishes.
   void finished();
 
+  /// This signal is fired when an operator is added.  The second argument
+  /// is the datasource that should be moved to become its output in the pipeline
+  /// view (or null if there isn't one).
+  void operatorAdded(Operator* op, DataSource* outputDS);
```
cjh1: Can we just default outputDS to ```nullptr```, then we don't have to pass ```nullptr``` at 273 and 276.

cjh1:
#1322

tomviz/Pipeline.h
mathturtle: I wasn't sure since it is a signal, but apparently it works.

mathturtle:
Updated

tomviz/Pipeline.h
cryos: Signals are just member functions, and so have all the same features available as member functions in terms of default arguments, etc.

Issue #1403, opened 2018-02-02
Compiling Tomviz - error adding symbols: DSO missing from command line


SirSharpest:
Hi,

I followed the build instructions, compiled ITK and Paraview myself and yes they are build with the same python librarys, I checked the 3 variables mentioend by cmake. 

I'm on Arch Linux ```gcc (GCC) 7.2.1 20180116```, if there's any specific software versions that would be helpful, or output that could help please say and I'll supply it too. 

```
Scanning dependencies of target tomviz
[ 96%] Building CXX object tomviz/CMakeFiles/tomviz.dir/AboutDialog.cxx.o
[ 97%] Building CXX object tomviz/CMakeFiles/tomviz.dir/main.cxx.o
[ 98%] Building CXX object tomviz/CMakeFiles/tomviz.dir/WelcomeDialog.cxx.o
[ 98%] Building CXX object tomviz/CMakeFiles/tomviz.dir/MainWindow.cxx.o
[ 99%] Building CXX object tomviz/CMakeFiles/tomviz.dir/tomviz_autogen/EWIEGA46WW/qrc_resources.cpp.o
[100%] Building CXX object tomviz/CMakeFiles/tomviz.dir/tomviz_autogen/mocs_compilation.cpp.o
[100%] Linking CXX shared module ../../lib/site-packages/tomviz/_wrapping.so
[100%] Linking CXX executable ../bin/tomviz
/usr/bin/ld: libtomviz.a(PythonUtilities.cxx.o): undefined reference to symbol 'PyDict_SetItem'
/usr/lib/libpython3.6m.so.1.0: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
make[2]: *** [tomviz/CMakeFiles/tomviz.dir/build.make:340: bin/tomviz] Error 1
make[1]: *** [CMakeFiles/Makefile2:115: tomviz/CMakeFiles/tomviz.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
[100%] Built target _wrapping
make: *** [Makefile:130: all] Error 2
```

SirSharpest:
From what I've read, this sounds like a library might be missing in the cmake config? I'm unsure, if anyone more familiar with function could lend a suggestion I'm more than happy to work on it.

cryos:
[...]

PR #1407, opened 2018-02-05
Ensure Tomviz uses the Python from ParaView

cryos:
This used to work as the find_package(ParaView) call would populate our
cache. This is no longer done in master, which I think is a step
forward, but it means that we must load their cache and populate our
own cache with matching entries.


cryos:
This should help with #1403.

PR #1411, opened 2018-02-07
WIP: Update contour coloring

cquammen:
This PR reworks how coloring in the Contour module is done. The main changes are:

* Only arrays available in the data set are available for coloring the contour; child datasets do not appear in the "Color By" combobox.
* Controls for setting the color map range limited to the color data array range have been added.

The modified UI looks like:

image

Additionally, an example of a segmentation transform that has been modified to add a data array to the data set rather than produce a child data set has been included.

Items remaining:

* Restore categorical color interpolation for label map data arrays. This is ultimately dependent on how we mark additional arrays as label maps. For now, I am appending "_LM" to the end of label map array names. There is sure to be a better way (maybe a map from array name to array type in `DataSource`?).
* Fix crash when BinaryThreshold is applied while a Contour visualization exists.
* Perhaps update the histogram view data range to be that of the "Color By" data array range rather than the active array in the data source.

### DIFF
--- a/tomviz/CentralWidget.cxx
+++ b/tomviz/CentralWidget.cxx
@@ -344,6 +344,8 @@ void CentralWidget::setActiveModule(Module* module)
   if (m_activeModule) {
     connect(m_activeModule, SIGNAL(colorMapChanged()),
             SLOT(onColorMapDataSourceChanged()));
+    connect(m_ui->histogramWidget, SIGNAL(colorMapUpdated()),
+            m_activeModule, SIGNAL(colorMapChanged()));
     setColorMapDataSource(module->colorMapDataSource());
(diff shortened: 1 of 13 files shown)


cquammen:
@cryos @Hovden It would be great to get some early feedback as this set of changes to the Contour visualization is being developed.

cjh1:
@cquammen @cryos Is this PR still relevant? If so may be @psavery could help push it over the finish line?

psavery:
@cryos We removed the contour resampling (for now) since it was slowing down the contour. Is the "Color By" still valid? It was removed along with the resampling.

cryos:
Color By is absolutely wanted, but I think we are likely better off starting again as several things have moved around since this was developed. I am closing the PR, we can talk more but it won't make the next release.

PR #1418, opened 2018-02-16
Always show the view code button for Python operators

mathturtle:
Previously it was only shown if the operator had parameters.


jtschwar:
It works! Good job.

tomviz/OperatorPropertiesPanel.cxx
```diff
@@ -84,29 +91,21 @@ void OperatorPropertiesPanel::setOperator(OperatorPython* op)
   if (!m_operatorWidget->layout() || m_operatorWidget->layout()->count() == 0) {
     m_operatorWidget->deleteLater();
     m_operatorWidget = nullptr;
-    return;
-  }
+  } else {
+    // For now add to scroll box, out operator widget tend to be a little
```
cryos: I think you mean 'our' in the comment above

PR #1419, opened 2018-02-16
Providing an additional tutorial in the help menu. 

jtschwar:
Hi, 
I added a new pdf tutorial from the Microscopy Today article (https://www.cambridge.org/core/journals/microscopy-today/article/tutorial-on-the-visualization-of-volumetric-data-using-tomviz/55B58F40A16E96CDEB644202D9FD08BB) in the help menu.


bdalevin:
I suggest just providing a web link to the article at the Microscopy Today website. 
If we embed a pdf (or an alternative) into tomviz, we may run into journal copyright issues. 

Barnaby

cryos:
That is a good point, I had thought initially that this was open access, and CC-BY or similar. Looking more deeply it is copyright of the journal but seems to offer access to the article/PDF for free. Taking a quick look at their terms I do not think that we can incorporate the PDF, and the most we can do is link to the article.

cryos:
Closing for now, we can certainly add a link to the menu, and I will get this updated on the website soon. We can open a new pull request with the link added.

PR #1427, opened 2018-02-28
Fix manual align operator

cjh1:


tomviz/DataSource.cxx
```diff
@@ -866,14 +858,10 @@ QVector<double> DataSource::getTiltAngles() const
...
-  vtkAlgorithm* tp = algorithm();
-  vtkDataObject* data = tp->GetOutputDataObject(0);
-  vtkFieldData* fd = data->GetFieldData();
-  if (fd->GetArray("tilt_angles") != this->Internals->TiltAngles) {
-    tiltAngles = fd->GetArray("tilt_angles");
-  }
-  if (tiltAngles) {
+  auto data = this->dataObject();
+  auto fd = data->GetFieldData();
+  if (fd->GetArray("tilt_angles")) {
```
mathturtle: Should this use HasArray?  Or store the result and check if tiltAngles is non-null to avoid calling it twice?

tomviz/DataSource.cxx
```diff
@@ -79,23 +78,15 @@ class DataSource::DSInternals
...
       array->FillComponent(0, 0.0);
-      if (!fd->HasArray("tilt_angles")) {
-        fd->AddArray(array.GetPointer());
-      }
-      this->TiltAngles = array.Get();
-    } else {
-      if (!fd->HasArray("tilt_angles")) {
-        fd->AddArray(this->TiltAngles);
-      }
+      fd->AddArray(array.GetPointer());
```
mathturtle: You don't need GetPointer anymore, vtkNew was updated to do the auto conversion like vtkSmartPointer.

tomviz/DataSource.cxx
cjh1: I agree, ```HasArray(...)``` is probably better here.

tomviz/DataSource.cxx
cjh1: Ok, will update

PR #1429, opened 2018-02-28
Install python from brew on Mac to get up to data openssl support

cjh1:


cjh1:
@cryos Could you take a look at this for me?

cryos:
Didn't see this until after it was merged - looks reasonable.

PR #1431, opened 2018-03-02
Add checkmarks in "Show legend" menu items

### LINKED ISSUE #1146: "Show legend as cube" menu item crashes tomviz
I don't know what this thing is supposed to do, but when I click it the program crashes.

cquammen:
Check the menu item for the currently displayed legend type. When
hidden, both menu items are unchecked.

Closes #1146.


cquammen:
Thanks!

> I wonder whether in the future, these setting should be persistent?

Perhaps. Though where to draw the persistence state from is tricky if you have different legend settings in different views.

PR #1434, opened 2018-03-08
WIP: Acquisition UI refactor

cjh1:


cjh1:
So these build errors seem to be related to CMake bug related to automoc and target objects. If I use 3.8.0 things seem to work, however, I don't think we want to push our min requirements that. I may just remove the use of target objects.

cryos:
Like I said, I was very excited about CMake's object library support until I started trying to use it... I think we can keep things simple for now, and look at some of the new improvements being made in latest CMake later when we have bigger reasons to bump the minimum CMake requirement.

tomviz/acquisition/ConnectionsWidget.cxx
```diff
@@ -0,0 +1,177 @@
...
+#include <QMenu>
+
+#include <algorithm>
+
+namespace tomviz {
+
+ConnectionsWidget::ConnectionsWidget(QWidget* parent)
+  : QWidget(parent), m_ui(new Ui::ConnectionsWidget)
+{
+  this->m_ui->setupUi(this);
```
cryos: this-> in a few places

tomviz/acquisition/PassiveAcquisitionWidget.cxx
```diff
@@ -0,0 +1,450 @@
...
+
+void PassiveAcquisitionWidget::onError(const QString& errorMessage,
+                                       const QJsonValue& errorData)
+{
+  auto message = errorMessage;
+  if (!errorData.toString().isEmpty()) {
+    message = QString("%1\n%2").arg(message).arg(errorData.toString());
+  }
+
+  this->stopWatching();
```
cryos: this->

cjh1:
> For me, the dialog feels intimidating and doesn't have simple defaults. 

I can definitely provide some defaults, just wasn't really sure what they should be.

> I would like to see a proposed watch path, I wonder if we just populate it with user home directory/data. 

We could use the user's home directory, but this is likely to get a little confusing if you are connected to a remote machine?

> Then what happens with no regex?

It will just pick up any file in the directory.

> Can we get a drop down with a couple of simple regex patters in, or populate it with a nice default where 
it is obvious how to extract angles?

I can look at the filename format we use in Michigan.

> I would just use Python Path: in the dialog, and the most important part is obscured with ellipses for me. 

Yes, not sure how best to display the path.

Setting it to python3 resulted in a connection refused for me.

cryos:
@cjh1 There is a conflict, but please merge once you have that resolved (not sure what else touched AcquisitionWidget.cxx).

cryos:
It looks like something went wrong with the conflict resolution, there are compiler errors.

PR #1437, opened 2018-03-13
Intro To Visualization Article

jtschwar:
Hi, 
I updated my previous pull request so the button in the help menu opens the link to the article instead of the actual document.


cryos:
It still cannot be merged as it has the copyrighted PDF in it, and would add this to our history. I will get back to my hotel later tonight and can edit this branch, but we must remove the PDF entirely. Leave it with me.

jtschwar:
I removed the pdf in this branch and now the button opens a url for the Microscopy today journal.

cjh1:
@jtschwar The problem is the PDF is still in the branch history ( and therefore end up in our repository ). @cryos will rewrite the history to remove it.

jtschwar:
I see. I didn't know that was also an issue. Thank you for the clarification @cjh1. 
I can also push a new branch without the previous history if need be.

cjh1:
@jtschwar I just merged in latest master to fix a AppVeyor issue.

cjh1:
Ok bots are happy, merging ...

Issue #1176, opened 2017-05-31
"processing tilt image" no. misleading?


CSchank:
When performing an auto alignment, the bottom-left displays "Processing tilt image No.180/180" for example, but since it did the alignment out of order and seems to display the current image you're working on, it could be confused with displaying how many images have been processed. In the case of 180/180, this could be interpreted as almost being done (on last image), only to jump back to 90/180.

cryos:
We should figure out what we want to do here, I think it is more informative to know how far through it is, than give the number of the image in the stack as most of us assume this is a measure of progress.

CSchank:
Yes, those were precisely my initial reactions / thoughts.

PR #1439, opened 2018-03-15
Track the progress of the Auto Alignment python transforms.

jtschwar:
Addressing #1176 to track the tilt image in the bottom left display.

### DIFF
--- a/tomviz/python/AutoTiltAxisShiftAlignment.py
+++ b/tomviz/python/AutoTiltAxisShiftAlignment.py
@@ -32,7 +32,8 @@ class AutoTiltAxisShiftAlignmentOperator(tomviz.operators.CancelableOperator):
         I = np.zeros(shifts.size)
 
         self.progress.maximum = shifts.size - 1
-        step = 0
+        step = 1
+        total = shifts.size-1
 
         for i in range(shifts.size):
             if self.canceled:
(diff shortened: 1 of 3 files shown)


tomviz/python/AutoTiltAxisShiftAlignment.py
```diff
@@ -32,17 +32,18 @@ def transform_scalars(self, dataset):
...
             if self.canceled:
                 return
             shiftedTiltSeries = np.roll(
                 tiltSeries[slices, :, :, ], shifts[i], axis=1)
             for s in range(numberOfSlices):
-                self.progress.message = ('Reconstructing slice No.%d with %d '
+                self.progress.message = ('Reconstructing slice No.%d/%d with %d'
                                          'pixels shift' %
-                                         (slices[s], shifts[i]))
+                                         (step, total, shifts[i]))
```
mathturtle: This one seems to have changed.  The old behavior printed out something like this:
```
"Reconstructing slice No.97 with 20 pixels shift"
"Reconstructing slice No.1 with 20 pixels shift"
"Reconstructing slice No.19 with 20 pixels shift"
"Reconstructing slice No.95 with 20 pixels shift"
"Reconstructing slice No.23 with 20 pixels shift"
```
For each shift from -20 to 20.

The new one prints out
```
"Reconstructing slice No.1/40 with -20 pixels shift"
"Reconstructing slice No.1/40 with -20 pixels shift"
"Reconstructing slice No.1/40 with -20 pixels shift"
"Reconstructing slice No.1/40 with -20 pixels shift"
"Reconstructing slice No.1/40 with -20 pixels shift"
```
So we seem to have lost information.  Maybe it should just print out once, something like `Reconstructing slices for shift X`.

mathturtle: Background: I changed the progress handler to print out the progress messages so I could see them in order.

mathturtle: @jtschwar how much detail do we want to give in the progress messages.  If there is too much, then noone will be able to read it since it will flash by too quick.  Would it make more sense to back off to 'Reconstructing slices for shift -20' or do we think the user cares about seeing information on every reconstruction?

mathturtle:
@jtschwar I've fixed the messages in the Auto Shift Alignment to do what I think you wanted them to.  I still worry about the progress messages being so fine-grained that they will be unreadable though.  I think it should back off to 'Reconstructing slices to test shift -20' or something like that.

jtschwar:
@mathturtle Thank you for fixing Auto Shfit Alignment, it works as intended now. I wanted to address the issue and properly track the calculated image, but for that script we could remove it and keep it general for the amount of 'pixel shift' at that instance.

Issue #1444, opened 2018-03-19
Pipeline layout issue


cjh1:
I am seeing the following layout issue if I add more than a single operator: 

![screen](https://user-images.githubusercontent.com/1031053/37617027-35d0c53c-2b88-11e8-9795-6b3b1ce7e773.png)
It seems like the tables columns widths are being truncated.

cjh1:
Can anyone confirm they are also seeing this?

cryos:
I mentioned this, but thought it was a local issue at the time. I think something in the live Python pipeline update branch broke it - it worked before that merged for me, but then scrunched on adding operators after.

cjh1:
@cryos I have done a bit of bisecting and as far as I can tell its unrelated to the live update changes. I will keep digging.

cjh1:
This appears to be a Qt bug. The layout is fine with 5.9.3.

cjh1:
I have a work around that I will push.

mathturtle:
I know ParaView backed off from Qt 5.10 due to some bugs (but I think they were mostly macOS specific).

cquammen:
Yep, see https://gitlab.kitware.com/paraview/paraview/issues/18013 for some of the problems in 5.10.

cryos:
Yeah I worked with Utkarsh a little on narrowing them down, but they look quite unrelated to the pipeline layout issue. Glad it looks like you might have a workaround for this, I guess they broke a few things in the Qt widgets module.

cjh1:
Fixed by #1445

PR #1445, opened 2018-03-21
Work around for bug in Qt 5.10

cjh1:
For some reason the selection resize mode is getting reset, which messes with the layout. So for now reset this when an operator is added. This can be removed if the underlying bug is resolved in a future version.


cjh1:
Fixes #1444

PR #1446, opened 2018-03-21
Filter duplicates and limit max number of recent files entries

mathturtle:
@cryos The logic in pushStateFile is still using the old XML-based recent files menu state... but state files don't work right now anyway... so we'll just need to open an issue?


cryos:
I am on the state file rewrite, and will take care of it in that issue.

cryos:
We can let the bots do their thing, and assuming they come back green please merge this PR.

cjh1:
@mathturtle I am going to merge this, the mac os Travis timeout is unrelated, I am working on it.

PR #1448, opened 2018-03-21
Add more structure to tomviz/* directory

### LINKED ISSUE #1255: Add more structure to the tomviz/* directory
As we start to get more file its getting a little harder to navigate. We seem to be using file prefixes like ModuleXXX and OperatorXXX to perform grouping. What about using directories instead so having a tomviz/modules/, tomviz/operators/ etc.

cjh1:
Fixed #1255

This PR separates out the module and operator related files into separate directories. There are probably some utility files that should be moved to as the current move is based on the filenames. We can probably go far, thoughts?


cjh1:
Ok, when things have quieted down a little in the repo. I will fix the conflicts

cryos:
:+1: once we get these branches merged

Issue #1450, opened 2018-03-22
Tomography Menu Not Accessible: Mark Data as Tilt Series


Hovden:
Mark data as a tilt series does not allow processes from the tomography menu to be run.

<img width="987" alt="screen shot 2018-03-22 at 1 15 53 pm" src="https://user-images.githubusercontent.com/6197997/37786816-f5460a38-2dd3-11e8-9807-083ab5974782.png">

cjh1:
@Hovden It looks to me like you have deleted the "Output" data source created by the "Set Tilt Angles" operator? That is the data source that would have the tomography menu enabled for ( as it has the tilt  angles ). So I think the problem here is more that you where able to delete the "Output" data source, we should probably not allow that.

mathturtle:
@cjh1 I think that is a good point: we should prevent the Output from being deleted.

But consider this:  If I have the Set Tilt Angles selected and add a square root to the pipeline it goes on the end.  But with that same selection I can't add a Image Alignment operator even though it would be legal to put it on the end (as the output is a tilt series).  This is inconsistent.

cjh1:
@mathturtle I will push a PR to prevent deletion, I agree the use case you detail should also be fixed up.

cjh1:
Related #1454

cjh1:
Related #1453

cjh1:
Yep!

PR #1452, opened 2018-03-23
Disable deletion of "Output" data sources

cjh1:


cjh1:
Related to #1450

tomviz/PipelineView.cxx
```diff
@@ -303,6 +303,14 @@ void PipelineView::contextMenuEvent(QContextMenuEvent* e)
...
   }
 
+  // Disable deletion for "Output" data sources
+  if (dataSource != nullptr) {
+    auto output = dataSource->property("output");
+    if (output.isValid() && output.toBool()) {
+      deleteAction->setEnabled(false);
+    }
+  }
+
```
mathturtle: Shouldn't you just modify the enableDeleteItems() function?

cjh1: Probably, thanks for the tip :smile_cat:

cjh1:
@mathturtle That should be better

PR #1462, opened 2018-04-02
Serialize/Deserialize module colorMap, opacityMap et al

cjh1:


cryos:
This is WIP because of issues crashing? I didn't do the gradient opacity map at the data source level, but this might cause conflicts with your other branch so maybe we can do it in a follow up branch/rebase this one once it is merged.

cjh1:
This crashes when trying to restore a 2d transfer function.

PR #1464, opened 2018-04-02
Flow units changes to child data source

### LINKED ISSUE #1461: Pixel size does not propagate from the input (parent) to the output (child).

cjh1:
Fixed #1461


cjh1:
@ercius This should resolve the issue you came across ( #1461 )

Issue #1475, opened 2018-04-05
Contour transformed data restores contour value to properties panel but not the view


mathturtle:
Steps to reproduce:

* Load a dataset
* Apply a transform (invert, square root or something else simple)
* Add a contour module and change the contour value to something other than the default
* Save State
* Reset/restart tomviz
* Load the state: the properties panel will have the correct value to contour by, but the displayed contour will be the default value

cjh1:
Fixed by #1479

PR #1479, opened 2018-04-06
Fix issue #1475 - update contour value

cryos:
This was just a typo in the proxy that was udpated.


cjh1:
@cryos Sorry looks like that was me :frowning:

Issue #1476, opened 2018-04-05
Relative paths from state files to data files is broken


mathturtle:
* Save a state file next to the data it loads
* Move both the data and state file (or copy to different machine)
* Load the state file... tomviz crashes.

Related to #72.

cjh1:
Fixed by #1480

PR #1480, opened 2018-04-09
Save data source file names relative to state file

cjh1:


cjh1:
Fixes #1476

tomviz/ModuleManager.cxx
```diff
@@ -665,8 +676,17 @@ void ModuleManager::onPVStateLoaded(vtkPVXMLElement*,
...
       }
       QString fileName = dsObject["fileName"].toString();
+      // If we have a fileName make it absolute
+      if (!fileName.isEmpty()) {
+        fileName = d->dir.absoluteFilePath(fileName);
+      }
+
       auto dataSource = LoadDataReaction::loadData(fileName, options);
       dataSource->deserialize(dsObject);
+      if (fileName.isEmpty()) {
```
mathturtle: I loaded a dataset (a TIFF if it matters), saved the state file next to it and exited tomviz.  Then I renamed the folder they were in and tried reloading the state file.  I'm getting a segfault in `LoadDataReaction::loadData` just above here.

cjh1: @mathturtle So it looks like the fileName is also being stored in the "pvReaderXml" property in the case of a ParaView reader ( I tested with EMD, which doesn't show this problem ).

cjh1: @mathturtle I have pushed a commit that should fix this up.

tomviz/LoadDataReaction.cxx
```diff
@@ -211,6 +211,11 @@ DataSource* LoadDataReaction::loadData(const QStringList& fileNames,
         pxm->NewProxy(node.attribute("xmlgroup").as_string(),
                       node.attribute("xmlname").as_string()));
       if (tomviz::deserialize(reader, node)) {
+        // Use fileNames provided rather than relying on the XML. When loading
+        // a state file the file names provide are resolved relative to the
+        // state files so will survive a directory move.
+        vtkSMPropertyHelper fileNamesHelper(reader, "FileNames");
+        fileNamesHelper.Set(fileName.toLatin1().data());
```
cjh1: @cryos I am overriding the "FileNames" property here, to ensure we get file paths that are resolved relative to the state file location. Its not particularly nice that we have a fileName property in the JSON and one in the pvXmlReader XML, particularly as they are now going to be different. Thoughts?

cryos:
@cjh1 we should remove the XML in most cases, for TIFF it certainly has nothing of any importance.

cjh1:
@cryos So don't include the pvXmlReader property? What are the case we need the XML?

cryos:
I am not sure if we have any strong cases for using the XML, and if we do I think I would favor building that into the JSON state anyway. I was being cautious when porting it.

mathturtle:
I think we need it for the RAW reader since that one specifies how big the volume is in the XML.  I'm not sure about the other readers, but I can't think of one that has anything we care about...

cryos:
For the raw reader I think we should serialize that into the JSON too, and ideally I want to save/restore the dialog values. There are probably cases for VTK file types and XDMF where you select the array(s) to load.

cryos:
We can absolutely use this approach for now, just wanted to make my intent clear - I would rather not dump any XML into the state file. And certainly most readers have no need.

cjh1:
@cryos Just to be clear, we are already dumping XML for the readers in the state file. I think what you are saying is that we should move away from this? I agree. I can remove the reader XML as part of this PR?

cryos:
@cjh1 I am saying we should move away, but the fix you propose is good for the readers where we currently need to retain the XML (i.e. raw, VTK, etc). So I think you can merge this as-is, it fixes the immediate issue, and we can remove the XML. If there are some simple ones feel free to remove as part of this topic, but I would keep the code you added in the short term to set the file name after the proxy is created.

cjh1:
@mathturtle Can you confirm that the latest commit fixes the issue you where seeing. If so I will merge this, I plan on removing the XML in a follow up commit.

mathturtle:
Yes, the issue I found (crash when loading a pvXML-based file when the state file and data were moved) is fixed.

cjh1:
@mathturtle Thanks

PR #1481, opened 2018-04-09
Allow saving generated datasets

### LINKED ISSUE #1478: Generated data sources do not work in new state files
* Sample Data -> Generate Constant Dataset
* Accept default parameters
* Save state
* Reset/Restart tomviz
* Load state... crash

mathturtle:
Fixes #1478.

### DIFF
--- a/tomviz/DataSource.h
+++ b/tomviz/DataSource.h
@@ -70,7 +70,8 @@ public:
   /// Create a new dataSource not associated with a source proxy
   DataSource(const QString& label = QString(), DataSourceType dataType = Volume,
              QObject* parent = nullptr,
-             PersistenceState persistState = PersistenceState::Saved);
+             PersistenceState persistState = PersistenceState::Saved,
+             const QJsonObject& sourceInfo = QJsonObject());
 
   ~DataSource() override;
 
--- a/tomviz/PythonGeneratedDatasetReaction.cxx
+++ b/tomviz/PythonGeneratedDatasetReaction.cxx
@@ -40,6 +40,8 @@
 #include <QDialogButtonBox>
 #include <QGridLayout>
 #include <QHBoxLayout>
+#include <QJsonArray>
+#include <QJsonObject>
 #include <QLabel>
 #include <QSpinBox>
 #include <QVBoxLayout>
@@ -101,10 +103,9 @@ public:
     this->pythonScript = script;
   }
 
-  vtkSmartPointer<vtkSMSourceProxy> createDataSource(const int shape[3])
+  tomviz::DataSource* createDataSource(const int shape[3])
   {
-    vtkNew<vtkImageData> image;
-    vtkSmartPointer<vtkSMSourceProxy> retVal;
+    vtkImageData* image = vtkImageData::New();
 
     {
       tomviz::Python python;
@@ -116,7 +117,7 @@ public:
       args.set(1, shape[1]);
       args.set(2, shape[2]);
       tomviz::Python::Object imageData =
-        tomviz::Python::VTK::GetObjectFromPointer(image.Get());
+        tomviz::Python::VTK::GetObjectFromPointer(image);
       args.set(3, imageData);
       args.set(4, this->GenerateFunction);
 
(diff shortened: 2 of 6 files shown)


tomviz/PythonGeneratedDatasetReaction.cxx
```diff
@@ -541,31 +546,24 @@ void PythonGeneratedDatasetReaction::addDataset()
...
-  dataSource->setFileName(proxy->GetAnnotation(Attributes::LABEL));
-  LoadDataReaction::dataSourceAdded(dataSource, true /* default modules */,
-                                    false /* child */);
-}
-
-vtkSmartPointer<vtkSMSourceProxy>
-PythonGeneratedDatasetReaction::getSourceProxy(const QString& label,
-                                               const QString& script,
-                                               const int shape[3])
+DataSource* PythonGeneratedDatasetReaction::createGeneratedDataSource(
```
cryos: I think shorten the member name to createDataSource, you already have the generated in the class name, and the name is already very long :-)

tomviz/PythonGeneratedDatasetReaction.h
```diff
@@ -38,18 +38,15 @@ class PythonGeneratedDatasetReaction : public pqReaction
 
   void addDataset();
 
-  static vtkSmartPointer<vtkSMSourceProxy> getSourceProxy(const QString& label,
-                                                          const QString& script,
-                                                          const int shape[3]);
+  static DataSource* createGeneratedDataSource(
```
cryos: As before - shorten to createDataSource - generated is already clear from the class name

tomviz/PythonGeneratedDatasetReaction.cxx
```diff
@@ -129,33 +130,35 @@ class PythonGeneratedDataSource : public QObject
...
+
+    QJsonObject argsJson = QJsonObject::fromVariantMap(vmap);
+
+    QJsonArray size;
+    size.append(shape[0]);
+    size.append(shape[1]);
+    size.append(shape[2]);
+
+    QJsonObject description;
+    description["pythonScript"] = this->pythonScript;
```
cryos: Let's just call it script, this was used in the python operator too, I would argue that the member variable should be renamed too, but fine leaving as-is

tomviz/PythonGeneratedDatasetReaction.cxx
```diff
@@ -541,31 +546,24 @@ void PythonGeneratedDatasetReaction::addDataset()
...
-PythonGeneratedDatasetReaction::getSourceProxy(const QString& label,
-                                               const QString& script,
-                                               const int shape[3])
+DataSource* PythonGeneratedDatasetReaction::createGeneratedDataSource(
+  const QJsonObject& sourceInformation)
 {
-  PythonGeneratedDataSource generator(label);
-  generator.setScript(script);
+  PythonGeneratedDataSource generator(sourceInformation["label"].toString());
+  generator.setScript(sourceInformation["pythonScript"].toString());
```
cryos: This should also be moved to script, I don't think we need to specify language, and if so would likely do so as another key-value pair.

mathturtle: Makes sense

tomviz/DataSource.h
```diff
@@ -70,7 +70,8 @@ class DataSource : public QObject
   /// Create a new dataSource not associated with a source proxy
   DataSource(const QString& label = QString(), DataSourceType dataType = Volume,
              QObject* parent = nullptr,
-             PersistenceState persistState = PersistenceState::Saved);
+             PersistenceState persistState = PersistenceState::Saved,
+             const QJsonObject* sourceInfo = nullptr);
```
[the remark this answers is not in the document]
mathturtle: I didn't realize there was a null-like state to the QJsonObject.  I wanted it to do nothing by default and so made it a pointer.  But you have a good point.  I'll change it to a const ref.

mathturtle:
Updated and rebased on master now that #1480 is merged.

PR #1482, opened 2018-04-13
Remove XML blob for reader from state file

cjh1:
This PR introduced a new "reader" object which is a property of the data source in the state file. This is used to store any reader properties that were being stored as part of the XML blob.


mathturtle:
Loading a state file with generated data results in a "Unable to locate filename" warning (it seems to work fine though).

cjh1:
@mathturtle Thanks for the testing. I have pushed a change that should silent the warnings when dealing with a generated data source.

PR #1484, opened 2018-04-16
Serialize the color and opacity map, combine

cryos:
This now saves and restores custom colors that were changed by the user.
Also makes some tweaks to the gradient opacity, and ensures the module
and data source level match.


cryos:
Thanks

PR #1485, opened 2018-04-16
Operator state

cryos:


mathturtle:
I found a crash...

* Load a dataset, apply some operators (include a Set Tilt Angles)
* Use the C++ reconstruction operator
* Save state, reset, load state
* Quickly cancel the reconstruction while it is running (not sure this is necessary, but I'm impatient)
* Delete the reconstruction operator (the crash is in a lambda in DataSource::deserialize that I think needs a better signal connection lifetime)

mathturtle:
And another:

* Load a dataset apply some operators
* Snapshot the data
* Save data for the snapshotted data (not sure this is necessary)
* Save state, reset, load state... crash.  Looks like the modules on the output are messed up?

cryos:
I don't think either of these issues is the result of this branch, I have created issues #1486 and #1487 to track them.

mathturtle:
Sounds good.  i didn't test master to see if these were reproducible there.

cryos:
I am still tracking down a few issues like that, some big changes

Issue #1435, opened 2018-03-08
Check Memory Usage and Stability


Hovden:
Check that the memory usage in tomviz is reasonable (i.e. does not fill up over time).

cjh1:
![mem](https://user-images.githubusercontent.com/1031053/37793876-422c190c-2de6-11e8-81cc-68c6b3c98e67.png)

cjh1:
The chart above shows 10 executions of the invert operator.

cjh1:
Looks like we have some leaking VTK object: 

```
vtkDebugLeaks has detected LEAKS!
Class "vtkCellData" has 1 instance still around.
Class "vtkPointData" has 1 instance still around.
Class "vtkPoints" has 1 instance still around.
Class "vtkInformation" has 3 instances still around.
Class "vtkInformationIntegerPointerValue" has 1 instance still around.
Class "vtkIdList" has 1 instance still around.
Class "vtkDoubleArray" has 1 instance still around.
Class "vtkInformationIntegerValue" has 1 instance still around.
Class "vtkImageData" has 1 instance still around.
Class "vtkVoxel" has 1 instance still around.
Class "vtkFieldData" has 1 instance still around.
Class "vtkUnsignedCharArray" has 1 instance still around.
Class "9vtkBufferIdE" has 1 instance still around.
Class "vtkCommand or subclass" has 1 instance still around.
Class "9vtkBufferIhE" has 1 instance still around.
```

mathturtle:
Rats... I thought I'd got that one with the pipeline fixes last week. :(

cquammen:
I have tracked down one memory leak of the image allocated at https://github.com/OpenChemistry/tomviz/blob/943e0b25d1aa04215ae79f9da6d1cda2f9fd4697/tomviz/Pipeline.cxx#L115 when applying the Binary Threshold transform to original data. I did not see how to fix it, though.

cryos:
It looks like the ITK-based operators are leaking, I just talked with @cjh1 and that data should flow through the pipeline and become the result, or if the operation is cancelled the result gets deleted. I am guessing this logic is breaking down in the ITK-based pipeline somewhere.

cryos:
It may be more general than that, the reconstruction operators too, so anything where it makes its own child data I guess. There are also some odd behaviors for reconstruction where the modules are moved to the output but they are showing the original data still.

cryos:
[...]

PR #1492, opened 2018-04-18
Data ownership

cryos:
Fixes issues referred to in #1435

### DIFF
--- a/tomviz/DataSource.cxx
+++ b/tomviz/DataSource.cxx
@@ -770,7 +770,6 @@ void DataSource::setData(vtkDataObject* newData)
   auto tp = producer();
   Q_ASSERT(tp);
   tp->SetOutput(newData);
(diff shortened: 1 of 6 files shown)


tomviz/DataSource.cxx
```diff
@@ -770,7 +770,6 @@ void DataSource::setData(vtkDataObject* newData)
   auto tp = producer();
   Q_ASSERT(tp);
   tp->SetOutput(newData);
-  newData->FastDelete();
```
mathturtle: Can you test this with generated data?  I remember having to tweak some generated data stuff related to this FastDelete call.

cryos: Worked fine with random particles here

mathturtle: I see, you fixed the other side of it too.  +1

PR #1493, opened 2018-04-19
Remove extra delete

cjh1:
We now do a fast delete at the start of the pipeline execution.


cjh1:
@cryos We missed this we the ownership changes.

PR #1515, opened 2018-04-20
Ensure we cast JSON arguments for Python operators

### LINKED ISSUE #1496: Restoring shift volume from state file fails
I end up with a pipeline that looks like this:
image
The error:
```
  File "Shift Volume", line 15, in transform_scalars
  File "/usr/lib/python3.6/site-packages/numpy/core/numeric.py", line 1393, in roll
    result[res_index] = a[arr_index]
TypeError: slice indices must be integers or None or have an __index__ method
Failed to execute the script.
```
Hitting apply repeats the error, and changing a shift and hitting apply causes a segfault.

cryos:
The JSON files that generate the GUIs already have type information, use
that to cast the JSON lists/values when deserializing. Fixes #1496.

### DIFF
--- a/tomviz/operators/OperatorPython.cxx
+++ b/tomviz/operators/OperatorPython.cxx
@@ -484,13 +484,68 @@ QJsonObject OperatorPython::serialize() const
   return json;
 }
 
+namespace {
+QJsonObject findJsonObject(const QJsonArray& array, const QString& name)
+{
+  for (int i = 0; i < array.size(); ++i) {
+    auto object = array[i].toObject();
+    if (object["name"].toString() == name) {
+      return object;
+    }
+  }
+  return QJsonObject();
+}
+
+QVariant castJsonArg(const QJsonValue& arg, const QString& type)
+{
+  if (arg.isArray()) {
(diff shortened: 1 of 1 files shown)


tomviz/operators/OperatorPython.cxx
```diff
@@ -484,13 +484,68 @@ QJsonObject OperatorPython::serialize() const
...
+    auto arr = arg.toArray();
+    QVariantList arrayList;
+    if (type == "int" || type == "enumeration") {
+      for (int i = 0; i < arr.size(); ++i) {
+        arrayList << arr[i].toInt();
+      }
+    } else if (type == "double") {
+      for (int i = 0; i < arr.size(); ++i) {
+        arrayList << arr[i].toDouble();
+      }
```
cjh1: I think we can avoid duplicating the int  || enumeration and double blocks by recursively calling ```castJsonArg ``` for each element in the array?

cryos: Agreed, especially if we want to get to nested arrays etc. Merging what we have for now.

PR #1528, opened 2018-04-24
Fix filenames

alesgenova:
Remove fileName / fileNames redundancy in state files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 *.pyc
 .DS_Store
 .cache
+.vscode
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
 *.pyc
 .DS_Store
 .cache
-.vscode
--- a/tomviz/DataSource.cxx
+++ b/tomviz/DataSource.cxx
@@ -232,16 +232,26 @@ bool DataSource::appendSlice(vtkImageData* slice)
 
 void DataSource::setFileName(const QString& filename)
 {
-  auto reader = m_json.value("reader").toObject(QJsonObject());
-  reader["fileName"] = filename;
-  m_json["reader"] = reader;
+  // AG: remove fileName and fileNames redundancy. but keep backward
+  // compatibility.
+  QStringList fileNames = QStringList(filename);
+  setFileNames(fileNames);
+
+  // auto reader = m_json.value("reader").toObject(QJsonObject());
+  // reader["fileName"] = filename;
+  // m_json["reader"] = reader;
 }
 
 QString DataSource::fileName() const
 {
+  // AG: remove fileName and fileNames reduntancy. but keep backward
+  // compatibility.
   auto reader = m_json.value("reader").toObject(QJsonObject());
-  if (reader.contains("fileName")) {
-    return reader["fileName"].toString();
+  if (reader.contains("fileNames")) {
+    auto fileNames = reader["fileNames"].toArray();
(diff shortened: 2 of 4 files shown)


tomviz/DataSource.cxx
```diff
@@ -232,16 +232,26 @@ bool DataSource::appendSlice(vtkImageData* slice)
 
 void DataSource::setFileName(const QString& filename)
 {
-  auto reader = m_json.value("reader").toObject(QJsonObject());
-  reader["fileName"] = filename;
-  m_json["reader"] = reader;
+  // AG: remove fileName and fileNames redundancy. but keep backward
```
cryos: We can see from git blame who authored comments, no need to prefix with initials etc. Ideally keep as a sentence, capital first letter, full stop/period. I think this should be removed (looks more like the commit message to me), or reworded to explain that it is a convenience method.

tomviz/DataSource.cxx
```diff
@@ -232,16 +232,26 @@ bool DataSource::appendSlice(vtkImageData* slice)
...
+  // compatibility.
+  QStringList fileNames = QStringList(filename);
+  setFileNames(fileNames);
+
+  // auto reader = m_json.value("reader").toObject(QJsonObject());
```
cryos: Should just remove code.

tomviz/DataSource.cxx
```diff
@@ -232,16 +232,26 @@ bool DataSource::appendSlice(vtkImageData* slice)
...
+  // reader["fileName"] = filename;
+  // m_json["reader"] = reader;
 }
 
 QString DataSource::fileName() const
 {
+  // AG: remove fileName and fileNames reduntancy. but keep backward
```
cryos: As above, change looks reasonable but commit message would be better place to explain the change.

tomviz/LoadDataReaction.cxx
```diff
@@ -168,16 +168,16 @@ DataSource* LoadDataReaction::loadData(const QStringList& fileNames,
   bool child = options["child"].toBool(false);
 
   DataSource* dataSource(nullptr);
-  QString fileName;
+  QString firstFile;
```
cryos: This feels odd, not sure if it is necessary but if you and @cjh1 want it for additional clarity it is fine by me.

tomviz/LoadDataReaction.cxx
```diff
@@ -246,10 +246,9 @@ DataSource* LoadDataReaction::loadData(const QStringList& fileNames,
   }
 
   // Now for house keeping, registering elements, etc.
-  dataSource->setFileName(fileName);
-  if (fileNames.size() > 1) {
-    dataSource->setFileNames(fileNames);
-  }
+  // always save it as a list, even if there is only one file.
```
cryos: Capitalize first letter of first work.

.gitignore
```diff
@@ -1,3 +1,4 @@
 *.pyc
 .DS_Store
 .cache
+.vscode
```
cryos: We don't add editor specific files to this, you can set up a global ignore locally.

cjh1:
@cryos Are you happy for this to go in now? I think @alesgenova has addressed the review comments?

cjh1:
Actually, looks like you wanted the commit messages reworked?

cryos:
It is fine, I just reverted the variable name change, and pushed that, I think we can merge now.

PR #1530, opened 2018-04-24
Remove json newlines

cryos:


cryos:
Abusing my superpowers...it is just removing new line breaks from some JSON :sweat_smile:

PR #1532, opened 2018-04-24
WIP: Serialize the camera paths from animations

mathturtle:
This saves and loads the values in the keyframes (as far as I can tell), but something breaks in the loading of the state file and the animation doesn't move the camera.  It may have to do with there being two orbits, but deleting one doesn't seem to help.

### DIFF
--- a/tomviz/Utilities.cxx
+++ b/tomviz/Utilities.cxx
@@ -222,6 +222,88 @@ bool deserialize(vtkSMProxy* proxy, const QJsonObject& json)
   return true;
 }
 
+QJsonObject serializeCameraAnimationKeyFrame(vtkSMProxy* keyframe)
+{
+  QJsonObject frameObj;
+  frameObj["closedFocalPath"] =
+    vtkSMPropertyHelper(keyframe, "ClosedFocalPath").GetAs<int>();
+  frameObj["closedPositionPath"] =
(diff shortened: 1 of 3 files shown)


cryos:
I need to head out, I pushed slight refactoring moving a little more into Utilities. I see two cameras - the default one, and then one that has my custom path in it. So it is certainly moving in the right direction, why don't you take a pass at doing what you and Utkarsh discussed from the tip of this branch.

mathturtle:
The problem is that the branch is already doing what Utkarsh and I discussed (it doesn't restore the timekeeper, I forgot to take out the save code).  I think we need to add an option to not add the camera orbit when loading the data from a state file (one is from the load data reaction and one is from the load).  But I'm not sure if the problem is the two orbits or if it is something else.  I can take another look at this tomorrow I guess.

cryos:
After some investigation it is clear that restoring animations in Tomviz 1.2.1 is not working, and on Windows/Linux is causing update issues, and other unexpected side effects. I am going to table this for now, as it looks like there are some bigger underlying issues we need to tackle there.

Comparing 1.2.1 and master, there are 2 camera objects, but other than that they exhibit the same bug - playing apparently does nothing, but pausing will then show an updated scene at a new angle. I reproduced this in master/latest on Linux, and saw the same issue in the last release on both Windows and Linux.

cryos:
Referenced from relevant issues, closing for now as the current approach is not functional.

PR #1533, opened 2018-04-24
Correctly restore operators added to child data source

cjh1:
This PR improves the restoration of operators added to child data source. This allow more heavily nested pipelines such as the one below that was successfully restored.

screen


cryos:
:shipit: !!!

Issue #1284, opened 2017-10-06
Physical pixel size issues


ercius:
The physical size (voxel size) is not set properly for some data sets. A couple examples below:
1) EMD files do not have the proper physical size scaling or units.
2) The example data sets for the star particle do not have proper physical dimensions set.
3) MRC files are read properly. However, loading an MRC and saving as TIFF and loading the TIFF loses the z-axis scaling. This seems connected to the saving operation because of 4) below.
4) Downloading data from Hovden's data set paper in TIFF format (tiltser_180.tiff) seems to get the correct scaling.

I do not have many more examples to try and load. These are some possible issues to look into though.

mathturtle:
There may be two issues here, only one of which can be solved.

1. Not all image formats support storing physical size of the pixels.  For example, TIFF does not have a field for pixel size and so tomviz has to guess when loading a TIFF what the pixel size should be.  I don't remember which formats support storing the physical size off the top of my head although I know OME-TIFF does store the size.  MRC may have it or you may have gotten lucky with the initial size guess.

2. I **think** EMD files should store the physical size of the pixel and if they are not reading it correctly that could be a bug.

cryos:
EMD should store/read the size, I don't think it does at present, but I can fix that. TIFF has poor support, the standard can store a spacing in x and y, but not z. We improved it slightly by having z match x I think (rather than being left at 1). If there is a way of using z spacing in TIFF I don't think our reader/write code is aware of it.

ercius:
This is why I dislike TIFF files for microscopy data. The header and pixel sizes are not well supported. ImageJ seems to have a decent handle on the XY size, but the third dimension is difficult to calibrate (i.e. I dont know how). OME-TIFF is a good way to go, but I am not sure how to implement it.

It seems odd to have example data that does not import a physical size. Maybe we should change the example data to a format that supports pixel sizes correctly (MRC)? Just a though.

MRC definitely supports pixel sizes although the standard vs. FEI header information sometimes gives headaches for where to get the correct information. tomviz correctly reads MRC file headers as far as I know.

For EMD I suggest using the first element of the dim_n vector as the offset and the dim_n[1] - dim_n[0] as the pixel size along that direction. Some dim directions will only consist of those 2 elements.

ercius:
[...]

PR #1536, opened 2018-04-25
Include full width of image in size calculation

### LINKED ISSUE #1506: Units and size calculation
I believe the size in the units and size property panel calculated incorrectly? If you load the example Nanoparticle tilt series then size in 255,255,73 nm. It should be 256,256,74 if the pixel size defaults to 1 pixel/nm.

mathturtle:
Before the size calculation measured pixel center to pixel center for
the edge pixels, but there was an off-by-1 that didn't include the half
pixel on each side between the pixel centers and the edge.

Fixes #1506.  I looked at this one because I thought I remembered fixing this problem already... apparently I didn't?

### DIFF
--- a/tomviz/DataPropertiesPanel.cxx
+++ b/tomviz/DataPropertiesPanel.cxx
@@ -227,11 +227,11 @@ void DataPropertiesPanel::updateData()
   dsource->getExtent(extent);
   dsource->getSpacing(spacing);
   m_ui->xLengthBox->setText(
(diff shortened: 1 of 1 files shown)


ercius:
> Fixes #1506. I looked at this one because I thought I remembered fixing this problem already... apparently I didn't?

I think it was looked at a couple of times in some previous issues (ex. #1284 and #1278). Im not sure what was fixed back then, but hopefully its fully fixed now.

PR #1537, opened 2018-04-25
Warn the user that loading state files resets the current state

mathturtle:
I'm not sure if we want a 'Do not show this again' checkbox... if so this will need more work.

### DIFF
--- a/tomviz/SaveLoadStateReaction.cxx
+++ b/tomviz/SaveLoadStateReaction.cxx
@@ -25,6 +25,7 @@
 #include <QFileDialog>
 #include <QJsonDocument>
 #include <QJsonObject>
+#include <QMessageBox>
 
 #include <QDebug>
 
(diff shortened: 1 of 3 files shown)


cjh1:
@mathturtle This looks good, it does however warn even if the workspace is empty. We could just check with the ```ModuleManager``` and just issue the warning is any data sources are loaded?

cryos:
This would be irritating for empty loads, it should minimally check for some loaded data sets. We may want to get a little cleverer in future, but that should be fine for now.

mathturtle:
Updated to check if there are any data sources before showing the dialog.

tomviz/SaveLoadStateReaction.cxx
```diff
@@ -79,6 +80,16 @@ bool SaveLoadStateReaction::loadState()
 
 bool SaveLoadStateReaction::loadState(const QString& filename)
 {
+  if (ModuleManager::instance().hasDataSources()) {
+    if (QMessageBox::No ==
+        QMessageBox::warning(
+          tomviz::mainWidget(), "Load State Warning",
+          "Current data and operators will be cleared when "
+          "loading a state file.  Proceed anyway?",
+          QMessageBox::Yes | QMessageBox::No, QMessageBox::No)) {
```
cjh1: @alesgenova Pointed out to me earlier today that ```QMessageBox::warning(...)``` are marked as obsolete and should not be used in new code. We doesn't have to do it in this PR, but we should move away from using them.

cjh1: Actually, it's only the signatures that take int's for the button types .... as you were.

cjh1:
@mathturtle Just a few clang-format issues.

PR #1546, opened 2018-04-26
Ensure we favor color map from data source when selected

cjh1:
Use the color map from the data source the user has selected, rather than the active module.


cjh1:
@cryos Better?

PR #1559, opened 2018-05-01
Pybind11 upgrade

cjh1:


cjh1:
@cryos Can we merge this one now?

Issue #1560, opened 2018-05-01
Crash when adding/deleting cloned data


cquammen:
* Launch tomviz
* Load a data set
* Clone the data (Data only, but it might not matter)
* Delete the clone
* Clone the data again
* Delete it again. Crash. It may take a few more cycles of cloning and deleting to see the crash.

Stack trace

```
0                                 	0x000000010fd64be3 tomviz::PipelineModel::parent(QModelIndex const&) const + 51
1   org.qt-project.QtCore         	0x0000000116ed84f3 QAbstractItemModelPrivate::rowsAboutToBeInserted(QModelIndex const&, int, int) + 179
2   org.qt-project.QtCore         	0x0000000116edd79f QAbstractItemModel::beginInsertRows(QModelIndex const&, int, int) + 399
3                                 	0x000000010fd65a69 tomviz::PipelineModel::dataSourceAdded(tomviz::DataSource*) + 105
4   org.qt-project.QtCore         	0x0000000116f5e5aa QMetaObject::activate(QObject*, int, int, void**) + 2954
5                                 	0x000000010fe2278f tomviz::ModuleManager::dataSourceAdded(tomviz::DataSource*) + 63
6                                 	0x000000010fdbaa66 tomviz::ModuleManager::addDataSource(tomviz::DataSource*) + 326
7                                 	0x000000010fd58458 tomviz::LoadDataReaction::dataSourceAdded(tomviz::DataSource*, bool, bool) + 168
8                                 	0x000000010fd27f9b tomviz::CloneDataReaction::clone(tomviz::DataSource*) + 443
9   org.qt-project.QtCore         	0x0000000116f5e5aa QMetaObject::activate(QObject*, int, int, void**) + 2954
10  org.qt-project.QtWidgets      	0x00000001108ec2e5 QAction::activate(QAction::ActionEvent) + 309
```

cquammen:
The data cloning may be incidental, but I haven't been able to get a crash when just adding/deleting operators that have been applied in succession to a loaded data set.

cryos:
This branch seems to have cleaned up issues outlined.

PR #1561, opened 2018-05-02
Fix up the way modules are remove from a data source

cjh1:


cjh1:
Fixes #1560

cryos:
I was going to let @cquammen take a look, but wanted to say that the commit messages look great!

cquammen:
Taking a look now after getting tomviz built with the required ParaView and VTK versions.

cquammen:
Small hiccup: After deleting a clone, the selected element in the pipeline browser is the original data set. However, the context menu that comes up when right clicking the mouse does not enable the `Clone` option again. I have to click a different pipeline object and then click back on the original data to update the context menu.

cquammen:
In fact, all the transforms in the Data Transforms and Segmentation menus are disabled as well until I click a different pipeline object and click the original data again.

cjh1:
@cquammen Is the behavior the same on master? Just try to work out if my commits broke something?

cquammen:
Okay, sorry. After getting ParaView up to the version tomviz needs I forgot to actually grab your branch :face_with_head_bandage: Menu items are enabled as expected. Sorry about that.

Still no crashes when deleting the data, which was the problem you addressed.

Maybe a minor thing, but when I clone a data set, then delete it, upon closing tomviz I get this message:

```
QAbstractItemModel::endRemoveRows:  Invalid index ( -1 , 0 ) in model tomviz::PipelineModel(0x2bb2850)
```

Probably doesn't need to be fixed in this MR, just wanted to note it here.

cjh1:
@cquammen Thanks for the testing. I will try to clear up that warning.

cjh1:
@cquammen I have cleaned up the warning. PTAL.

cquammen:
The warning is gone. Thanks, this is working nicely now! 👏👏👏

Issue #1569, opened 2018-05-07
File name missing when state file restored


cryos:
I think this is related to #1528 where the single file name code path was setting the file name on the data source. You can load a data set, save the state, restore, and see the following pipeline now:
![image](https://user-images.githubusercontent.com/42214/39720382-222c6a2e-520a-11e8-9f22-6f91817da0e7.png)
This should list the name of the file, in the case of a series maybe the name of the first file (or it could indicate it is a stack of images).

cryos:
Fixed by #1570

PR #1570, opened 2018-05-07
Remove incorrect empty check

cjh1:


cjh1:
Fixes #1569

cryos:
I see what you mean now, it fixes things for the two test cases I was looking at. Let's hold off on merging until we see what we are doing with the Linux binary.

Issue #1571, opened 2018-05-08
Warning from ParaView on loading state files


cryos:
This is probably a simple bump, but the release issues the following warning on loading state files that should be fixed:
`vtkSMStateVersionController (0x31d2330): State file version is less than 4.0.We will try to load the state file. It's recommended, however, that you load the state in ParaView 4.0.1 (or 4.1.0) and save a newer version so that it can be loaded more faithfully. Loading state files generated from ParaView versions older than 4.0.1 is no longer supported.`

cryos:
This is actually the interaction of versioning the ParaView the release built against, hence not seeing it locally, and a hardwired check for versions less than 4.0.1... We do set the version to the current ParaView version, which resolves to 1.3.0. You live and learn, I can hardwire our version to be greater than that version.

cryos:
Fixed.

PR #1574, opened 2018-05-08
Hardwire the ParaView state file version

cryos:
This fixes issue #1571 where a hardwired 4.0.1 check is triggered.


cryos:
CircleCI never came back, but was there for other PRs, assuming a temporary issue.

PR #1577, opened 2018-05-09
GUI support for external pipeline execution

cjh1:
Open issues/TODO:

- [x] Test on Windows. ( require some changes! )
- [x] Test on OSX.
- [x] Currently re-runs the entire pipeline.
- [x] We are currently using QConcurrent to avoid a deadlock with a blocking connection. This may not be necessary.
- [ ] We should check that docker is installed if the user selects the docker execution mode.
- [ ] Currently doesn't work with live updates
- [ ] Currently doesn't work with child data sources.
- [x] The executors should probably to move to separate files.
- [ ] Operators that have dialogs that need the image data  are currently not supported.


cjh1:
@cryos @alesgenova This is ready for review and testing.

cryos:
For me on the first run it said it was pulling the docker container, and then just left a question marker for the operator and nothing happened, I then added the invert operator, and got this,
image
I don't think the square root ever ran. The second run through things went smoother, so it may just be the initial pull of the container where something didn't quite go right?

cryos:
I can't reproduce the issue I posted above, even when deleting/redownloading the image. Guessing it was a very unlucky timing issue, or something like this. Things seem to work really well for me locally in general, especially once I have the image downloaded.

alesgenova:
Cool, it works!!

Couple of things I noticed while first testing it:

1) After the dialog informing me that the image was being downloaded closed, nothing happened and I though the operation didn't go through (operator showing question mark). But after about 5-10 seconds, the operator was applied. I think you're missing showing the dialog saying that the image is being started

2) I tried the invert and square root operators and they worked fine. I randomly tried the "add a constant" operator and I received an error from docker (attached below)

```
Docker container exited with non-zero exit code: 1.

Docker logs below: 

 /usr/local/lib/python3.6/site-packages/h5py/__init__.py:36: FutureWarning: Conversion of the second argument of issubdtype from `float` to `np.floating` is deprecated. In future, it will be treated as `np.float64 == np.dtype(float).type`.
  from ._conv import register_converters as _register_converters
[2018-06-05 12:59:33,082] INFO: Executing 'Add a Constant' operator
  File "/usr/local/bin/tomviz-pipeline", line 11, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/tomviz/cli/__init__.py", line 80, in main
    output_file_path, progress_method, socket_path)
  File "/usr/local/lib/python3.6/site-packages/tomviz/executor.py", line 435, in execute
    data = _execute_transform(label, transform, arguments, data, progress)
  File "/usr/local/lib/python3.6/site-packages/tomviz/executor.py", line 387, in _execute_transform
    transform(input, **arguments)
  File "/tmp/tmpvg45k1fa.py", line 15, in transform_scalars
AttributeError: 'int' object has no attribute 'is_integer'
3
1
Operator doesn't support progress updates.
```

cjh1:
All outstanding issues have been moved to the [Docker milestone](https://github.com/OpenChemistry/tomviz/milestone/16)

PR #1578, opened 2018-05-10
Improve saving the view state in the state file

mathturtle:
This saves the visibility state of the axes grid, the perspective vs.
orthographic projection mode and more camera properties (needed for
orthographic projection mode) in the state file and loads them when the
state is restored.

@cryos These are the easy bits.  Getting to the scale legend's state from the serialization code will require more refactoring.

### DIFF
--- a/tomviz/modules/ModuleManager.cxx
+++ b/tomviz/modules/ModuleManager.cxx
@@ -496,6 +496,30 @@ bool ModuleManager::serialize(QJsonObject& doc, const QDir& stateDir,
           if (helper.GetAsInt()) {
             backgroundColor.append(jsonArrayFromXmlDouble(node));
           }
+        } else if (name == "CameraParallelScale") {
+          vtkSMPropertyHelper helper(view, "CameraParallelScale");
+          camera["parallelScale"] = jsonArrayFromXmlDouble(node)[0];
+        } else if (name == "CameraParallelProjection") {
+          // orthographic or perspective projection
+          vtkSMPropertyHelper helper(view, "CameraParallelProjection");
+          jView["isOrthographic"] = helper.GetAsInt() != 0;
+        } else if (name == "InteractionMode") {
+          vtkSMPropertyHelper helper(view, "InteractionMode");
+          QString mode = "3D";
+          if (helper.GetAsInt() == 1) {
+            mode = "2D";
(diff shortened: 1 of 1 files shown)


tomviz/modules/ModuleManager.cxx
```diff
@@ -496,6 +496,21 @@ bool ModuleManager::serialize(QJsonObject& doc, const QDir& stateDir,
...
+          vtkSMPropertyHelper helper(view, "InteractionMode");
+          jView["interactionMode"] = helper.GetAsInt();
+        }
+      }
+      if (view->GetProperty("AxesGrid")) {
+        vtkSMPropertyHelper helper(view, "AxesGrid");
+        vtkSMProxy* axesGridProxy = helper.GetAsProxy();
+        if (axesGridProxy) {
+          vtkSMPropertyHelper visibilityHelper(axesGridProxy, "Visibility");
+          jView["axesGridVisibility"] = visibilityHelper.GetAsInt();
```
cryos: We should do an != 0 to turn into a bool in JSON.

tomviz/modules/ModuleManager.cxx
```diff
@@ -723,6 +743,23 @@ bool ModuleManager::deserialize(const QJsonObject& doc, const QDir& stateDir)
...
+      vtkSMPropertyHelper axesGridProp(viewProxy, "AxesGrid");
+      vtkSMProxy* proxy = axesGridProp.GetAsProxy();
+      if (!proxy) {
+        vtkSMSessionProxyManager* pxm = viewProxy->GetSessionProxyManager();
+        proxy = pxm->NewProxy("annotations", "GridAxes3DActor");
+        axesGridProp.Set(proxy);
+        proxy->Delete();
+      }
+      vtkSMPropertyHelper(proxy, "Visibility")
+        .Set(view["axesGridVisibility"].toInt());
```
cryos: This may then need a ternary to turn back into an int

mathturtle:
Ok, I've switched visibility of the axes grid and isOrthographic to booleans.

mathturtle:
This new version saves the interaction mode and the camera parallel projection state independently (as they are in the ParaView view).

cryos:
This now works as expected, restoring the parameters in the view menu. The interaction mode is and should be separate.

PR #1579, opened 2018-05-11
Save scale legend state

mathturtle:
This required a bit more refactoring that I think we want in a bugfix release...

I had to change the scale legend from being one global object that operated on the first view ever created to having one instance per view.  Then I could save it along with the rest of the view settings.


tomviz/modules/ModuleManager.cxx
```diff
@@ -724,6 +736,21 @@ bool ModuleManager::deserialize(const QJsonObject& doc, const QDir& stateDir)
...
+    if (view.contains("scaleLegend")) {
+      auto scaleLegend = ScaleLegend::getScaleLegend(viewProxy);
+      if (scaleLegend) {
+        auto legendType = view["scaleLegend"].toString();
+        if (legendType == "none") {
+          scaleLegend->setVisibility(false);
+        } else if (legendType == "cube") {
+          scaleLegend->setVisibility(true);
+          scaleLegend->setStyle(ScaleLegendStyle::Cube);
+        } else if (legendType == "ruler") {
```
cryos: This shouldn't happen, but we should probably favor a default, and perhaps issue a warning. This is not enough to correctly update the view menu with the state.

mathturtle:
> I mean that upon loading the state, either from a newly opened Tomviz, or one that had data and reset/loaded state, that the view menu is showing the default (neither scale type ticked, hide option disabled).

Ah, ok.  This should be fixed in the new version (it was missing an update call to make the view menu reset itself after the additional properties are loaded).

cryos:
Once we start reviewing code it is a lot easier to see how you fixed something by adding new commits, rather than rewriting. It shows up easily in the GitHub UI, and I can just look at the last commit, this would be good to bear in mind in future. Taking a look now.

tomviz/modules/ModuleManager.cxx
```diff
@@ -782,7 +794,29 @@ bool ModuleManager::deserialize(const QJsonObject& doc, const QDir& stateDir)
...
+          scaleLegend->setStyle(ScaleLegendStyle::Ruler);
+        } else { // corrupted file?
+          qWarning() << "Unknown scale legend type" << legendType
+                     << "encountered in state file.  Defaulting to \"none\".";
+          scaleLegend->setVisibility(false);
+        }
+      }
+    }
   }
+  // force the view menu to update its state based on the settings we have
```
cryos: Minor, prefer sentence with capital first letter, period at end of sentence. This could be shorter, just `// Ensure the view menu reflects restored state.` perhaps

PR #1583, opened 2018-05-16
Histogram fixes

cryos:


cryos:
We require C++11, and so if calling isfinite should go straight for the STL version, the call only makes sense for double/float too, integers are inherently finite, so some specialization would help even if we keep finite checks for float/double. Need to fix my formatting too.

Issue #1511, opened 2018-04-19
Passive acquisition usability


ercius:
Can this be used now for testing? I get a dialog box asking for where the server is located:
![image](https://user-images.githubusercontent.com/15657108/39018388-bc4e4f48-43da-11e8-9941-8d98bf0ec2d4.png)

File name regex is very complex for a non-expert. Can you add an example filename which would be properly formatted? Or a button to "test" read a directory to see if the regex matching works properly?

![image](https://user-images.githubusercontent.com/15657108/39018228-4d26ad54-43da-11e8-86df-191abd446d94.png)

cryos:
This is getting more usable, @cjh1 developed a lot of this, I think that a tutorial document would help, trying to remember if we have an up to date one. Hopefully Chris will chime in.

cjh1:
I think have a test button is a good idea. I think in the future we will move the regex's into an advanced view and just provide a few "canned" formats that will match common file naming conventions. Regex's give us the maximum flexibility but I agree they are a little intimidating to the end user so we should provide them with a little more help.

alesgenova:
@ercius 
I have almost finished redesigning the passive acquisition dialog, you can find some screenshots of the current status at #1585 
Let me know if you have further suggestions on the interface

PS:
The old dialog will still be available when clicking the "advanced" option in the dropdown (not implemented yet)

ercius:
The new dialog looks excellent. Thanks for the work on this and for the feedback. Ill have to test it out to make more suggestions. I think this goes a long way to making this useful for non-expert users. Ill close this now and open a new issue for other suggestions.

PR #1585, opened 2018-05-19
Redesign the passive acquisition dialog

alesgenova:
Related to #1511 

### New Dialog
screenshot from 2018-05-24 13-38-31

### Custom Filename Format
screenshot from 2018-05-24 13-42-11

### Advanced Filename Format
screenshot from 2018-05-24 13-41-09

### Wrong Filename Format
screenshot from 2018-05-24 13-40-22

### DIFF
--- a/tomviz/acquisition/BasicFormatWidget.h
+++ b/tomviz/acquisition/BasicFormatWidget.h
@@ -33,49 +33,50 @@ class BasicFormatWidget;
 
 namespace tomviz {
 
-enum class TestRegexFormat {
+enum class TestRegexFormat
+{
   npDm3,
   pmDm3,
   npTiff,
   pmTiff,
-  Custom,
-  Advanced
+  Custom
 };
 
-class BasicFormatWidget : public QWidget {
+class BasicFormatWidget : public QWidget
+{
   Q_OBJECT
 
-  public:
-    BasicFormatWidget(QWidget *parent = nullptr);
-    ~BasicFormatWidget() override;
-    
-    QString getRegex() const;
-    QString getPythonRegex() const;
-    MatchInfo matchFileName(QString) const;
-    QJsonArray getRegexGroups() const;
-    QJsonObject getRegexSubsitutions() const;    
+public:
+  BasicFormatWidget(QWidget* parent = nullptr);
+  ~BasicFormatWidget() override;
 
-  public slots:
+  QString getRegex() const;
+  QString getPythonRegex() const;
+  MatchInfo matchFileName(QString) const;
+  QJsonArray getRegexGroups() const;
+  QJsonObject getRegexSubsitutions() const;
 
-  private slots:
-    void formatChanged(int);
+public slots:
 
+private slots:
+  void onComboChanged(int);
 
-  signals:
-    void fileFormatChanged(TestRegexFormat);
-    void regexChanged(QString);
+signals:
+  void fileFormatChanged(TestRegexFormat);
+  void regexChanged(QString);
 
-  private:
+private:
   QScopedPointer<Ui::BasicFormatWidget> m_ui;
 
-  QRegExp m_fileNameRegex;
+  QString m_fileNameRegex;
   QString m_negChar;
   QString m_posChar;
   QString m_pythonFileNameRegex;
   QString m_testFileName;
 
   void setupFileFormatCombo();
-  
+  void setupRegexDisplayLine();
+
   QList<TestRegexFormat> makeDefaultFormatOrder() const;
   QMap<TestRegexFormat, QString> makeDefaultFileNames() const;
   QMap<TestRegexFormat, QString> makeDefaultLabels() const;
@@ -90,9 +91,6 @@ class BasicFormatWidget : public QWidget {
   void customFileRegex();
   void validateFileNameRegex();
 };
-
 }
-
-
 
 #endif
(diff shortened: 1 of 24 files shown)


tomviz/acquisition/MatchInfo.h
```diff
@@ -0,0 +1,21 @@
...
+namespace tomviz {
+
+struct CapGroup {
+  CapGroup(QString name_, QString text) : name(name_), capturedText(text) {}
+  QString name;
+  QString capturedText;
+};
+
+struct MatchInfo {
+  bool matched;
```
cjh1: Why do we need this? If groups contains items then we know the regex matched?

tomviz/acquisition/BasicFormatWidget.h
```diff
@@ -0,0 +1,96 @@
...
+  BasicFormatWidget(QWidget* parent = nullptr);
+  ~BasicFormatWidget() override;
+
+  QString getRegex() const;
+  QString getPythonRegex() const;
+  MatchInfo matchFileName(QString) const;
+  QJsonArray getRegexGroups() const;
+  QJsonObject getRegexSubsitutions() const;
+
+public slots:
```
cryos: Don't need empty public slots, just remove this and the empty lines

tomviz/acquisition/CustomFormatWidget.cxx
```diff
@@ -0,0 +1,94 @@
+#include "CustomFormatWidget.h"
```
cryos: Missing copyright header

tomviz/acquisition/CustomFormatWidget.h
```diff
@@ -0,0 +1,45 @@
...
+
+class CustomFormatWidget : public QWidget
+{
+  Q_OBJECT
+
+public:
+  CustomFormatWidget(QWidget* parent = nullptr);
+  ~CustomFormatWidget() override;
+  QStringList getFields() const;
+  void setFields(QString, QString, QString, QString, QString);
```
cryos: Why isn't this a QStringList?

tomviz/acquisition/CustomFormatWidget.h
```diff
@@ -0,0 +1,45 @@
+#ifndef tomvizCustomFormatWidget_h
```
cryos: Missing copyright header

tomviz/acquisition/MatchInfo.h
```diff
@@ -0,0 +1,22 @@
+#ifndef tomvizMatchInfo_h
```
cryos: Missing copyright header

tomviz/acquisition/MatchInfo.h
```diff
@@ -0,0 +1,22 @@
...
+#define tomvizMatchInfo_h
+
+#include <QList>
+
+namespace tomviz {
+
+struct CapGroup
+{
+  CapGroup(QString name_, QString text) : name(name_), capturedText(text) {}
```
cryos: Why not const ref for name/text?

tomviz/acquisition/PassiveAcquisitionWidget.h
```diff
@@ -16,11 +16,15 @@
 #ifndef tomvizPassiveAcquisitionWidget_h
 #define tomvizPassiveAcquisitionWidget_h
 
+#include <QDialog>
+
+#include "MatchInfo.h"
+
 #include <QLabel>
+#include <QMap>
```
cryos: Not sure why this was added, don'

cryos:
I would still like to see a sample file for the default file name format, instead it asks me to paste a sample. I tried my_p4.dm3 and it didn't do anything, I am not sure how it is supposed to run in the passive data acquisition dialog.

tomviz/acquisition/PassiveAcquisitionWidget.cxx
```diff
@@ -444,38 +435,89 @@ void PassiveAcquisitionWidget::checkEnableWatchButton()
...
+  validateTestFileName();
 }
 
-void PassiveAcquisitionWidget::stopWatching()
+void PassiveAcquisitionWidget::onRegexChanged(QString regex)
 {
-  m_watchTimer->stop();
-  m_ui->stopWatchingButton->setEnabled(false);
-  m_ui->watchButton->setEnabled(true);
+  qDebug() << regex;
```
cryos: This looks like leftover debug

cryos:
Looks better with the latest commit, but still surprising that when I open the dialog it is populated, the text entry is green, but when I make changes it turns an orange color and doesn't show me the result in the table below. Only custom seems to be evaluating the string as I change it.

alesgenova:
The orange color indicates that the filename doesn't match the current regex, and as a consequence the table has no captured text to show. Each time the filename changes, the match is reevaluated and UI components should be updated as needed.
I think that the example you provided ("my_p4.dm3") doesn't match the default options because it expects something like "my_p4degree.dm3". Let me know if that's not the case, and I can dig deeper

cryos:
I see now, I don't see a good reason to add the required suffix degree to dm3 and not tiff. Better that neither need the suffix degree, but that now makes more sense. I think our default should just be <prefix>[n|p]<angle><suffix>.*. We have a bit of space on this, and I think I would rather have a file name format, or an extension combo box. Switching to custom could just grey out the suffix.

We could then just offer a couple of format, and the extensions which are much easier to see when specified separately. We could also just hide the tif[f] by offering TIFF, DM3, etc as valid extensions we support.

PR #1593, opened 2018-05-28
Remove the deep copy of the image for data sources

cryos:
The data source has always performed a deep copy of the image data read
by the reader/from other sources. I don't think this is necessary, and
in the cases of large image data can be quite costly. In my superficial
testing I am not seeing any regressions, I think it is reasonable to
expect data source to take strong ownership of the image data passed in.


cjh1:
@cryos Did you test this with a reconstruction? (i.e. a child data source)

Issue #1565, opened 2018-05-04
Transparency for slice tools (opacity transfer function)


cophus:
Slices (both orthogonal and arbitrary directions) do not support the opacity transfer function (of intensity) the way the volume render does.  This feature would be extremely useful for us!  

<img width="1717" alt="screenshot 2018-05-04 16 01 31" src="https://user-images.githubusercontent.com/796309/39656414-73498a16-4fb4-11e8-9913-a7144519ebcf.png">

cryos:
Just to make sure I am understanding, in this example you have zero opacity assigned to the blue and blue/purple, so you would want them to be transparent, with increasing opacity as you moved into the red. Some level of per-pixel opacity mapping for slices?

cryos:
So, in a surprising twist, after asking on our VTK developer list, this is already implemented, and it looks great (to my untrained eyes at least). Should this be enabled by default? Should we add a toggle for it? They are probably going to take most of the development time. The color transfer function also then reflects the opacity mapping.
![image](https://user-images.githubusercontent.com/42214/40624198-1f0f87ca-6279-11e8-9b4d-82a60e721236.png)

ercius:
My opinion is to not enable this by default. I want to see the full opaque
slice, but have the option to enable this if wanted. This is definitely a
useful feature though!

mathturtle:
I don't like it on the histogram, but that is just me.  I feel like the histogram's coloring shouldn't change as you adjust opacity so that you to see the counts of values that are currently transparent.

ercius:
Most users I show tomviz do not understand the opacity. They don't quite understand that you move the linear line on the histogram which makes voxels more or less transparent. I think that having the histogram also look transparent would get this point across more easily.

Also, people do not understand that the transparency currently only affects volume renders. The addition of mapping transparency should help with this.

cryos:
I share your feeling on this @ercius, you accurately see what will be rendered, and what the color ends up looking like. The question kind of remains for me though - should we globally enable this, and then manipulate the color map (possibly detached) to be opaque by default? I think that would give a better preview, but may be confusing - you could still see the opaque slice by setting all opacity to 1.0.

cryos:
We could also get slightly fancy and show a gray histogram behind the main one that would show populations not being rendered.

cryos:
[...]

PR #1594, opened 2018-05-28
Add toggle to make slices honor the colormap opacity

cryos:
This fixes issue #1565, setting the default to enable opacity mapping
for slices, etc. This affects the look of the color-opacity transfer
function histogram too, yielding a truer preview of the combined color
and opacity transfer.

### DIFF
--- a/tomviz/modules/ModuleSlice.cxx
+++ b/tomviz/modules/ModuleSlice.cxx
@@ -232,11 +232,16 @@ void ModuleSlice::addToPanel(QWidget* panel)
 
   QVBoxLayout* layout = new QVBoxLayout;
 
+  m_opacityCheckBox = new QCheckBox("Color Map Opacity");
+  layout->addWidget(m_opacityCheckBox);
+
   QCheckBox* mapScalarsCheckBox = new QCheckBox("Color Map Data");
   layout->addWidget(mapScalarsCheckBox);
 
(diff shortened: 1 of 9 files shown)


tomviz/modules/ModuleSlice.cxx
```diff
@@ -231,8 +232,20 @@ void ModuleSlice::addToPanel(QWidget* panel)
 
   QVBoxLayout* layout = new QVBoxLayout;
 
+  m_opacityCheckBox = new QCheckBox("Color Map Opacity");
```
cryos: I think this is just "Map Opacity"

tomviz/modules/ModuleSlice.h
```diff
@@ -44,6 +45,7 @@ class ModuleSlice : public Module
   QJsonObject serialize() const override;
   bool deserialize(const QJsonObject& json) override;
   bool isColorMapNeeded() const override { return true; }
+  bool isColorMapOpaque() const override { return m_opaqueMap; }
```
cryos: I think this naming is odd, maybe more isOpacityMapped would get across what we mean, when enabled it could be fully opaque, or fully transparent, whatever the color map asks for,

cryos: m_mapOpacity would probably be simpler/convey the meaning more clearly.

PR #1606, opened 2018-06-05
Refactor python reader/writer registration

cjh1:
Registration is now controlled by a singleton rather static classes. The primary motivation was to avoid a static destruction issue.

### DIFF
--- a/tomviz/LoadDataReaction.cxx
+++ b/tomviz/LoadDataReaction.cxx
@@ -116,7 +116,6 @@ void LoadDataReaction::onTriggered()
 
 QList<DataSource*> LoadDataReaction::loadData()
 {
-  FileFormatManager::instance().registerPythonReaders();
   QStringList filters;
   filters << "Common file types (*.emd *.jpg *.jpeg *.png *.tiff *.tif *.raw"
              " *.dat *.bin *.txt *.mhd *.mha *.vti *.mrc *.st *.rec *.ali "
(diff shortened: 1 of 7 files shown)


tomviz/LoadDataReaction.cxx
```diff
@@ -181,7 +180,6 @@ DataSource* LoadDataReaction::loadData(const QString& fileName,
 DataSource* LoadDataReaction::loadData(const QStringList& fileNames,
                                        const QJsonObject& options)
 {
-  LoadDataReaction::registerPythonReaders();
```
alesgenova: You have to make another call to `FileFormatManager::instance().registerPythonReaders();` 

We could get here without passing through the open file dialog (e.g. from recent files or image stack dialog), so it's important to ensure the python readers are loaded.

(Although since we preemptively register them in the main window we would never really run into issues in practice)

tomviz/SaveDataReaction.cxx
```diff
@@ -150,8 +149,11 @@ bool SaveDataReaction::saveData(const QString& filename)
       updateSource(filename, source);
       return true;
     }
-  } else if (m_pythonExtWriterMap.contains(info.suffix().toLower())) {
-    auto writer = m_pythonExtWriterMap[info.suffix().toLower()]->createWriter();
+  } else if (FileFormatManager::instance().pythonWriterFactory(
+               info.suffix().toLower()) != nullptr) {
+    auto factory = FileFormatManager::instance().pythonWriterFactory(
+      info.suffix().toLower());
```
alesgenova: Add `Q_ASSERT(factory != nullptr);` check like in the reader case

Issue #1596, opened 2018-05-30
Contour default value can freeze tomviz for long periods


ercius:
Currently, the Contour default value is set to half of the histogram range. This can produce very long computations with no feedback to the user. I suggest 1 or both of the following:
1) Add a progress bar for this computation
2) Add a dialog box to allow the user to pick the value before the calculation proceeds

cryos:
Ideally we should actually thread the calculation of the contour, then it would take place in a background thread without freezing the main UI. We had an experimental feature doing this with streaming using Dax way back, without moving it to a background thread it will be tough to show progress. Popping up a dialog would be relatively easy.

cryos:
@psavery @cjh1

cryos:
Digging deeper, there was a huge performance regression where the resample filter is running unconditionally. It causes memory use spikes too, and should have only ever run for segmented/labeled data. Removing that gets rid of the performance issue, and enables much faster contouring. We will be simplifying the contour module, and removing that first.

cryos:
The regression is fixed, I think we may want to think a little about better use of background threads for some of this, or confirmation dialogs, but with typical data sizes this seems less critical now.

Issue #1595, opened 2018-05-30
Double clicking on histogram produces contour


ercius:
If you double click on the histogram then a contour is created with its value set to where you double click. I think this functionality is not necessary and cumbersome for some users. In some cases the user double clicks on the histogram by accident and then the computer freezes for some time while calculating the contour. 

I suggest removing this functionality or at least providing a pop-up dialog with the prefilled value and ok/cancel to create a contour.

cryos:
I think there is value in this, Rob and I worked on it early on. It would probably be good to confirm at least the first time they want to do this, with a possibility of suppressing it in future clicks.

cjh1:
@psavery Can add a confirmation dialog for the double click event? I think that and the other improvements you have made should resolve this issue.

ercius:
I think that is a good idea. It at least makes it clear to the user what
happened.

psavery:
Will do.

psavery:
How does something like this look?

![im3](https://user-images.githubusercontent.com/9558430/43160715-f8515002-8f53-11e8-8350-d1eb8b2ab471.png)

When the user double-clicks on the histogram, this dialog would appear instead of the one in PR #1618. The double slider widget is only enabled if "Create contour?" is checked.

If the user checks 'Don't ask again', it will from now on do one of the following:
1. If "Create contour?" was checked: always create a contour when the user double-clicks the histogram.
2. If "Create contour?" was not checked: never create a contour when the user double-clicks the histogram.

If the user clicks the "x", the contour is not created and no settings are saved.

ercius:
I suggest changing this in the following ways:
1. Use an Ok/Cancel dialog box. Then the user can easily cancel if it
happened by accident. I see many new users accidentally create contours
this way.
2. Change the title to Create contour
3. Remove the "create contour?" check box. It's unnecessary.
4. Remove the "don't ask again" option. How would the user reenable this if
they wanted the option back? (Otherwise keep it if there is a way to
reenable it)
5. The slider should always be enabled

Lastly, can there be a live preview option check box? Or will that be
prohibitively slow? I think the contour algorithm is faster now.

cryos:
I think we need to offer a don't ask again, I and many others find these dialogs very annoying. Also, I would rather not add a live preview - you can already do that in the panel, duplicating here is not necessary. Agreed on removing the create contour - with ok/cancel it is duplicating things.

cjh1:
[...]

PR #1616, opened 2018-06-07
Removed resampling from the contour module

psavery:
Resampling was noticeably slowing down contour operations
(and it was running even without segmentation), so it has been removed.

"Color By" was removed from the GUI Contour panel as it is no longer
needed. Any functions and variables that appeared to be no longer needed
were removed as well.

### DIFF
--- a/tomviz/modules/ModuleContour.cxx
+++ b/tomviz/modules/ModuleContour.cxx
@@ -47,7 +47,6 @@
 #include <vector>
 
 #include <QCheckBox>
-#include <QComboBox>
 #include <QFormLayout>
 #include <QJsonArray>
 #include <QJsonObject>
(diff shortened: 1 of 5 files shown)


psavery:
Here is an image of what the contour panel now looks like with the GUI changes. Let me know if anything should be changed.

contour-panel

psavery:
I tried to remove as many variables and functions that are no longer used as I could (so that the module is cleaner).

Let me know if I may have removed something I shouldn't have (although, as far as I can tell, the GUI functions properly), or if I haven't removed something I should have.

psavery:
I had accidentally pushed my branch to openchemistry, which is why I assume there is an extra AppVeyor test on display here. But I deleted that extra branch. So it should be good now.

cryos:
This addresses performance issues pointed out by #1595 and #1596.

PR #1617, opened 2018-06-07
Added a function to DataSource to obtain the range

psavery:
I plan to use this function for the contour range selector that appears when a user begins a contour calculation. I tested it and it seems to work fine.

I mostly followed the code [here](https://github.com/OpenChemistry/tomviz/blob/master/tomviz/CentralWidget.cxx#L67) to do this. I also followed the same pattern as the functions above it (i.e., it takes a double[] as its argument).

Let me know if I should make any changes (including something like returning a QVector\<double\> instead of using the double[2] parameter).

### DIFF
--- a/tomviz/DataSource.cxx
+++ b/tomviz/DataSource.cxx
@@ -480,6 +480,24 @@ void DataSource::getBounds(double bounds[6])
   }
 }
 
+void DataSource::getRange(double range[2])
+{
+  vtkAlgorithm* tp = algorithm();
+  if (tp) {
+    vtkImageData* data = vtkImageData::SafeDownCast(tp->GetOutputDataObject(0));
+    if (data) {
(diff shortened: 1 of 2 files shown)


tomviz/DataSource.cxx
```diff
@@ -480,6 +480,25 @@ void DataSource::getBounds(double bounds[6])
   }
 }
 
+void DataSource::getRange(double range[2])
+{
+  vtkAlgorithm* tp = algorithm();
+  if (tp) {
+    vtkImageData* data = vtkImageData::SafeDownCast(tp->GetOutputDataObject(0));
+    if (data) {
+      vtkSmartPointer<vtkDataArray> arrayPtr =
```
cryos: There is no need to use a smart pointer here, see the getSpacing code.

psavery:
I removed the smart pointer.

PR #1618, opened 2018-06-08
Added dialog to select initial contour value

psavery:
Now, when a contour is being initialized, a dialog box will appear
asking the user to confirm the initial contour value. The new default
value for the contour is 2/3 of the range instead of 1/2.

If the user initializes the contour by double-clicking on the histogram,
the value at which they double-clicked will be the default value.

The user may also check "Don't ask again" if they do not wish to
see this dialog box again.

### DIFF
--- a/tomviz/DataSource.cxx
+++ b/tomviz/DataSource.cxx
@@ -73,6 +73,7 @@ public:
   PersistenceState PersistState = PersistenceState::Saved;
   bool UnitsModified = false;
   bool Forkable = true;
+  double initialContourValue = DBL_MAX;
 
   // Checks if the tilt angles data array exists on the given VTK data
   // and creates it if it does not exist.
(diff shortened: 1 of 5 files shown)


psavery:
Here is what the initial contour dialog looks like:
initialcontourvaluedialog

Note that there is no option to cancel. Since we are generating this during the initialization of the contour module, we can't currently cancel and exit. If we want to add a cancel button, we will need to put this dialog somewhere else other than in the initialization of the contour module.

cryos:
This is where independent pull requests can be an issue, there is a conflict that needs resolving now. Keep it in a separate commit, but it would have been fine to put the simplification of the setIsoValue member into this branch.

psavery:
Whoops. I forgot that I had added "getIsoValue()" here, and I assumed the two branches would be merge-able. Hopefully it is fixed now...

PR #1626, opened 2018-06-12
Map scalars to colors when exporting slices as images

### LINKED ISSUE #1608: Export slice as image usability
Having the ability to export ortho and non-ortho slices in Tomviz is really useful. However....

1) The ability is buried in the right click menu of the object in the pipeline. Would this be better placed in the Properties of the object as a GUI button?
2) The default option to export to is PNG. But PNG only allows 8-bit output (it warns you). Should we change this to TIF so that users dont get weird output the first time?
3) The exported data does not have the colormap applied. However, the right click menu says "Export as Image" and thus I expect to get the same thing that I see in the tomviz window (colormap applied, 8bit).
4) The data exported is probably the raw data from the interpolation or array slicing in VTK. I personally like having this option, but its not an "Image". We could allow the user to export an image or the raw data.

alesgenova:
Fixes #1608

Works for both OrthogonalSlice and Slice.
If "Color Map Data" is selected in the slice property panel, in the
resulting output image scalars will be mapped to colors.

If "Color Map Data" is de-selected, the raw data will be written to
image (with potential conversion from float/double to unsigned char).

Opacity also works as expected for the image formats that support it.

As a result of this commit, the exported image will always end up
looking like the slice being drawn in the program.

slice
alpha_slice
slice_nomap

### DIFF
--- a/tomviz/ExportDataReaction.cxx
+++ b/tomviz/ExportDataReaction.cxx
@@ -27,12 +27,15 @@
 #include <pqSettings.h>
 #include <vtkArrayCalculator.h>
 #include <vtkDataArray.h>
+#include <vtkImageCast.h>
 #include <vtkImageData.h>
+#include <vtkImageMapToColors.h>
 #include <vtkNew.h>
 #include <vtkPointData.h>
 #include <vtkSMProxyManager.h>
 #include <vtkSMSessionProxyManager.h>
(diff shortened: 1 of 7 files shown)


cryos:
Looks good, if the data is not color mapped I get an error about unsupported type. It would be much more intuitive to default to TIFF when the color is not being mapped.

PR #1630, opened 2018-06-14
Add support for external pipelines with child data sets

cjh1:


cjh1:
Will need to adapt to this change: #1814

cjh1:
#1812

cjh1:
This is ready to people to try out and review. I think we might want to hold of merging until we have release a point release for 1.7?

Issue #1486, opened 2018-04-17
Open a modal dialog while state is restored


cryos:
The state is restored in multiple steps now, this means that someone canceling operators initiated by loading state can cause undefined behavior. I think we should probably pop up a modal dialog, longer term we should also offer the option to cancel but that may take longer.

Issue #1302, opened 2017-10-18
State file does not save the 2D gradient transfer function for volume rendering


yijiang1:
Nightly build 1.1.0-115-g14f1e3c

cryos:
On Linux I am seeing a new bug with the 2D gradient transfer function, any adjustment in our widget causes the rendering to disappear, and a failure to compile the GLSL shader.
```
ERROR: In /home/marcus/src/paraview/VTK/Rendering/OpenGL2/vtkShaderProgram.cxx, line 462
vtkShaderProgram (0x55c8e7bb6d50): 0(165) : error C1008: undefined variable "g_gradients_0"
0(223) : error C1008: undefined variable "g_gradients_0"
0(233) : error C1008: undefined variable "g_gradients_0"
0(476) : error C1008: undefined variable "g_gradients_0"
```
It looks like something shifted in VTK, I will dig into it, but without the ability to modify the restoration of the state will not work.

mathturtle:
@cryos I'm going to start updating my branch for this unless there is something else you'd rather I worked on.  IIRC, you wanted it refactored so there is a TransferFunction2D object that contains the boxes etc. rather than passing around a list of box items.

cryos:
There is only one box, I think this should be a new branch (we can abandon the old one). It should expose a set and a get on a vtkRect, along with with restoring the mode to transfer 2D correctly. Let's go for minimal extra code/classes, enough to enable setting and getting the single rectangle we support. If we augment this in future, the state can also be augmented then.

cryos:
This should be resolved now, we can open new issues to track further improvements.

PR #1631, opened 2018-06-14
Save 2D transfer functions

mathturtle:
This was a bigger change than I wanted... but it was necessary to get this to work.  This branch:

* Splits the transfer function 2D so that there is one per data source and a module with separate color map has its own
* Supports saving the transfer function 2D to the state file (adds the 4 numbers for the transfer function box)
* Refactors the histogram creation so that it can be called from the volume module to get histograms for non-active DataSources/Modules
* Refactors the generate transfer function 2D so that it can be called from the volume module if the transfer function is not initialized and the histogram is done
* Supports loading volumes with transfer function 2Ds from the state file

Note the refactors were necessary for loading the transfer function 2d from the state.

### DIFF
--- a/tomviz/DataSource.cxx
+++ b/tomviz/DataSource.cxx
@@ -71,6 +71,7 @@ public:
   vtkSmartPointer<vtkStringArray> Units;
   vtkVector3d DisplayPosition;
   PersistenceState PersistState = PersistenceState::Saved;
+  vtkRectd m_transferFunction2DBox;
   bool UnitsModified = false;
   bool Forkable = true;
   double initialContourValue = DBL_MAX;
@@ -895,6 +896,11 @@ vtkImageData* DataSource::transferFunction2D() const
   return this->Internals->m_transfer2D;
 }
 
+vtkRectd* DataSource::transferFunction2DBox() const
+{
+  return &this->Internals->m_transferFunction2DBox;
+}
+
 bool DataSource::hasLabelMap()
 {
   auto dataSource = proxy();
@@ -940,6 +946,10 @@ void DataSource::init(vtkImageData* data, DataSourceType dataType,
   this->Internals->Type = dataType;
   this->Internals->PersistState = persistState;
   this->Internals->DisplayPosition.Set(0, 0, 0);
+
+  // Set up default rect for transfer function 2d.
+  // These were the default box params in the UI.
+  this->Internals->m_transferFunction2DBox.Set(1, 1, 19, 19);
 
   vtkNew<vtkSMParaViewPipelineController> controller;
   auto pxm = ActiveObjects::instance().proxyManager();
--- a/tomviz/DataSource.cxx
+++ b/tomviz/DataSource.cxx
@@ -329,6 +329,13 @@ QJsonObject DataSource::serialize() const
   // Serialize the color map, opacity map, and others if needed.
   json["colorOpacityMap"] = tomviz::serialize(colorMap());
   json["gradientOpacityMap"] = tomviz::serialize(gradientOpacityMap());
+  QJsonObject boxJson;
+  auto& transfer2DBox = this->Internals->m_transferFunction2DBox;
+  boxJson["x"] = transfer2DBox.GetX();
+  boxJson["y"] = transfer2DBox.GetY();
+  boxJson["width"] = transfer2DBox.GetWidth();
+  boxJson["height"] = transfer2DBox.GetHeight();
+  json["colorMap2DBox"] = boxJson;
 
   // Serialize the operators...
   QJsonArray jOperators;
@@ -366,6 +373,13 @@ bool DataSource::deserialize(const QJsonObject& state)
   if (state.contains("gradientOpacityMap")) {
     tomviz::deserialize(gradientOpacityMap(),
                         state["gradientOpacityMap"].toObject());
+  }
+  if (state.contains("colorMap2DBox")) {
+    auto boxJson = state["colorMap2DBox"].toObject();
+    auto& transfer2DBox = this->Internals->m_transferFunction2DBox;
+    transfer2DBox.Set(boxJson["x"].toDouble(), boxJson["y"].toDouble(),
+                      boxJson["width"].toDouble(),
+                      boxJson["height"].toDouble());
   }
 
   if (state.contains("spacing")) {
@@ -947,9 +961,10 @@ void DataSource::init(vtkImageData* data, DataSourceType dataType,
   this->Internals->PersistState = persistState;
   this->Internals->DisplayPosition.Set(0, 0, 0);
 
-  // Set up default rect for transfer function 2d.
-  // These were the default box params in the UI.
-  this->Internals->m_transferFunction2DBox.Set(1, 1, 19, 19);
+  // Set up default rect for transfer function 2d...
+  // The widget knows to interpret a rect with negative width as
+  // uninitialized.
+  this->Internals->m_transferFunction2DBox.Set(0, 0, -1, -1);
 
   vtkNew<vtkSMParaViewPipelineController> controller;
   auto pxm = ActiveObjects::instance().proxyManager();
--- a/tomviz/Histogram2DWidget.cxx
+++ b/tomviz/Histogram2DWidget.cxx
@@ -112,9 +112,9 @@ void Histogram2DWidget::addFunctionItem(
   m_chartHistogram2D->AddFunction(item);
 }
 
-void Histogram2DWidget::setTransfer2D(vtkImageData* transfer2D)
+void Histogram2DWidget::setTransfer2D(vtkImageData* transfer2D, vtkRectd* box)
 {
-  m_chartHistogram2D->SetTransfer2D(transfer2D);
+  m_chartHistogram2D->SetTransfer2D(transfer2D, box);
   m_histogramView->Render();
 }
 
--- a/tomviz/Utilities.cxx
+++ b/tomviz/Utilities.cxx
@@ -39,12 +39,10 @@
 #include <vtkBoundingBox.h>
 #include <vtkCamera.h>
 #include <vtkColorTransferFunction.h>
-#include <vtkFloatArray.h>
 #include <vtkImageData.h>
 #include <vtkImageSliceMapper.h>
 #include <vtkNew.h>
 #include <vtkPiecewiseFunction.h>
-#include <vtkPointData.h>
 #include <vtkPoints.h>
 #include <vtkRenderer.h>
 #include <vtkSmartPointer.h>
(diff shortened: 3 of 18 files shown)


mathturtle:
Fixes #1302.

cjh1:
@mathturtle There are some clang-format issues to clean up.

cjh1:
@mathturtle There are still some issues.

mathturtle:
Huh?  I ran this: `git diff -U0 master | clang-format-diff-6.0 -i -p1`.  I guess it is the clang-format version then...  Just a second, I'll get a machine that still has 3.8.

cjh1:
@mathturtle Still having problems with this, I have moved my ParaView to tomviz/release. On tomviz master I see:

before
With this branch I can't make a selection in the transfer function:
after

mathturtle:
@cjh1 I think I've fixed the memory corruption with the 'in progress' state for histograms and preventing the foreground thread from accessing them until they are done.

cjh1:
@mathturtle clang-format?

cjh1:
@mathturtle If I switch the active volume module during loading then the transfer function never seems to get rendered?

mathturtle:
So after talking to @cjh1 this has expanded to also fix #1486.  There wasn't really another good way around the bug he found in the last version.  The dialog it throws up is pretty basic, just a "Please wait... loading state file".  It lasts until all histograms are computed and all pipelines are finished, which may not cover everything load state does but should be most of it.  Do we want to make the dialog fancier?

tomviz/SaveLoadStateReaction.cxx
```diff
@@ -110,11 +114,34 @@ bool SaveLoadStateReaction::loadState(const QString& filename)
...
+  auto layout = new QHBoxLayout;
+  auto label =
+    new QLabel("Please wait... loading state file in progress", &dialog);
+  layout->addWidget(label);
+  dialog.setLayout(layout);
+  dialog.show();
+  // At least locally I need 2 processEvents calls for the dialog to show
+  // and appear non-blank.
+  QApplication::processEvents();
+  QApplication::processEvents();
```
cjh1: Even with these it doesn't seem to show up for me.

tomviz/SaveLoadStateReaction.cxx
```diff
@@ -110,11 +114,34 @@ bool SaveLoadStateReaction::loadState(const QString& filename)
...
   if (doc.isObject() && ModuleManager::instance().deserialize(
                           doc.object(), QFileInfo(filename).dir())) {
     RecentFilesMenu::pushStateFile(filename);
+    // Volume modules with 2D transfer functions need the histograms
+    // finsished before the user can interact.
+    while (HistogramManager::instance().hasPendingHistograms() ||
+           ModuleManager::instance().hasRunningOperators()) {
+      QApplication::processEvents();
+    }
+    dialog.hide();
```
cjh1: Think it would be cleaner to exec the dialog and the have signals produced by the HistogramManager and ModuleManager which we could connect to to close the dialog.

tomviz/modules/Module.cxx
```diff
@@ -197,6 +212,13 @@ QJsonObject Module::serialize() const
     if (m_useDetachedColorMap) {
       json["colorOpacityMap"] = tomviz::serialize(d->detachedColorMap());
       json["gradientOpacityMap"] = tomviz::serialize(gradientOpacityMap());
+      QJsonObject boxJson;
+      auto transfer2DBox = d->detachedTransferFunction2DBox();
+      boxJson["x"] = transfer2DBox->GetX();
+      boxJson["y"] = transfer2DBox->GetY();
+      boxJson["width"] = transfer2DBox->GetWidth();
+      boxJson["height"] = transfer2DBox->GetHeight();
+      json["colorMap2DBox"] = boxJson;
```
cjh1: Seem like this is duplicated from DataSource?

tomviz/DataSource.cxx
```diff
@@ -894,6 +909,11 @@ vtkImageData* DataSource::transferFunction2D() const
   return this->Internals->m_transfer2D;
 }
 
+vtkRectd* DataSource::transferFunction2DBox() const
```
cryos: Why a pointer, rather than return by value?

tomviz/Histogram2DWidget.cxx
```diff
@@ -112,9 +112,9 @@ void Histogram2DWidget::addFunctionItem(
   m_chartHistogram2D->AddFunction(item);
 }
 
-void Histogram2DWidget::setTransfer2D(vtkImageData* transfer2D)
+void Histogram2DWidget::setTransfer2D(vtkImageData* transfer2D, vtkRectd* box)
```
cryos: Would have expected vtkRectd to be passed as a const reference

tomviz/SaveLoadStateReaction.cxx
cryos: This is unlikely to work as expected on all OSes, and we need to structure this so that is uses signals and slots. It might be better to split off the modal dialog from this PR, but I don't think this is the right way of adding a modal dialog that loops on processEvents until it thinks everything is complete.

tomviz/SaveLoadStateReaction.cxx
cryos: Agreed.

tomviz/Utilities.cxx
```diff
@@ -762,6 +764,77 @@ QWidget* mainWidget()
   return pqCoreUtilities::mainWidget();
 }
 
+void rasterTransferFunction2DBox(vtkImageData* histogram2D, const vtkRectd& box,
```
cryos: Why add this to Utilities? It seems very histogram specific. Also, I don't see where this was moved from/deletion of equivalent code. Maybe I am missing something here?

tomviz/vtkChartTransfer2DEditor.cxx
```diff
@@ -48,14 +51,48 @@ vtkStandardNewMacro(vtkChartTransfer2DEditor)
...
+  // Set the box first, GenerateTransfer2D writes to it
+  // so avoid writing to the old one.
+  if (box != nullptr) {
+    this->Transfer2DBox = box;
+  } else {
+    // Make sure to use internal dummy box.  This is set to
+    // nullptr when the active datasource/module is cleared
+    // which usually means it was deleted.  Avoid pointers to
+    // deleted data by setting this to a local dummy value.
+    this->Transfer2DBox = &this->DummyBox;
```
cryos: The vtkRect classes are simple, and should generally be setting the values of the object, rather than passing around/setting the pointer to a single object. It is just 4 doubles, and the use of pointers looks overly complex when a simpler set/get paradigm could be used avoiding the headaches of memory management.

tomviz/SaveLoadStateReaction.cxx
mathturtle: Yes, it seems that is the right way to go.  This way the dialog inconsistently shows up across machines. :(

tomviz/modules/Module.cxx
mathturtle: It didn't seem worth it to make separate functions to store/load 4 values.

mathturtle:
> prefer just amending to the commit the code was changed in

Didn't you just ask me NOT to amend commits once they were put up for review?  Over here: https://github.com/OpenChemistry/tomviz/pull/1579

tomviz/vtkChartTransfer2DEditor.cxx
mathturtle: The widget has to be able to edit the box in the DataSource/Module the same way it edits the transfer function.  Thus the pointers.  I think they should all be wrapped in a transfer function 2D object, but you didn't like that approach.

I could change it to emit some vtk event when the box changes and hook a Qt signal to that that updates the other box.  But managing what gets updated as the user changes the active objects will be just as complex as the pointers.

tomviz/DataSource.cxx
mathturtle: Because the widget needs to be able to update it the same way it updates the vtkColorTransferFunction*.

tomviz/Utilities.cxx
mathturtle: Oops.  It moved to Utilities because it needed it in ModuleVolume's load state.

mathturtle: I forgot to delete the original.

cryos:
> Once we start reviewing code it is a lot easier to see how you fixed something by adding new commits, rather than rewriting. It shows up easily in the GitHub UI, and I can just look at the last commit, this would be good to bear in mind in future.

If this wasn't clear, I meant in the context of responding to review on actual code. There is no value in me or anyone reviewing fixing of formatting changes. Hopefully the difference between the two is clear, I want to see how you changed something functionally in response to code review, but applying formatting changes is just extra noise and the whole reason for automating formatting was to move away from reviewing code formatting.

mathturtle:
Yes, that is clear.  Amending for style is fine, but for requested changes is not.

I'm in favor of splitting this into two branches, one for 2d transfer functions and one for the modal dialog.  They don't really have anything to do with each other code-wise.

cryos:
Agreed, while you are splitting the two, please edit to make the actual move of code clearer for 2D transfer (by deleting the original), integrate code style fixes. It would be far better to move that code to the histogram specific classes rather than the Utilities. Ideally all commits as originally pushed should have been run through clang-format before anyone sees them, if things creep through like I said there is no value in reviewing code style fixes and it makes history more difficult to navigate.

mathturtle:
* Rebased on master
* Every commit clang-formatted with 6.0 and the 3.8 fixup done at the end (rather than the two fighting every commit)
* Extra bugfix commits from working with Chris squashed in
* Function moved to Utilities deleted from original location

You think the `rasterTransfer2DBox` function fits better somewhere other than Utilities?  The only other place I can think of is to put it back in `vtkChartTransfer2DEditor` or `vtkTransferFunctionBoxItem` as a static function, but then `ModuleVolume` has to know about unrelated UI classes.

Actually the more I think about it the more I like putting it in `vtkTransferFunctionBoxItem`... new version coming up.

mathturtle:
> Actually the more I think about it the more I like putting it in vtkTransferFunctionBoxItem... new version coming up.

And done.

mathturtle:
> but packaging the image data with the vtkRect2d in a simple object looks like it would make sense to me

That is what I was trying to say... maybe I wasn't clear enough :(

I'd thought of a few other ways to organize it, but this was already hard enough to get the update right when you switch data sources (see all the comments in `vtkChartTransfer2DEditor::SetTransfer2D`).  The other ways would have been more complex and I feel like bundling the two is the right solution anyway.

Issue #684, opened 2016-10-14
"Duplicate Module" feature for visualizations


ElliotPadgett:
It would be nice if there were an option to duplicate a visualization module (next to the delete button in the right click menu, maybe).  I sometimes want to compare to similar but slightly different visualizations, and this would be a nice way to do that without creating them from scratch.

PR #1633, opened 2018-06-15
Add ability to duplicate modules

mathturtle:
I've tested all of the module types and I hope I've found all the crashes.  Please test anyway before approving!  Anything that updates its UI in deserialize crashes unless the UI access is null-guarded.

@alesgenova I've changed the UI elements to `QPointer`s, this fixes two issues: use-after free is impossible since the pointer will reset itself to null and the pointer is initialized to null in the constructor (it wasn't before, and one of my crashes was use of garbage data).

### DIFF
--- a/tomviz/PipelineView.cxx
+++ b/tomviz/PipelineView.cxx
@@ -332,11 +332,6 @@ void PipelineView::contextMenuEvent(QContextMenuEvent* e)
     hideAction = contextMenu.addAction("Hide");
     showAction = contextMenu.addAction("Show");
 
-    if (selectedIndexes().size() == 1) {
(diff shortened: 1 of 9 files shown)


tomviz/PipelineView.cxx
```diff
@@ -330,6 +332,11 @@ void PipelineView::contextMenuEvent(QContextMenuEvent* e)
     hideAction = contextMenu.addAction("Hide");
     showAction = contextMenu.addAction("Show");
 
+    if (selectedIndexes().size() == 1) {
```
alesgenova: The action is never added to the context menu, but removing the if condition makes it appear

mathturtle:
Fixes #684

Issue #1589, opened 2018-05-28
Set color map range dialog update


cryos:
The set range dialog for Tomviz should be updated, it should be a little more compact and just offer options relevant to Tomviz. The existing dialog comes from ParaView:
![image](https://user-images.githubusercontent.com/42214/40620121-ad3ce556-6265-11e8-8371-27b72db89b5f.png)
We should just use the standard OK and Cancel buttons, and I think a much shorter one line dialog, i.e. Range: then two boxes for min and max, we can retain the label with the -.

cryos:
This is fixed now.

PR #1634, opened 2018-06-18
Simplified the set range dialog

psavery:
There is now just one line for the range and the
standard OK and Cancel buttons.

Tooltips show the min and max values that can be set. The bottom
value in the range can never exceed the top value. And the top
value in the range can never be less than the bottom value.

The dialog box is fixed and cannot be resized.


psavery:
This is in response to issue #1589.

psavery:
Let me know if we should use QLineEdits instead of QDoubleSpinBoxes, or if anything else should change.

psavery:
Below is an example of what this now looks like.
example

PR #1635, opened 2018-06-18
Perform a more accurate automatic rescaling

psavery:
For some reason, it seems that the automatic rescaling
usually adds "0.5" to the min and max values in the data range.
For instance, if the range is 0 - 255 and we perform automatic
rescaling, it rescales it to 0.5 - 255.5.

This correctly rescales the data from 0 - 255 instead.


psavery:
I also noticed for our medium CoP nanoparticle example, that the previous automatic rescaling would rescale it to: 128.5 - 65663.5.

This PR rescales it 0 - 65535.

Issue #592, opened 2016-09-06
Center moved and reset button disabled


ercius:
tomviz 0.9.1, Windows 10
tomviz 0.9.1-101, WIndow 10

I see strange behavior with regards to the data center connected to the slice and orthoslice modules. Ive also seen this in previous builds but only just figured out how to reproduce it. Here is a set of instructions to create the issue.
1) Load the sample data reconstruction
2) Turn on Show Center. It is at the center of the data as expected
3) Click on the Slice module. The center is now at the edge and centered on the Slice
4) Notice that Reset Center button is grayed out
5) Click on Orthoslice module. Reset Center button is now enabled. The center is also now placed in the center of the data with the orthoslice.

There are other strange behaviors you can get with adding and deleting the Slice and Orthoslice modules with respect to the data center position and status of the Reset Center button. For example:
a) Place the center anywhere but the data center. Click on OrthoSlice or Slice modules and the center is set to the center of the generated slice.
or...
b) Load the example data set. Delete the automatically generated Orthoslice module. Notice that "Reset Center" is grayed out.

ercius:
This still occurs in 0.9.3. Interestingly, it occurs when you add a slice (non-orthogonal) and the arrow end is outside the coordinate system. The coordinate system increases to include the arrow end and then the rotation center is offset from the center of the data. Its a slightly annoying usability issue, but nothing major.

cquammen:
Developer's note: ParaView sets the enabled state of the "Reset Center" button based on whether a representation is active - when no representation is active, it is disabled, when a representation is active, it is enabled. ModuleSlice does not create a representation but instead adds a widget directly to the view's render window. As a result, when it is selected, no representation is active, and the "Reset Center" button is disabled.

cryos:
Good to know, we should override that behavior as we don't want to be restricted to representations. Thanks for the extra information.

psavery:
Hi Peter,

I'm looking at this bug now. However, I haven't been able to reproduce the change-of-center bugs on the latest Windows (or Ubuntu) version of Tomviz. My Windows version is Tomviz 1.3.1. Maybe the problem has been fixed? I am able to click on "slice", and the center axes do not move - even if I place the center axes somewhere else. If I click "orthogonal slice", the center axes increase in size, but the center does not move either. Indeed, left-clicking and dragging in the render window shows that it is still rotating around the displayed center axes as well.

If you get a chance, can you confirm that this no longer happens?

[...]

PR #1636, opened 2018-06-19
Fix disabling of 'Reset Center' button on Slice

psavery:
Before, the 'Reset Center' button would be disabled
when a slice module was initialized - even if the center
was not in the original center position.

This is a fix for the problem. The 'Reset Center' button
now remains enabled and behaves as expected.

### DIFF
--- a/tomviz/modules/ModuleSlice.cxx
+++ b/tomviz/modules/ModuleSlice.cxx
@@ -95,7 +95,7 @@ bool ModuleSlice::initialize(DataSource* data, vtkSMViewProxy* vtkView)
   controller->PreInitializeProxy(m_passThrough);
   vtkSMPropertyHelper(m_passThrough, "Input").Set(producer);
   controller->PostInitializeProxy(m_passThrough);
-  controller->RegisterPipelineProxy(m_passThrough);
+  controller->RegisterRepresentationProxy(m_passThrough);
 
   // Give the proxy a friendly name for the GUI/Python world.
   if (auto p = convert<pqProxy*>(proxy)) {


psavery:
Old (notice disabled "Reset Center" button):
reset-center-grayed-out
New (notice enabled "Reset Center" button):
fixed-reset-center-button

psavery:
This is to fix one of the problems mentioned in issue #592.

tomviz/modules/ModuleSlice.cxx
```diff
@@ -97,6 +97,9 @@ bool ModuleSlice::initialize(DataSource* data, vtkSMViewProxy* vtkView)
   controller->PostInitializeProxy(m_passThrough);
   controller->RegisterPipelineProxy(m_passThrough);
 
+  // We are calling this so that the "Reset Center" button will be enabled
+  controller->Show(m_passThrough, 0, vtkView);
```
mathturtle: This creates an outline representation by default.  So if you turn off the visibility of the outline module (or delete it), you still have the outline from this.

psavery: Hmm... yeah I see that. I will look into that.

psavery:
Unfortunately, setting the visibility of the pass through representation to 0 causes "Reset Center" to put the center in the corner of the cell instead of in the correct place.

psavery:
I found that if I changed the representation of the "PassThrough" to be that of "Volume" instead of "Outline" that it works as we would want it to work. The volume is invisible since there is no data associated with it. The "Reset Center" button is enabled, and it functions as we would expect. The current update to my commit reflects these changes.

However, this is still a fairly "hacky" solution. A clean or more-straightforward solution would probably be preferred.

mathturtle:
This works... but I'm also not sure if it is the preferred implementation or not.

psavery:
[Here](https://github.com/OpenChemistry/tomviz/blob/0a7f0e6307cec66cf9a3e2957105e6ef341bac4b/tomviz/modules/ModuleSlice.cxx#L98) is the specific line at which the reset button gets disabled. Note that it is the registering of the "PassThrough" filter in the pipeline. In theory, we may be able to keep the reset button from disabling if we were able to register the "PassThrough" filter without changing the vtk view.

The Scale Cube module and the Volume module do not register a pipeline proxy, so they do not run into the same problem of the reset button disabling.

If we could eliminate the usage of the pass through filter, that could fix the problem as well.

psavery:
I'm closing this in favor of eventually re-working the toolbar instead of using quick fixes to get around ParaView...

PR #1644, opened 2018-06-22
Better import/export of alignments

### LINKED ISSUE #1632: Extend ability to save and load alignment to automatic alignment operators

alesgenova:
Fix #1632 

The CenterOfMass, CrossCorrelation, and Manual alignment operators now return an OperatorResult that contains the alignments and appears in the PipelineView. The alignments can be saved to a json file and later be loaded from the ManualAlignment operator dialog.

screenshot from 2018-06-22 16-20-00

screenshot from 2018-06-25 09-55-15

### DIFF
--- a/tomviz/operators/OperatorResult.h
+++ b/tomviz/operators/OperatorResult.h
@@ -38,16 +38,16 @@ public:
   virtual ~OperatorResult() override;
 
   /// Set name of object.
-  void setName(const QString& name);
-  const QString& name();
+  void setName(QString name);
+  QString name() const;
 
   /// Set label of object.
-  void setLabel(const QString& label);
-  const QString& label();
+  void setLabel(QString label);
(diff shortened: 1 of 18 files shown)


cjh1:
@alesgenova This looking good!

tomviz/operators/OperatorResult.h
```diff
@@ -45,6 +45,10 @@ class OperatorResult : public QObject
   void setLabel(const QString& label);
   const QString& label();
 
+  /// Set description of object.
+  void setDescription(const QString& desc);
+  const QString& description();
```
cryos: It isn't worth returning as const ref, just return as a value. QString is copy-on-write.

cryos:
Also, can we adjust the UI to read "Save as JSON"

Issue #1646, opened 2018-06-25
Loading state files with multiple pipelines with python operators crashes


mathturtle:
Make a state file that has 2 data sources, each having a few python operators.  Save the state file out.  I had two operators on each of the two data sources.

Load in the state file just saved out.  Sometimes (5/6 times for me), there is a segfault deep in the guts of python when one of the operators finishes.

I can provide a sample state file, just come ask me.

PR #1647, opened 2018-06-26
Show a modal dialog while loading a state file

### LINKED ISSUE #1486: Open a modal dialog while state is restored

mathturtle:
This prevents several bugs caused by the user messing with stuff during
loading the state.  Fixes #1486.

@cjh1 I implemented it the way we discussed, with the count of how many pipelines still need to finish.  I've tested on a single pipeline with multiple segments.  I can't test multiple datasources with independent pipelines because of #1646.

### DIFF
--- a/tomviz/DataSource.cxx
+++ b/tomviz/DataSource.cxx
@@ -425,6 +425,12 @@ bool DataSource::deserialize(const QJsonObject& state)
                               QObject::disconnect(*connection);
                               delete connection;
                             });
+      // If the child datasource has its own pipeline of operators increment the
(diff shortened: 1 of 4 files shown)


mathturtle:
Now that the 2d transfer function branch is merged do you want to wait until this can updated it to wait for the histograms to complete too?  I started that yesterday afternoon, but it still has bugs.  Alternatively I can just make that a follow-up branch.

Issue #1637, opened 2018-06-19
Deleting pipeline elements often disables "Reset Center" and "Zoom To Data" buttons


psavery:
When deleting the most recent pipeline element, many modules will have an effect where the "Reset Center" and "Zoom To Data" buttons become disabled. These buttons are not re-enabled until a new module is created.

The modules where I have observed this issue:
Contour
Orthogonal Slice
Outline
Ruler
Slice
Threshold

The modules where this issue does not occur:
Scale Cube
Volume

psavery:
This was fixed via PR #1650 and PR #1655.

PR #1650, opened 2018-07-02
Combined camera toolbar with utilities toolbar

psavery:
This combines most actions from the camera toolbar with
the utilites toolbar. A few things were also done to clean
it up - such as removing the "Zoom to Data" button and creating
a drop-down menu for the "Reset view direction" buttons.

The CameraReaction class was copied from ParaView's "pqCameraReaction"
class with some modifications.

### DIFF
--- a/tomviz/CMakeLists.txt
+++ b/tomviz/CMakeLists.txt
@@ -27,6 +27,8 @@ set(SOURCES
   AlignWidget.h
   Behaviors.cxx
   Behaviors.h
+  CameraReaction.cxx
+  CameraReaction.h
   CentralWidget.cxx
   CentralWidget.h
   CloneDataReaction.cxx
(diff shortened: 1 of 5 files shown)


psavery:
This is mostly complete. The only thing that I want to change is that I think the "Reset view direction" menu is too wide. QMenu::QSizeHint() returns a width of about 120, and I can't figure out why. The width should be more like 60 (in my opinion). See image below.

screenshot

QMenu::actionGeometry() also shows that the actions are predicted to have a width of 120.

psavery:
I updated the commit to remove the icons (that we will be using in ParaView) and fix the clang-format.

psavery:
This is not quite finished yet. I have a few small changes to make, and then I will squash all the commits into one.

psavery:
This PR and PR #1655 fix issue #1637.

PR #1655, opened 2018-07-03
Combined axes toolbar with utilities toolbar

psavery:
This combines the axes toolbar with the utilites toolbar,
and it re-writes a little bit of the axes toolbar functionality
to make it more specific to Tomviz. One thing that was changed is
that the "Reset Center" button acts on the bounds of the active
data source, and it does not gray out unless there is no data.

### DIFF
--- a/tomviz/AxesReaction.cxx
+++ b/tomviz/AxesReaction.cxx
@@ -1,0 +1,182 @@
+/******************************************************************************
+
+  This source file is part of the tomviz project.
+
+  Copyright Kitware, Inc.
+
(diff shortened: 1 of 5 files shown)


psavery:
This seems to be complete, and it fixes issue #592. We may wish to modify some of the different settings in the future (the enabling/disabling of the various buttons, and what is the actual "center"), but it should be relatively easy since we have our own implementation of this now.

This will for sure have a conflict with PR ##1650 when one of these gets merged, but it should be relatively easy to fix.

PR #1657, opened 2018-07-03
Fix the unused compiler warnings with Q_UNUSED()

psavery:


cjh1:
@psavery These will be fixed by #1630 lets leave them for now to avoid merge conflicts.

psavery:
@cjh1 Okay sounds good.

PR #1668, opened 2018-07-13
Remove the scale legend from the view menu

### LINKED ISSUE #1661: Remove scale cube/bar from the view menu
After discussion we decided to remove them as their value was in question, will favor the in scene scale cube.

mathturtle:
After discussion with @Hovden it was determined to be not very useful,
so we are removing it.  Fixes #1661.


mathturtle:
Rebased in hopes of the internal compiler error going away :)

PR #1669, opened 2018-07-16
WIP: Wait to close the state loading dialog until histograms finish

mathturtle:
The volume module in 2d transfer function mode needs the histogram to be
complete before its state is fully loaded and user interaction during
the interval between the state loading and the histogram finishing can
cause its state to get messed up.  This delays closing the modal 'state
loading' dialog until the histograms are finished generating and the
volume module is in a consistent state.

### DIFF
--- a/tomviz/DataSource.cxx
+++ b/tomviz/DataSource.cxx
@@ -16,6 +16,7 @@
 #include "DataSource.h"
 
 #include "ActiveObjects.h"
+#include "HistogramManager.h"
 #include "ModuleFactory.h"
 #include "ModuleManager.h"
 #include "Operator.h"
@@ -395,6 +396,13 @@ bool DataSource::deserialize(const QJsonObject& state)
     auto units = state["units"].toString();
     setUnits(units);
   }
+
+  // Trigger the histograms to be generated.  These are needed for Volume
+  // modules to be correctly rendered in some cases.
+  vtkSmartPointer<vtkImageData> imageData =
+    vtkImageData::SafeDownCast(this->dataObject());
+  HistogramManager::instance().getHistogram(imageData);
+  HistogramManager::instance().getHistogram2D(imageData);
 
   // Check for modules on the data source first.
   if (state.contains("modules") && state["modules"].isArray()) {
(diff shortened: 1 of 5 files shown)


cryos:
This isn't going to work as it really degrades the experience for all state files, if I have a simple outline and a slice I see everything I need to, and the whole point of pushing the histogram calculations to a background thread is to enable interaction while they are being computed.

I am also not certain what harm is being prevented in this case, as the interaction before seems far better to me as an end user. We may want to move this to an issue, but from my point of view I have a few use cases, and playing with MEDIUM to give some processing time:

**No volume rendering**: none of the histograms have any impact, why would I wait for them.

**Before this PR**: volumes with scalar show up fine, with 2D they show up as scalar until the histogram is computed, then switch to 2D, seems fine. Changing the volume renderer state while waiting didn't cause any apparent issues or crashes, once the 2D histogram was ready my box was restored, and I could look at the 2D histogram.

**After this PR**: I wait a really long time, while apparently nothing is happening, can't interact.

> user interaction during the interval between the state loading and the histogram finishing can cause its state to get messed up

What things should we be looking for? I think if we choose to go down this path it should be far more selective, and there may be better alternatives than keeping a modal dialog open. I want the user to be as free to interact as soon as possible when opening files, or restoring state. Where possible the act of loading a data file should be as close to the act of restoring state containing that data file too.

mathturtle:
It is easiest to see if you have two datasets loaded with a volume in 2d mode on each.  Click back and forth on the two volume modules.  If you switch the active module/datasource while the histogram is loading for the volumes, then the 2d transfer function gets messed up.  Most of the time this results in a fully transparent volume whose transfer function can't be adjusted (fixed).

cryos:
I think we need to look at how we can mitigate the bug/bad state when restoring the state and the active data source changes, rather than locking the UI for all state files regardless of whether they contain 2D transfer functions or not.

At the very least this should only ever be invoked when 2D transfer functions are present in the state file, not for any and all state files.

mathturtle:
Closing in favor of #1679.

PR #1673, opened 2018-07-17
Add support for Focused Ion Beam DataSource type

### LINKED ISSUE #1625: Support FIB sectioning data
We need to support focused ion beam (FIB) sectioning data, which essentially adds a third type of DataSource. If marked as FIB then the image alignment tool would be available, but the reconstructions/tilt axis alignment would not. In all other respects we should probably treat this like a regular volume - it can be directly visualized, post-processed, etc. When setting to FIB we probably want to let the user select the dimensions in x, y, and the slice spacing but there are no angles.

alesgenova:
Fixes #1625


cryos:
Really high level comment from testing the application (no code review yet) - the preprocessing menu should also be enabled in my opinion.

PR #1674, opened 2018-07-17
Fix dialog not being shown for python operators

### LINKED ISSUE #1671: Python filters do not open dialog with parameters (first run)
Operators such as Gaussin Blur, Pad, etc run immediately on first execution, this was not the case in the 1.3.1 release. Operators expecting user input should open the dialog before execution.

alesgenova:
Fixes #1671

### DIFF
--- a/tomviz/operators/OperatorPython.cxx
+++ b/tomviz/operators/OperatorPython.cxx
@@ -257,6 +257,14 @@ void OperatorPython::setJSONDescription(const QString& str)
     }
   }
 
+  // Get the number of parameters
+  QJsonValueRef parametersNode = root["parameters"];
+  if (!parametersNode.isUndefined() && !parametersNode.isNull()) {
+    QJsonArray parametersArray = parametersNode.toArray();
+    QJsonObject::size_type numParameters = parametersArray.size();
+    setNumberOfParameters(numParameters);
+  }
+
   // Get child dataset information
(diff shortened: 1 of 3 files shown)


tomviz/operators/OperatorPython.cxx
```diff
@@ -257,6 +257,14 @@ void OperatorPython::setJSONDescription(const QString& str)
     }
   }
 
+  // Get the number of parameters
+  QJsonValueRef parametersNode = root["parameters"];
+  if (!parametersNode.isUndefined() && !parametersNode.isNull()) {
+    QJsonArray parametersArray = parametersNode.toArray();
+    QJsonObject::size_type numParameters = parametersArray.size();
+    setNumberOfParameters(numParameters);
```
mathturtle: What happens if the operator doesn't have JSON?  Delete Slices for example has arguments, but no JSON description.

mathturtle: Please test this with https://github.com/OpenChemistry/tomviz/pull/1663 (you may need that to make Delete Slices work right).

alesgenova: The EditOperatorDialog wouldn't be display on first run, but that was the case before as well, since the check was was whether or not a json description was provided.

The new numberOfParameters() method exists alongside the previous arguments() method, so I don't think it should break anything

PR #1679, opened 2018-07-23
More fixes relating to 2D transfer functions and state files

mathturtle:
This commit fixes 2 bugs relating to 2d transfer functions loading from
state files and the user interacting with the view as this was
happening.

1) If the user clicked around so that each dataset was not active when
its histogram finished generating, then the transfer function editor
would never get its box for user interaction and the transfer function
would become completely transparent.

2) If there were multiple datasources that were active when their
histograms finsihed, the box widget could be added twice (this caused
delays and lagginess when interacting).  This was exacerbated by the fix
to 1), so I blocked adding multiple box widgets to the editor.  We will
need to fix this when we add support for multiple box items in the 2D
transfer function.

@cryos I *think* this fixes the issues with interacting while the histogram is generating, making #1669 obsolete.

### DIFF
--- a/tomviz/CentralWidget.cxx
+++ b/tomviz/CentralWidget.cxx
@@ -278,6 +278,7 @@ void CentralWidget::setColorMapDataSource(DataSource* source)
   }
   if (histogram2D) {
     m_ui->histogram2DWidget->setHistogram(histogram2D);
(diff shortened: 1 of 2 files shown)


cryos:
I will take a look at this soon, hadn't forgotten about it.

PR #1698, opened 2018-08-10
Exporting Image Shifts on Auto Cross Correlation

jtschwar:
Addressing @yijiang1 's request to restrict the python operator from exporting image shifts that exceeds the image's coordinates

### DIFF
--- a/tomviz/python/AutoCrossCorrelationTiltImageAlignment.py
+++ b/tomviz/python/AutoCrossCorrelationTiltImageAlignment.py
@@ -66,6 +66,12 @@ class CrossCorrelationAlignmentOperator(tomviz.operators.CancelableOperator):
 
         utils.set_array(dataset, tiltSeries)
 
+        # Assign Negative Shifts when Shift > N/2.
+        indices_X = np.where(offsets[:, 0] > tiltSeries.shape[0] / 2)
+        offsets[indices_X, 0] -= tiltSeries.shape[0]
+        indices_Y = np.where(offsets[:, 1] > tiltSeries.shape[1] / 2)
+        offsets[indices_Y, 1] -= tiltSeries.shape[1]
+
         # Create a spreadsheet data set from table data
         column_names = ["X Offset", "Y Offset"]
         offsetsTable = utils.make_spreadsheet(column_names, offsets)


tomviz/python/AutoCrossCorrelationTiltImageAlignment.py
```diff
@@ -68,6 +68,10 @@ def transform_scalars(self, dataset):
 
         # Create a spreadsheet data set from table data
         column_names = ["X Offset", "Y Offset"]
+        offsets[np.where(offsets > tiltSeries.shape[0] / 2), :] = 
+            offsets[np.where(offsets > tiltSeries.shape[0] / 2), :] - tiltSeries.shape[0] / 2
+        offsets[:, np.where(offsets > tiltSeries.shape[1] / 2)] = 
+            offsets[:, np.where(offsets > tiltSeries.shape[1] / 2)] - tiltSeries.shape[1] / 2
```
mathturtle: I think you want less than `-shape[0]/2` and greater than `shape[0]/2` (replace with shape[1] for Y) since you should also consider large negative shifts.  I'm also not sure what you are doing with the subtraction... wouldn't you want to set it to `0` or `shape[0]/2` rather than subtract an arbitrary amount?

Isn't the format of offsets [[x, y], [x, y], [x, y]], where the x and y are the offsets for that slice?  In that case you would want to fix the second coordinate to 0 or 1 based on which you want to set, and then always do your `np.where` in the first dimension.  As it is this will affect both the X and Y offsets if the X offset is bad and I think it will edit all offsets in the column (X or Y) if the first or second one is bad. Try this out with some simple numpy arrays on the python console to see if it does what you expect (or hardcode some bad values and see what the result is).

tomviz/python/AutoCrossCorrelationTiltImageAlignment.py
```diff
@@ -68,6 +68,10 @@ def transform_scalars(self, dataset):
 
         # Create a spreadsheet data set from table data
         column_names = ["X Offset", "Y Offset"]
+        offsets[np.where(offsets > tiltSeries.shape[0] / 2), :] = 
+            offsets[np.where(offsets > tiltSeries.shape[0] / 2), :] - tiltSeries.shape[0] / 2
```
mathturtle: BTW: you can store the result of your `np.where(offsets > tiltSeries.shape[0] / 2)` in a variable if you don't want to repeat the whole thing.  This will help with readability.

jtschwar:
@mathturtle Hi Shawn, the image shifts are currently being exports as values between [0, N] where N is the number of pixels in either dimension. However, the issue is that the image isn't wrapping around to the opposite corner when the shifts are manually imported. Nevertheless, you were right that I was subtracting by the wrong amount. I've made changes to my script so now the image correctly shifts in the opposite direction whenever too large of a shift is calculated and it should now be easier to be read. Please let me know if any other changes need to be made, thanks.

tomviz/python/AutoCrossCorrelationTiltImageAlignment.py
```diff
@@ -66,6 +66,12 @@ def transform_scalars(self, dataset):
 
         utils.set_array(dataset, tiltSeries)
 
+        # Assign Negative Shifts when Shift > N/2.
+        max_indeces_X = np.where(offsets > tiltSeries.shape[0] / 2)
+        offsets[max_indeces_X] = offsets[max_indeces_X] - tiltSeries.shape[0]
```
mathturtle: This still affects both X and Y when X is too high.  I just tested it by setting the X offset for the first slice to well above the size.  This subtracted `tiltSeries.shape[0]` from both the X and Y offsets of the first slice.

Your condition to find X offsets that are too high also found the Y offset I found that was "too high" (actually it was a valid Y offset, I just forced it to be higher than the X limit.

Offsets (according to my print statement) look like this (note this print is from before I started changing values):
```
[ [ 98.  43.]
 [ 98.  36.]
 [ 98.  30.]
 [ 98.  25.]
 [ 98.  21.]
 [ 98.  20.]
...
]
```

So your search needs to take into account that there are two columns and only search the X or Y column, not both.

jtschwar:
@mathturtle Thanks (: I've just made the changes to address your recent comments. I was going to wait until all the checks were completed before requesting another review

tomviz/python/AutoCrossCorrelationTiltImageAlignment.py
```diff
@@ -66,6 +66,12 @@ def transform_scalars(self, dataset):
 
         utils.set_array(dataset, tiltSeries)
 
+        # Assign Negative Shifts when Shift > N/2.
+        indeces_X = np.where(offsets[:, 0] > tiltSeries.shape[0] / 2)
```
cryos: I think you mean indexes or indices as the prefix for the variable names, updating would make this more readable.

PR #1699, opened 2018-08-16
Add clamping to rasterization to account for bad input boxes

### LINKED ISSUE #1697: Crashes when deleting from pipeline
Stability of tomviz reduced by deleting objects from pipeline.

(1) Load default datasets
(2) Delete everything from pipeline
(3) Run "generate random particles" change values to 256

Crashes.

Everything is OK if repeated without loading default datasets and deleting.

mathturtle:
After deleting a data source some of its data (i.e. chart bounds) are
retained by the 2D transfer function chart.  Add clamping to the bounds
used in rasterization to avoid these leftovers causing segfaults when
data is loaded later (the invalid bounds leads to an invalid default box
and invalid indices being computed and set... accessing unallocated
memory).  Fixes #1697.


psavery:
This fixed the crash for me.

PR #1701, opened 2018-08-21
Emd read/write tilt angles

mathturtle:
This branch add support for round-tripping tilt angle data to/from the .emd format.

### DIFF
--- a/tomviz/DataSource.cxx
+++ b/tomviz/DataSource.cxx
@@ -56,6 +56,31 @@
 #include <cstring>
 #include <sstream>
 
+namespace {
+void createOrResizeTiltAnglesArray(vtkDataObject* data)
+{
+  auto fd = data->GetFieldData();
+  if (!fd->HasArray("tilt_angles")) {
+    int* extent = vtkImageData::SafeDownCast(data)->GetExtent();
+    int numTiltAngles = extent[5] - extent[4] + 1;
+    vtkNew<vtkDoubleArray> array;
+    array->SetName("tilt_angles");
+    array->SetNumberOfTuples(numTiltAngles);
+    array->FillComponent(0, 0.0);
+    fd->AddArray(array);
+  } else {
+    // if it exists, ensure the size of the tilt angles array
+    // corresponds to the size of the data
+    int* extent = vtkImageData::SafeDownCast(data)->GetExtent();
+    int numTiltAngles = extent[5] - extent[4] + 1;
+    auto array = fd->GetArray("tilt_angles");
+    if (numTiltAngles != array->GetNumberOfTuples()) {
+      array->SetNumberOfTuples(numTiltAngles);
+    }
+  }
+}
+} // namespace
+
 namespace tomviz {
 
 class DataSource::DSInternals
@@ -84,25 +109,7 @@ public:
       vtkAlgorithm::SafeDownCast(this->ProducerProxy->GetClientSideObject());
     Q_ASSERT(alg);
     auto data = alg->GetOutputDataObject(0);
-    auto fd = data->GetFieldData();
-    if (!fd->HasArray("tilt_angles")) {
-      int* extent = vtkImageData::SafeDownCast(data)->GetExtent();
-      int numTiltAngles = extent[5] - extent[4] + 1;
-      vtkNew<vtkDoubleArray> array;
-      array->SetName("tilt_angles");
-      array->SetNumberOfTuples(numTiltAngles);
-      array->FillComponent(0, 0.0);
-      fd->AddArray(array);
-    } else {
-      // if it exists, ensure the size of the tilt angles array
-      // corresponds to the size of the data
-      int* extent = vtkImageData::SafeDownCast(data)->GetExtent();
-      int numTiltAngles = extent[5] - extent[4] + 1;
-      auto array = fd->GetArray("tilt_angles");
-      if (numTiltAngles != array->GetNumberOfTuples()) {
-        array->SetNumberOfTuples(numTiltAngles);
-      }
-    }
+    createOrResizeTiltAnglesArray(data);
   }
 };
 
@@ -896,32 +903,14 @@ bool DataSource::hasTiltAngles()
 
 QVector<double> DataSource::getTiltAngles() const
 {
-  QVector<double> result;
   auto data = this->dataObject();
-  auto fd = data->GetFieldData();
-
-  if (fd->HasArray("tilt_angles")) {
-    auto tiltAngles = fd->GetArray("tilt_angles");
-    result.resize(tiltAngles->GetNumberOfTuples());
-    for (int i = 0; i < result.size(); ++i) {
-      result[i] = tiltAngles->GetTuple1(i);
-    }
-  }
-  return result;
+  return getTiltAngles(data);
 }
 
 void DataSource::setTiltAngles(const QVector<double>& angles)
 {
   auto data = this->dataObject();
-  auto fd = data->GetFieldData();
-  this->Internals->ensureTiltAnglesArrayExists();
-  if (fd->HasArray("tilt_angles")) {
-    auto tiltAngles = fd->GetArray("tilt_angles");
-    for (int i = 0; i < tiltAngles->GetNumberOfTuples() && i < angles.size();
-         ++i) {
-      tiltAngles->SetTuple1(i, angles[i]);
-    }
-  }
+  setTiltAngles(data, angles);
   emit dataChanged();
 }
 
@@ -1097,6 +1086,35 @@ bool DataSource::hasTiltAngles(vtkDataObject* image)
   vtkFieldData* fd = image->GetFieldData();
 
   return fd->HasArray("tilt_angles");
+}
+
+QVector<double> DataSource::getTiltAngles(vtkDataObject* data)
+{
+  QVector<double> result;
+  auto fd = data->GetFieldData();
+
+  if (fd->HasArray("tilt_angles")) {
+    auto tiltAngles = fd->GetArray("tilt_angles");
+    result.resize(tiltAngles->GetNumberOfTuples());
+    for (int i = 0; i < result.size(); ++i) {
+      result[i] = tiltAngles->GetTuple1(i);
+    }
+  }
+  return result;
+}
+
+void DataSource::setTiltAngles(vtkDataObject* data,
+                               const QVector<double>& angles)
+{
+  auto fd = data->GetFieldData();
+  createOrResizeTiltAnglesArray(data);
+  if (fd->HasArray("tilt_angles")) {
+    auto tiltAngles = fd->GetArray("tilt_angles");
+    for (int i = 0; i < tiltAngles->GetNumberOfTuples() && i < angles.size();
+         ++i) {
+      tiltAngles->SetTuple1(i, angles[i]);
+    }
+  }
 }
 
 } // namespace tomviz
--- a/tomviz/DataSource.h
+++ b/tomviz/DataSource.h
@@ -232,6 +232,9 @@ public:
   void setInitialContourValue(double initialContourValue);
 
   static bool hasTiltAngles(vtkDataObject* image);
+  static QVector<double> getTiltAngles(vtkDataObject* image);
+  static void setTiltAngles(vtkDataObject* image,
+                            const QVector<double>& angles);
 
 signals:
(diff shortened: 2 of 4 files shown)


tomviz/EmdFormat.cxx
```diff
@@ -113,6 +113,28 @@ class EmdFormat::Private
                      H5T_STD_U32LE);
   }
 
+  bool attribute(const std::string& group, const std::string& name,
+                 std::string& value)
```
mathturtle: I'm fairly sure this function needs some more error handling, but my Googling examples of reading string attributes from HDF5 files didn't show what types of errors you need to handle, just how to read the attributes.

cjh1: Does the API documentation not enumerate them?

tomviz/DataSource.cxx
```diff
@@ -56,6 +56,31 @@
...
+    int* extent = vtkImageData::SafeDownCast(data)->GetExtent();
+    int numTiltAngles = extent[5] - extent[4] + 1;
+    auto array = fd->GetArray("tilt_angles");
+    if (numTiltAngles != array->GetNumberOfTuples()) {
+      array->SetNumberOfTuples(numTiltAngles);
+    }
+  }
+}
+} // namespace
+
```
cjh1: Whats the reason for moving this outside the internals object?

tomviz/EmdFormat.cxx
```diff
@@ -607,38 +642,53 @@ bool EmdFormat::write(const std::string& fileName, vtkImageData* image)
...
-  d->setAttribute("/data/tomography/dim1", "units", "[n_m]", true);
+  std::vector<int> side(1);
+  side[0] = imageDimDataZ.size();
+  d->writeData("/data/tomography", "dim1", side, imageDimDataZ);
+  if (!hasTiltAngles) {
+    d->setAttribute("/data/tomography/dim1", "name", "z", true);
+    d->setAttribute("/data/tomography/dim1", "units", "[n_m]", true);
+  } else {
+    d->setAttribute("/data/tomography/dim1", "name", "angles", true);
+    d->setAttribute("/data/tomography/dim1", "units", "[deg]", true);
```
cjh1: We assume the angles are always in degrees?

tomviz/EmdFormat.cxx
```diff
@@ -607,38 +642,53 @@ bool EmdFormat::write(const std::string& fileName, vtkImageData* image)
...
-    d->setAttribute("/data/tomography/dim3", "name", "z", true);
-    d->setAttribute("/data/tomography/dim3", "units", "[n_m]", true);
-  } else {
-    d->setAttribute("/data/tomography/dim3", "name", "angles", true);
-    d->setAttribute("/data/tomography/dim3", "units", "[deg]", true);
-  }
+  side[0] = imageDimDataX.size();
+  d->writeData("/data/tomography", "dim3", side, imageDimDataX);
+  d->setAttribute("/data/tomography/dim3", "name", "x", true);
+  d->setAttribute("/data/tomography/dim3", "units", "[n_m]", true);
```
cjh1: Always nm?

tomviz/DataSource.cxx
mathturtle: I need to call this from a static member function where I don't have an internals object (`setTiltAngles`)

tomviz/EmdFormat.cxx
mathturtle: I think so... I've never heard of anyone using radians and I think all our reconstruction routines assume degrees.

cjh1: One of the EMD examples on their website uses radians, so thats why I was wondering. @cryos Thoughts?

tomviz/EmdFormat.cxx
mathturtle: No, we support the user assigning units in the properties panel.  I guess I should read/write those too, although currently this is no worse than before.

cjh1: This can be done in another PR. I can raise a issue.

tomviz/EmdFormat.cxx
cryos: I am a fan of keeping it simple for now, I would always save in degrees, but it would be good to do the conversion from radians to degrees when we read stuff in. I am fine with keeping this small, and iterating though, so feel free to limit to degrees as that is all I have seen.

tomviz/EmdFormat.cxx
cryos: We need to get better at units, but for now always nm is fine.

tomviz/DataSource.h
```diff
@@ -232,6 +232,9 @@ class DataSource : public QObject
   void setInitialContourValue(double initialContourValue);
 
   static bool hasTiltAngles(vtkDataObject* image);
+  static QVector<double> getTiltAngles(vtkDataObject* image);
```
cryos: What is the reasoning for making tilt angle stuff static? These are the only static methods as far as I can tell, and the data object knows what data it contains. This can be moved into a separate issue, but I think we should migrate these methods to operate on the vtkImageData in the DataObject, and remove the static methods.

mathturtle: In the EMD reader I only have the image data since the data source is not created yet, but I want to set/get the tilt angles and didn't want to duplicate code.  So I followed the same pattern as the existing static `hasTiltAngles` to make these.  These could be moved out of data source to somewhere else if that makes more sense.  (The non-static versions of these functions had their implementation changed to use the static versions similar to `hasTiltAngles`).

tomviz/EmdFormat.cxx
mathturtle: I've pushed an extra commit that handles reading files with radians... I think you are right about always saving in degrees though.

tomviz/EmdFormat.cxx
mathturtle: I've added a bit more error handling based on the API, but I was wondering about type issues.  Apparently the method of reading strings I used is only for variable length strings.... if the units string in the file is fixed-length it will fail.  But I can't figure out from the API how to test that, since the ways I have tried failed to read the files I know are right.

ercius:
This feature really useful! Thanks for adding it.

I found this page very useful for strings in HDF5 even though it's for h5py
and Numpy.

http://docs.h5py.org/en/latest/strings.html

I try to use numpy.string_ and the S datatype as discussed in the beginning
when writing dim vectors.

tomviz/EmdFormat.cxx
```diff
@@ -552,6 +582,23 @@ bool EmdFormat::read(const std::string& fileName, vtkImageData* image)
...
+      QVector<double> angles;
+      for (unsigned i = 0; i < dim1.size(); ++i) {
+        angles.push_back(dim1[i]);
+      }
+      DataSource::setTiltAngles(image, angles);
+    } else if (units == "[rad]") {
+      QVector<double> angles;
+      for (unsigned i = 0; i < dim1.size(); ++i) {
+        // Convert radians to degrees since tomviz assumes degrees everywhere.
+        angles.push_back(dim1[i] * 180.0 / M_PI);
```
cryos: The use of M_PI is causing a build failure on Windows.

tomviz/EmdFormat.cxx
```diff
@@ -113,6 +113,33 @@ class EmdFormat::Private
...
+      cout << group << name << " not found!" << endl;
+      return false;
+    }
+
+    hid_t attr = H5Aopen_by_name(fileId, group.c_str(), name.c_str(),
+                                 H5P_DEFAULT, H5P_DEFAULT);
+    hid_t type = H5Aget_type(attr);
+    char* tmpString;
+    if (H5Aread(attr, type, &tmpString) < 0) {
+      cout << "Failed to read attribute " << group << " " << name <<endl;
```
cryos: The endl needs a space in front, causing a clang-format warning.

mathturtle: I've squashed in this style fix.

tomviz/EmdFormat.cxx
mathturtle: I've switched to `vtkMath::Pi()`

cryos:
In my testing looking at a tilt series, I was able to save and load the tilt series however it kept losing the X dimension. For some reason it would go to 512 when I set it to 20. The Y dimension and the angles were preserved. When saving and loading a volume the volume was preserved, i.e. 20x30x40 came back in the right order, and was rendered in place.

We have to treat stacks differently, and I suspect something there needs tweaking.

mathturtle:
> For some reason it would go to 512 when I set it to 20.

I can reproduce this here... weird.  I'll dig into it.

mathturtle:
A missing flipped X/Z dimension was the cause of this... Fixed as a part of the new commit I just pushed after discussion with @cjh1 yesterday that removes the need to flip dimensions.

We were reading in the data and flipping the axes to avoid a row/column major fix to the whole volume.  That doesn't work when reading in tilt angles since all our algorithms assume they are on the Z-axis.  They need to be on the Z-axis the same as they are in the file, so I had to change the reader to do the row major -> column major flip on the volume instead.

I've tried to test everything: spacing, dimensions, tilt angles, etc.  The one thing that I know still fails is reading in some files where the strings are stored differently (see discussion above).

tomviz/EmdFormat.cxx
mathturtle: I found a file where the strings are stored with a fixed width and the current approach segfaults.  Apparently when the strings are stored fixed width, you have to allocate memory to put them in and pass in a char*.  When the strings are stored non-fixed width, you pass in a char** and it allocates the memory for you.  But I haven't figured out the API to tell how the string is stored.

mathturtle:
If finally found an example that showed reading in a string with the proper error handling so I've fixed that.  But using vtkImagePermute doesn't seem to do the row-major -> column-major flip right  on that example file.  I need to investigate further (do not merge).

cryos:
Reordering isn't a cheap operation, and so if the dimension flipping works for volumes (which it did in my testing) we should keep that, and special case where angles are present. Image stacks will tend to be smaller, and so the cost is not as high, I also don't see why we can't switch the angle to the correct axis in the case of image stacks for C ordering, but that may turn out to be more complex and not worth the time.

Let's put this on hold for now, I would like to try and get some examples of image stacks in EMD if such a thing exists. They are not volumes, and so will always require some special handling. I don't think I have any examples of stacks with angles, and so would rather not spend too much time on reordering if no one else will use the output.

mathturtle:
> I would like to try and get some examples of image stacks in EMD if such a thing exists

Example files would be great!  I've been guessing based of the format specs and their example programs so far.  Based on those it seems like the tilt angles should be on dim3 which is flipped to the X-axis without reordering.  Tomviz expects them on the Z-axis and it will be far easier to reorder the image than change all our reconstructions and alignments to accept x-axis tilt angles.

ercius:
> I know it isn't a cheap operation.

Its not cheap on the input and output, but it makes processing internally (cross-correlation, sum, media, etc.) much faster. So, its probably worth doing.

> My concern is this: If we exchange the X and Z axes we are changing from a right-handed to a left-handed coordinate system. This means anything with handedness (helices & some other molecules IIRC from high school chemistry) will be displayed as having the opposite handedness. I don't know how much we care about this though.

There are sufficient issues with determining handedness objectively that a researcher trying to do such a thing would have to be careful throughout their pipeline. See for example [this paper](https://www.sciencedirect.com/science/article/pii/S1047847713001081) on the subject. I would not worry about flipping things in this manner.

>Example files would be great! I've been guessing based of the format specs and their example programs so far. Based on those it seems like the tilt angles should be on dim3 which is flipped to the X-axis without reordering. Tomviz expects them on the Z-axis and it will be far easier to reorder the image than change all our reconstructions and alignments to accept x-axis tilt angles.

I sent an example file to @cryos at his request by email, which hopefully you received as well. The format and program was not written with C-ordering in mind originally. So, I think that dim1 should be considered as the tilt angles to keep with C-ordering expected of HDF5 and to keep operations on the images fast. This is how I am writing all EMD files now. We will likely be adopting whatever LibreTEM does and/or the NEXUS format in the future. Reordering the image when reading in is my preferred method to minimize the computations later.

cryos:
It seems like this is a step in the right direction, let's merge as Chris confirmed that round tripping data for tilt series and volumes continues to work as expected within Tomviz at least.

PR #1709, opened 2018-09-18
Emit dataSourceChanged every time a source is activated

### LINKED ISSUE #1691: Pipeline glitchy when deleting items from pipeline
The new pipeline has become sensitive glitches when items are deleted. In particular, the pipeline often gets confused as to whether or not it has been marked as a tiltseries:

This image shows a tiltseries that thinks it is a volume:

screen shot 2018-08-05 at 9 32 11 pm

alesgenova:
Fixes #1691 

When the last operator is removed, the transformed dataSource is destroyed, and the parent dataSource is activated.

It appears that the `dataSourceChanged` signal is not fired, and all the actions observing that signal to set their enabled/disabled state get out of sync.

This "fix" fires the dataSourceChanged event every time dataSourceActivated is fired, solving the problem observed in #1691

### DIFF
--- a/tomviz/ActiveObjects.cxx
+++ b/tomviz/ActiveObjects.cxx
@@ -106,12 +106,12 @@ void ActiveObjects::setActiveDataSource(DataSource* source)
       m_activeDataSourceType = source->type();
     }
     m_activeDataSource = source;
-    emit dataSourceChanged(m_activeDataSource);
 
     // Setting to nullptr so the traverse logic is re-run.
     m_activeParentDataSource = nullptr;
   }
   emit dataSourceActivated(m_activeDataSource);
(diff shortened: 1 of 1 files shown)


alesgenova:
From my memory, the slots that would need to be updated to listen to both signals are:

- Python operator actions (can be fixed once for all of them in the superclass)
- C++ operator actions that need to be activated/deactivated based on the datasource type. (they need to be patched one by one, by connecting to both signals)

I started going down that way when I was working on this PR, but then decided that it was an even dirtier solution than the one proposed here

cryos:
I just created an issue to track whether we want separate signals, but think we should merge this too, and look at the impact on fixing issues outlines in reports.

Issue #553, opened 2016-08-08
Import XYZ file reader from Paraview


Hovden:
Can we look into adding the XYZ file reader from Paraview? I want to see what kind of features we can easily add for the atom probe tomography users.

PR #1714, opened 2018-09-24
Add molecule as dataSource

alesgenova:
Will allow tomviz to open `xyz` files and display atomic structures along other classic dataSources.

Tasks:
- [x] Add a MoleculeSource class, and instantiate it in loadDataReaction
- [x] Extend the pipeline to display MoleculeSource
- [x] Extend ModuleMolecule to be attached to a MoleculeSource
- [x] Reuse the same panel widget between OperatorResult and MoleculeSource
- [x] Enable removing MoleculeSource from the pipeline
- [x] Add MoleculeSource to the recently opened files
- [x] Save MoleculeSource to the state file

screenshot from 2018-10-01 08-22-28

MEMO:
Cherrypick a fix in `VTK` for the `vtkXYZMolReader2`
```
git cherry-pick 7ac470dfcdf6b259d4481ac4fa12b46d0e7bb2e1
```

Fixes: #553

### DIFF
--- a/tomviz/LoadDataReaction.cxx
+++ b/tomviz/LoadDataReaction.cxx
@@ -150,7 +150,10 @@ QList<DataSource*> LoadDataReaction::loadData()
   QList<DataSource*> dataSources;
   if (dialog.exec()) {
     QStringList filenames = dialog.selectedFiles();
-    if (filenames.size() > 1) {
+    QString fileName = filenames.size() > 0 ? filenames[0] : "";
+    QFileInfo info(fileName);
+    QStringList tiffExt = { "tif", "tiff" };
+    if (filenames.size() > 1 && tiffExt.contains(info.suffix().toLower())) {
       dataSources << LoadStackReaction::loadData(filenames);
     } else {
       dataSources << loadData(filenames);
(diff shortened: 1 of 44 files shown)


cjh1:
We have a few VTK leaks. You should be able to see these by setting VTK_DEBUG_LEAKS on the ParaView build.
```
vtkDebugLeaks has detected LEAKS!
Class "9vtkBufferItE" has 4 instances still around.
Class "9vtkBufferIxE" has 2 instances still around.
Class "vtkDataSetAttributes" has 4 instances still around.
Class "vtkUnsignedShortArray" has 4 instances still around.
Class "vtkGraphInternals" has 2 instances still around.
Class "vtkPoints" has 2 instances still around.
Class "vtkInformation" has 10 instances still around.
Class "vtkMolecule" has 2 instances still around.
Class "vtkInformationIntegerValue" has 8 instances still around.
Class "vtkFloatArray" has 2 instances still around.
Class "vtkIdTypeArray" has 2 instances still around.
Class "vtkFieldData" has 2 instances still around.
Class "9vtkBufferIfE" has 2 instances still around.
```

tomviz/modules/ModuleFactory.cxx
```diff
@@ -50,31 +50,38 @@ QList<QString> ModuleFactory::moduleTypes()
...
-        << "Threshold";
+        << "Threshold"
+        << "Molecule";
   qSort(reply);
   return reply;
 }
 
 bool ModuleFactory::moduleApplicable(const QString& moduleName,
                                      DataSource* dataSource,
+                                     MoleculeSource* moleculeSource,
```
cjh1: Why do we pass in both `dataSource` and `moleculeSource`? Should this method just be overloaded, not sure why we would need both?

tomviz/LoadDataReaction.cxx
```diff
@@ -191,7 +198,10 @@ DataSource* LoadDataReaction::loadData(const QStringList& fileNames,
     fileName = fileNames[0];
   }
   QFileInfo info(fileName);
-  if (info.suffix().toLower() == "emd") {
+  if (info.suffix().toLower() == "xyz") {
+    LoadDataReaction::loadMolecule(fileNames);
+    return nullptr;
```
cjh1: Why return `nullptr` and not the new `MoleculeSource`?

tomviz/PipelineExecutor.cxx
```diff
@@ -451,7 +450,7 @@ void DockerPipelineExecutor::cancel(std::function<void()> canceled)
           });
 }
 
-bool DockerPipelineExecutor::cancel(Operator* op)
+bool DockerPipelineExecutor::cancel(Operator*)
```
cjh1: I would prefer we didn't touch this, as it will cause conflicts with the existing docker PR.

tomviz/modules/ModuleFactory.cxx
cryos: Looks like it would be clearer as an overload to me too.

tomviz/modules/ModuleFactory.cxx
```diff
@@ -50,31 +50,38 @@ QList<QString> ModuleFactory::moduleTypes()
...
                                      vtkSMViewProxy* view)
 {
   if (dataSource && view) {
+    if (moduleName == "Molecule") {
```
cryos: This logic would then be resolved by the compiler.

PR #1717, opened 2018-10-03
TV filter

jtschwar:
Total Variation based filter to remove stripes and structured artifacts.

### DIFF
--- a/tomviz/DataTransformMenu.cxx
+++ b/tomviz/DataTransformMenu.cxx
@@ -74,6 +74,7 @@ void DataTransformMenu::buildTransforms()
   auto laplaceFilterAction = menu->addAction("Laplace Sharpen");
   auto gaussianFilterAction = menu->addAction("Gaussian Blur");
   auto wienerAction = menu->addAction("Wiener Filter");
+  auto TVminAction = menu->addAction("Remove Stripes, Curtaining, Scratches");
   auto peronaMalikeAnisotropicDiffusionAction =
     menu->addAction("Perona-Malik Anisotropic Diffusion");
   auto medianFilterAction = menu->addAction("Median Filter");
@@ -139,6 +140,9 @@ void DataTransformMenu::buildTransforms()
   new AddPythonTransformReaction(
     wienerAction, "Wiener Filter", readInPythonScript("WienerFilter"), false,
     false, false, readInJSONDescription("WienerFilter"));
+  new AddPythonTransformReaction(
+    TVminAction, "TV_Filter", readInPythonScript("TV_Filter"), false,
+    false, false, readInJSONDescription("TV_Filter"));
   new AddPythonTransformReaction(
     gaussianFilterAction, "Gaussian Blur", readInPythonScript("GaussianFilter"),
     false, false, false, readInJSONDescription("GaussianFilter"));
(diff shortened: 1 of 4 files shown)


tomviz/DataTransformMenu.cxx
```diff
@@ -139,6 +140,9 @@ void DataTransformMenu::buildTransforms()
   new AddPythonTransformReaction(
     wienerAction, "Wiener Filter", readInPythonScript("WienerFilter"), false,
     false, false, readInJSONDescription("WienerFilter"));
+  new AddPythonTransformReaction(
```
alesgenova: This lines don't conform with clang-format (see the CI fail).
When running clang-format on my machine it shows it should be
```cpp
new AddPythonTransformReaction(TVminAction, "TV_Filter",
                               readInPythonScript("TV_Filter"), false, false,
                               false, readInJSONDescription("TV_Filter"));

```

tomviz/python/TV_Filter.py
```diff
@@ -0,0 +1,121 @@
...
+                      (phi <= (theta+dtheta/2)))] = 0
+        mask[np.where((phi >= (np.pi+theta-dtheta/2)) &
+                      (phi <= (np.pi+theta+dtheta/2)))] = 0
+        mask[np.where((phi >= (-np.pi+theta-dtheta/2)) &
+                      (phi <= (-np.pi+theta+dtheta/2)))] = 0
+        mask[np.where(rr < np.square(kmin))] = 1 # Keep values below rmin.
+        mask = np.array(mask, dtype=bool)
+
+        # Initialize Progress bar.
+        if nz == 1:
```
alesgenova: Instead of having two different logic for the progress depending on how many images are in the stack, I think it could be generalized.
Something like `self.progress.maximum = nx * Niter`
And then in the inner loop over `Niter` do `self.progress.value = i * Niter + j `
And you can also get rid of manually setting the `step` variable

tomviz/python/TV_Filter.py
```diff
@@ -0,0 +1,116 @@
...
+                    self.progress.value = step
+
+                # Initializte the Next Loop.
+                recon_init = recon_minTV
+
+            # Return reconstruction into stack.
+            array[:, :, i] = recon_constraint
+
+            # Update the Progress Bar.
+            self.progress.value = i* Niter + j
```
alesgenova: This should be moved to line 78 and replace the `if nz ==1:` block

tomviz/python/TV_Filter.py
```diff
@@ -0,0 +1,111 @@
...
+        #Main Loop
+        for i in range(nz):
+
+            #FFT of the Original Image.
+            FFT_image = fftshift(fftn(array[:, :, i]))
+
+            # Reconstruction starts as random image.
+            recon_init = np.random.rand(nx, ny)
+
+            if self.canceled:
```
alesgenova: I would move this inside the `for k in range(20)` loop, so that if the operator is cancelled, the computation is stopped promptly, instead of up to 10 min later

tomviz/python/TV_Filter.py
```diff
@@ -0,0 +1,111 @@
...
+                    recon_minTV = recon_minTV - a*d*vst
+
+                # Initializte the Next Loop.
+                recon_init = recon_minTV
+
+            # Return reconstruction into stack.
+            array[:, :, i] = recon_constraint
+
+            # Update the Progress Bar.
+            self.progress.value = i*Niter + j
```
alesgenova: This needs to be inside the `for j in range(Niter)` loop to work properly

jtschwar:
@alesgenova Hi, thanks for your suggestions! I've just finished implementing all your advice. I'm unsure why my branch is failing the testson CircleCI.

alesgenova:
Looks great, approved!

Issue #1716, opened 2018-10-03
Profile application startup time for delays


cryos:
It was pointed out that recent Tomviz releases suffer from slow start up times, we should look at what is causing this, especially on Windows where it seems most pronounced. I suspect it is the Python file formats, and if so it would be good to delay loading these if they are blocking.

If we can't avoid this, or perhaps in addition, we should consider displaying a splash screen to at least confirm that Tomviz is in the process of loading.

PR #1718, opened 2018-10-05
Improve startup time by making python init non-blocking

alesgenova:
On a cold start, the main windows is shown after 12 seconds.
Before this PR it would take 18 seconds.

Fixes: #1716 (?)


cryos:
There is a build error on Windows

PR #1747, opened 2018-11-26
Minimize memory consumption in Direct Fourier Transform 

jtschwar:
Move the mapping of the data from Fourier to real space at the end of the operation to speed up computational time. Also deprecated the file size of the pyFFTW objects.

### DIFF
--- a/tomviz/python/Recon_DFT.py
+++ b/tomviz/python/Recon_DFT.py
@@ -1,6 +1,7 @@
 import pyfftw
 import numpy as np
 import tomviz.operators
+from tomviz import utils
 import time
 
 
@@ -10,9 +11,6 @@ class ReconDFMOperator(tomviz.operators.CancelableOperator):
         """3D Reconstruct from a tilt series using Direct Fourier Method"""
 
         self.progress.maximum = 1
-
-        from tomviz import utils
-        import numpy as np
 
         # Get Tilt angles
         tiltAngles = utils.get_tilt_angles(dataset)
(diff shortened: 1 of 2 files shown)


tomviz/python/Recon_DFT.py
```diff
@@ -33,17 +31,12 @@ def transform_scalars(self, dataset):
...
-            (Nx, Ny, Nz), 16, dtype='float64', order='F')
-        recon_fftw_object = pyfftw.FFTW(
-            v, recon, direction='FFTW_BACKWARD', axes=(0, 1, 2))
+        v = pyfftw.empty_aligned(
+            (Nx, Ny, Nz // 2 + 1), dtype='complex64', n=8)
 
-        p = pyfftw.n_byte_align_empty((Nx, Npad), 16, dtype='float64')
-        pF = pyfftw.n_byte_align_empty(
-            (Nx, Npad // 2 + 1), 16, dtype='complex128')
+        p = pyfftw.empty_aligned((Nx, Npad), dtype='float32', n=8)
```
cjh1: Why choose 8 as the alignment here?

jtschwar: _n_ is the number of bytes for this object and I want to reduce the memory consumption in half.

tomviz/python/Recon_DFT.py
```diff
@@ -64,14 +57,14 @@ def transform_scalars(self, dataset):
             projection = tiltSeries[:, :, a] #2D projection image
             p = np.lib.pad(projection, ((0, 0), (pad_pre, pad_post)),
                            'constant', constant_values=(0, 0)) #pad zeros
-            p = np.fft.ifftshift(p)
+            p = np.float32(np.fft.ifftshift(p))
             p_fftw_object.update_arrays(p, pF)
             p_fftw_object()
+            p = [] #Garbage collector (gc)
```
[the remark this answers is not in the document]
jtschwar: Ok, that makes more sense.

tomviz/python/Recon_DFT.py
cjh1: Doesn't ```n``` determine the byte alignment?

jtschwar: Ah, yes you're right. I misinterpreted it. Let's keep n at 16 then.

tomviz/python/Recon_DFT.py
```diff
@@ -33,17 +31,12 @@ def transform_scalars(self, dataset):
...
         w = np.zeros((Nx, Ny, Nz // 2 + 1)) #store weighting factors
-        v = pyfftw.n_byte_align_empty(
-            (Nx, Ny, Nz // 2 + 1), 16, dtype='complex128')
-        v = np.zeros(v.shape) + 1j * np.zeros(v.shape)
-        recon = pyfftw.n_byte_align_empty(
-            (Nx, Ny, Nz), 16, dtype='float64', order='F')
-        recon_fftw_object = pyfftw.FFTW(
-            v, recon, direction='FFTW_BACKWARD', axes=(0, 1, 2))
+        v = pyfftw.empty_aligned(
+            (Nx, Ny, Nz // 2 + 1), dtype='complex64', n=16)
```
cryos: 16 bit aligned is likely our best choice to take advantage of SSE in modern 64 bit processors, and should likely remain our default choice unless we get fancier in querying for the best alignment.

Issue #1745, opened 2018-11-20
Python Warning bin images x2


Hovden:
**Describe the bug**
Python warning thrown when using bin images x2

<img width="517" alt="screen shot 2018-11-20 at 2 54 06 pm" src="https://user-images.githubusercontent.com/6197997/48799368-f2786b00-ecd4-11e8-9131-00eefc39667f.png">

**To Reproduce**
Load dataset. Mark as tiltseries. Bin Images x2.

**Desktop (please complete the following information):**
 - Operating system: OSX
 - Tomviz version: 1.5.0

**Hardware (please complete the following information):**
 - Device: Macbook Pro

cryos:
This should be resolved now.

PR #1755, opened 2018-12-03
Suppress warning messages.

jtschwar:
Addressing Issue [#1745](https://github.com/OpenChemistry/tomviz/issues/1745)

### DIFF
--- a/tomviz/python/BinTiltSeriesByTwo.py
+++ b/tomviz/python/BinTiltSeriesByTwo.py
@@ -4,6 +4,7 @@ def transform_scalars(dataset):
     from tomviz import utils
     import scipy.ndimage
     import numpy as np
+    import warnings
 
     array = utils.get_array(dataset)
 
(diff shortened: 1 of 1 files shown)


tomviz/python/BinTiltSeriesByTwo.py
```diff
@@ -4,13 +4,15 @@ def transform_scalars(dataset):
...
     import numpy as np
+    import warnings
 
     array = utils.get_array(dataset)
 
     zoom = (0.5, 0.5, 1)
     result_shape = utils.zoom_shape(array, zoom)
     result = np.empty(result_shape, array.dtype, order='F')
     # Downsample the dataset x2 using order 1 spline (linear)
+    warnings.simplefilter("ignore")  #suppress warnings
```
cryos: As far as I know this ignores all warnings, which may not be good. What about something like (I haven't tested it):
`warnings.filterwarnings('ignore', '.*output shape of zoom.*')`
To more specifically ignore the warning we know is not relevant here.

jtschwar: That is a good point, it would be better not to filter all possible warnings. I've tested your recommendation and it works as well.

PR #1760, opened 2018-12-07
Search for datasource files in the same directory as the state file

### LINKED ISSUE #1748: Loading state file crashes without output
**Describe the bug**
I received a state file and MRC tilt series from a collaborator. I placed the state file and MRC tilt series in the same folder. Then loaded the state file into tomviz. The program eventually crashes without any output.

I tested this on two separate Windows 10 machines one with tomviz 1.4 and the other with a fresh 1.5.0 installation.

I dug into the state file and see these lines:
            "reader": {
                "fileNames": [
                    "../Tomo_CSH_01.mrc"
                ],
                "name": "MRCSeriesReader"

I removed the "../" in front of the file name and the state file loads properly.

I then tried to reproduce the issue and can confirm that saving a state file in a child directory will add this relative "../" path to the state file file loading. 

**To Reproduce**
1. Load a data set. (not the example data sets since their path is known)
2. Save a state file in a new directory created within the directory of the data set:
data set location: ../Desktop/data.mrc
state file location: ../Desktop/newFolder/state1.tvsm
3. Move the state file to the parent directory
data set location: ../Desktop/data.mrc
state file location: ../Desktop/state1.tvsm
4. Load the state file
5. tomviz crashes.

**Expected behavior**
This is especially a problem when a collaborator sends the state file and data file to another collaborator. The directory structure of the sender should not need to be replicated by the sendee. There should be a warning rather than a crash.

Ideally, tomviz should
1. look in the location written in the state file.
2. look in the same directory as the state file. 
3a. Warn the use that the data file can not be found. Tell the user the expected location. Reset tomviz
OR
3b. Give the user the option to find the data set or cancel.

Not sure about 3b is better but I do not know how hard that would be to implement.

 - Operating system: Windows 10
 - Tomviz version:  1.4.0 and 1.5

**Hardware (please complete the following information):**
 - Device: Dell XPS 13; Dell Precision tower
 - Graphics card: Intel card; NVIDIA card

alesgenova:
Fixes #1748

### DIFF
--- a/tomviz/modules/ModuleManager.cxx
+++ b/tomviz/modules/ModuleManager.cxx
@@ -948,7 +948,6 @@ void ModuleManager::onPVStateLoaded(vtkPVXMLElement*,
       QStringList fileNames;
       if (dsObject.contains("reader")) {
         auto reader = dsObject["reader"].toObject();
-        options["reader"] = reader;
 
         if (reader.contains("fileNames")) {
           foreach (const QJsonValue& value, reader["fileNames"].toArray()) {
@@ -956,13 +955,23 @@ void ModuleManager::onPVStateLoaded(vtkPVXMLElement*,
             if (QFileInfo::exists(value.toString())) {
               fileNames << value.toString();
             } else {
-              qCritical() << "File" << value.toString() 
-                          << "not found, skipping.";
+              // If the file cannot be found in the path relative to the state
+              // file, make another attempt to locate it in the same directory
+              QString altLocation =
+                d->dir.absoluteFilePath(QFileInfo(value.toString()).fileName());
+              if (QFileInfo::exists(altLocation)) {
+                fileNames << altLocation;
+              } else {
+                qCritical() << "File" << value.toString()
+                            << "not found, skipping.";
+              }
             }
           }
+          reader["fileNames"] = QJsonArray::fromStringList(fileNames);
         } else {
           qCritical() << "Unable to locate file name(s).";
         }
+        options["reader"] = reader;
       }
 
       DataSource* dataSource = nullptr;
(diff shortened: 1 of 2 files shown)


tomviz/modules/ModuleManager.cxx
```diff
@@ -948,21 +948,31 @@ void ModuleManager::onPVStateLoaded(vtkPVXMLElement*,
...
             // Verify the file exists before adding it to the list.
             if (QFileInfo::exists(value.toString())) {
               fileNames << value.toString();
             } else {
-              qCritical() << "File" << value.toString() 
-                          << "not found, skipping.";
+              // If the file cannot be found in the path relative to the state
+              // file
+              // make another attempt to locate it in the same directory instead
+              QString altLocation = this->d->dir.absoluteFilePath(
```
cryos: Don't need this->, just d->

tomviz/modules/ModuleManager.cxx
```diff
@@ -948,21 +948,31 @@ void ModuleManager::onPVStateLoaded(vtkPVXMLElement*,
...
         if (reader.contains("fileNames")) {
           foreach (const QJsonValue& value, reader["fileNames"].toArray()) {
             // Verify the file exists before adding it to the list.
             if (QFileInfo::exists(value.toString())) {
               fileNames << value.toString();
             } else {
-              qCritical() << "File" << value.toString() 
-                          << "not found, skipping.";
+              // If the file cannot be found in the path relative to the state
+              // file
```
cryos: Try to fix the odd wrapping here to make it flow better (guessing clang-format did this).

tomviz/modules/ModuleManager.cxx
```diff
@@ -1010,21 +1020,34 @@ void ModuleManager::onPVStateLoaded(vtkPVXMLElement*,
...
       if (dsObject.contains("reader")) {
         auto reader = dsObject["reader"].toObject();
-        options["reader"] = reader;
 
         if (reader.contains("fileName")) {
           fileName = reader["fileName"].toString();
+          // Verify the file exists.
+          if (!QFileInfo::exists(fileName)) {
+            // If the file cannot be found in the path relative to the state
+            // file
```
cryos: Here too, reflow the comment.

tomviz/modules/ModuleManager.cxx
```diff
@@ -1010,21 +1020,34 @@ void ModuleManager::onPVStateLoaded(vtkPVXMLElement*,
...
+            // make another attempt to locate it in the same directory instead
+            fileName =
+              this->d->dir.absoluteFilePath(QFileInfo(fileName).fileName());
```
cryos: Don't need this->, just d->

PR #1764, opened 2018-12-24
Let modules select active scalars independently from the dataSource

alesgenova:


tomviz/DataSource.cxx
```diff
@@ -594,6 +594,17 @@ void DataSource::setActiveScalars(const QString& arrayName)
   emit dataPropertiesChanged();
 }
 
+void DataSource::setActiveScalars(int arrayIdx)
+{
+  QStringList scalars = listScalars();
+
+  if (arrayIdx < 0 || arrayIdx >= scalars.length()) {
```
cryos: I think this should be done first, and then go get the scalars if they are going to be used.

tomviz/DataSource.h
```diff
@@ -177,6 +177,14 @@ class DataSource : public QObject
   /// Set the active scalars by array name.
   void setActiveScalars(const QString& arrayName);
   QString activeScalars() const;
+  /// Set the active scalars by component index.
+  void setActiveScalars(int arrayIdx);
+  int activeScalarsIdx() const;
```
cryos: I think we should consider being a little more explicit with the API, and call it Index rather than Idx. It doesn't save that many characters.

tomviz/modules/Module.cxx
```diff
@@ -285,6 +294,10 @@ bool Module::deserialize(const QJsonObject& json)
     setUseDetachedColorMap(useDetachedColorMap);
   }
 
+  if (json.contains("activeScalars")) {
```
cryos: I am a little torn on whether activeScalars is a little vague, consider activeScalarArray. Would be a pain to rename, I think we should think about the name before it gets too ingrained, but we can change it in a follow up commit.

tomviz/modules/ModuleSlice.cxx
```diff
@@ -219,20 +225,30 @@ void ModuleSlice::addToPanel(QWidget* panel)
   }
 
   QVBoxLayout* layout = new QVBoxLayout;
+  QFormLayout* formLayout = new QFormLayout;
```
cryos: Consider using auto to avoid repetition of the class names. We should take a pass to make some of this more consistent.

tomviz/modules/ModuleVolume.cxx
```diff
@@ -287,8 +306,7 @@ void ModuleVolume::onTransferModeChanged(const int mode)
 
 vtkSmartPointer<vtkDataObject> ModuleVolume::getDataToExport()
 {
-  vtkTrivialProducer* trv = this->dataSource()->producer();
-  return trv->GetOutputDataObject(0);
+  return m_imageData.GetPointer();
```
cryos: We don't need GetPointer()/Get() anymore - can use the implicit cast to pointer in the vtkNew/vtkSmartPointer objects.

PR #1765, opened 2018-12-28
Combine Slice and OrthogonalSlice into a single Module

alesgenova:


alesgenova:
@cryos 
As it turns out, the `vtkNonOrthoImagePlaneWidget` was adding large amounts of padding to the plane, and it would rely on clipping planes to hide plane exceeding the boundaries of the dataset.

screenshot from 2019-01-10 15-13-36

However, when the direction or the plane is orthogonal, there is no need for padding at all.
In https://github.com/OpenChemistry/tomviz/pull/1765/commits/97669ff84855276733e409e28e63017bb4545ec8 the padding for orthogonal planes  is removed, and when a custom plane is selected only the minimal needed amount is added.
This fixes the default zoom when the slice is present in the scene.

PR #1775, opened 2019-01-16
Deprecate data types in all the reconstruction algorithms 

jtschwar:
Now reconstruction variable types are 'float32'.

Signed-off-by: Jonathan Schwartz <jtschw@umich.edu>


cryos:
Do you mean deprecate here? It looks like you mean demote, or simply "Prefer 32 bit floating point output from reconstructions". This looks good to me aside from being uncertain about the log message.

jtschwar:
@cryos You're right, demote is a much better definition.

PR #1777, opened 2019-01-17
Load a cached dummy version of the python readers on startup

### LINKED ISSUE #1767: File formats take too long to load - hardcode/cache
The Python file format delayed loading takes too long in several scenarios, we should cache custom formats, and I think ideally hard code the formats we ship with to avoid a very slow initial start up. We could then background update these and store a cache in the settings file.

alesgenova:
Fixes #1767

### DIFF
--- a/tomviz/FileFormatManager.cxx
+++ b/tomviz/FileFormatManager.cxx
@@ -76,8 +76,8 @@ QMap<QString, T*> registerFactories(const QString& name,
     settings->beginWriteArray(name);
     for (int i = 0; i < cached.size(); ++i) {
       settings->setArrayIndex(i);
-      settings->setValue("description", cached.at(i).first);
-      settings->setValue("extensions", cached.at(i).second.join(","));
+      settings->setValue("description", cached[i].first);
+      settings->setValue("extensions", cached[i].second.join(","));
     }
     settings->endArray();
   }
--- a/tomviz/FileFormatManager.cxx
+++ b/tomviz/FileFormatManager.cxx
@@ -7,6 +7,9 @@
 #include "PythonWriter.h"
 #include <QDebug>
 
+#include <pqApplicationCore.h>
+#include <pqSettings.h>
+
 namespace tomviz {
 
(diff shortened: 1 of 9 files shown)


tomviz/FileFormatManager.cxx
```diff
@@ -16,87 +19,150 @@ FileFormatManager& FileFormatManager::instance()
...
       foreach (auto extension, extensions) {
         factories[extension] = factory;
       }
     }
+
+    auto settings = pqApplicationCore::instance()->settings();
+    settings->beginWriteArray(name);
+    for (int i = 0; i < cached.size(); ++i) {
+      settings->setArrayIndex(i);
+      settings->setValue("description", cached.at(i).first);
```
cryos: Prefer [] over at as you already did bounds checking in the loop initialization.

tomviz/FileFormatManager.h
```diff
@@ -4,6 +4,8 @@
 #ifndef tomvizFileFormatManager_h
 #define tomvizFileFormatManager_h
 
+#include <mutex>
```
cryos: We may want to consider the Qt mutex classes which have read locks (multiple readers) and write locks (only one write lock, no read locks). I suspect we aren't that multithreaded, but worth considering. In pure Qt code there are often advantages in just using the Qt classes for better integration, this may not matter too much here.

tomviz/FileFormatManager.cxx
```diff
@@ -16,87 +19,150 @@ FileFormatManager& FileFormatManager::instance()
...
+      factories[extension] = factory;
+    }
   }
+  settings->endArray();
+  return factories;
+}
+
+void FileFormatManager::prepopulatePythonReaders()
+{
+  auto factories = populateFactories<PythonReaderFactory>("pythonReaders");
```
cryos: So this just reads the settings, what is the behavior on first load of Tomviz, that it doesn't have any Python readers, but also doesn't have any recent files so we are all good. That seems reasonable but want to check I am following what is happening here.

Issue #1778, opened 2019-01-17
Canceling a Reconstruction Loses Tilt Angles State


jtschwar:
**Describe the bug**

If a Reconstruction is canceled before running, Tomviz seems to lose its Tilt Angles state. 

**To Reproduce**
1. Load any tilt series.
2. Click on 'Mark Data as Tilt Series'
3. Scroll down to 'Any Python Reconstruction and Cancel'

![image](https://user-images.githubusercontent.com/32110921/51334285-85235500-1a4d-11e9-98ef-9a09abcf9bba.png)

**Expected behavior**

For Tomviz not require to re-mark as Tilt series. 

**Screenshots**

**Desktop (please complete the following information):**
 - Operating system: macOS 10.11
 - Tomviz version: 1.5.1

**Hardware (please complete the following information):**
 - Device: Macbook Pro

**Additional context**

psavery:
@cryos I don't think this has been fixed yet. I can look into this.

PR #1790, opened 2019-01-18
Fix removal of operators with childDataSource

alesgenova:
It is my understanding that the new pipeline doesn't really support branching anymore.
These leftover bits of code were causing wrong behavior when removing operators that have a `hasChildDataSource` flag.

Fixes: #1778


cryos:
Let's think about this a little, there is branching from reconstructions, there are two duplicate code paths that are effectively doing the same thing, and I need to load some of that back into my head as I would like to get rid of one of the two approaches and only do it one way.

PR #1807, opened 2019-03-07
Added Array Wrangler

psavery:
The component selection is still WIP.

### DIFF
--- a/tomviz/DataTransformMenu.cxx
+++ b/tomviz/DataTransformMenu.cxx
@@ -40,7 +40,7 @@ void DataTransformMenu::buildTransforms()
   auto cropDataAction = menu->addAction("Crop");
   auto convertDataAction = menu->addAction("Convert to Float");
   auto arrayWranglerAction =
-    menu->addAction("Wrangle Data");
+    menu->addAction("Convert to Type");
   auto reinterpretSignedToUnignedAction =
     menu->addAction("Reinterpret Signed to Unsigned");
   menu->addSeparator();
--- a/tomviz/DataTransformMenu.cxx
+++ b/tomviz/DataTransformMenu.cxx
@@ -39,8 +39,7 @@ void DataTransformMenu::buildTransforms()
 
   auto cropDataAction = menu->addAction("Crop");
   auto convertDataAction = menu->addAction("Convert to Float");
-  auto arrayWranglerAction =
-    menu->addAction("Convert to Type");
+  auto arrayWranglerAction = menu->addAction("Convert Type");
   auto reinterpretSignedToUnignedAction =
     menu->addAction("Reinterpret Signed to Unsigned");
   menu->addSeparator();
--- a/tomviz/DataTransformMenu.cxx
+++ b/tomviz/DataTransformMenu.cxx
@@ -9,6 +9,7 @@
 
 #include "AddExpressionReaction.h"
 #include "AddPythonTransformReaction.h"
+#include "ArrayWranglerReaction.h"
 #include "CloneDataReaction.h"
 #include "ConvertToFloatReaction.h"
 #include "CropReaction.h"
@@ -38,6 +39,8 @@ void DataTransformMenu::buildTransforms()
 
   auto cropDataAction = menu->addAction("Crop");
   auto convertDataAction = menu->addAction("Convert to Float");
+  auto arrayWranglerAction =
+    menu->addAction("Wrangle Data");
   auto reinterpretSignedToUnignedAction =
     menu->addAction("Reinterpret Signed to Unsigned");
   menu->addSeparator();
@@ -77,6 +80,7 @@ void DataTransformMenu::buildTransforms()
   new AddExpressionReaction(customPythonAction);
   new CropReaction(cropDataAction, mainWindow);
   new ConvertToFloatReaction(convertDataAction);
+  new ArrayWranglerReaction(arrayWranglerAction, mainWindow);
   new AddPythonTransformReaction(
     reinterpretSignedToUnignedAction, "Reinterpret Signed to Unsigned",
     readInPythonScript("ReinterpretSignedToUnsigned"));
--- a/tomviz/operators/ArrayWranglerOperator.cxx
+++ b/tomviz/operators/ArrayWranglerOperator.cxx
@@ -6,13 +6,14 @@
 #include "EditOperatorWidget.h"
 
 #include <vtkFloatArray.h>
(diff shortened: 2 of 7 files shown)


psavery:
@cryos As far as I can tell, the conversion to UInt8 and UInt16 work correctly. Let me know if you see any issues.

The loop is [here](https://github.com/OpenChemistry/tomviz/pull/1807/files#diff-388464ebacd9067619e2b01a0323c4bcR113). I added 0.5 before casting to int to ensure flooring rounds correctly.

I tried to use templating to make it generic in case we want to add other type conversions in the future. But it still needs some cleanup I think to make it more compact and readable.

psavery:
By the way, in the UI, it is called "Wrangle Data" right under "Convert to Float".

tomviz/DataTransformMenu.cxx
```diff
@@ -38,6 +39,8 @@ void DataTransformMenu::buildTransforms()
 
   auto cropDataAction = menu->addAction("Crop");
   auto convertDataAction = menu->addAction("Convert to Float");
+  auto arrayWranglerAction =
+    menu->addAction("Wrangle Data");
```
cryos: I think we should call it "Convert to Type" in the menu, the class names are fine, but it may not be obvious to end users. Not sure on that, but that is my second attempt at menu name.

tomviz/operators/ArrayWranglerOperator.cxx
```diff
@@ -0,0 +1,180 @@
...
+{
+  Q_OBJECT
+
+public:
+  ArrayWranglerWidget(tomviz::ArrayWranglerOperator* source,
+             vtkSmartPointer<vtkImageData> imageData, QWidget* p)
+    : tomviz::EditOperatorWidget(p), m_operator(source),
+      m_outputTypes(nullptr), m_componentToKeep(nullptr)
+  {
+    QLabel* label = new QLabel("Convert to:", this);
```
cryos: Prefer auto where type is obvious.

tomviz/operators/ArrayWranglerOperator.cxx
```diff
@@ -0,0 +1,180 @@
...
+    label->setAlignment(Qt::AlignRight | Qt::AlignVCenter);
+
+    m_outputTypes = new QComboBox(this);
+
+    using OutputType = tomviz::ArrayWranglerOperator::OutputType;
+    // This will ensure the indexing matches that of the enum...
+    m_outputTypes->insertItem(static_cast<int>(OutputType::UInt8),  "UInt8");
+    m_outputTypes->insertItem(static_cast<int>(OutputType::UInt16), "UInt16");
+
+    QHBoxLayout* hboxlayout = new QHBoxLayout(this);
```
cryos: Use auto

tomviz/operators/ArrayWranglerOperator.cxx
```diff
@@ -0,0 +1,180 @@
...
+
+  // Can be simplified to:
+  // new = ((old - oldmin) * multiplier;
+  // where multiplier is newmax / oldrange
+
+  double multiplier = newmax / (oldrange[1] - oldrange[0]);
+
+  auto d = static_cast<vtkInputDataType*>(data);
+  auto a = static_cast<outputType*>(array->GetVoidPointer(0));
+  for (int i = 0; i < nComps * nTuples; ++i) {
```
cryos: Use vtkIdType for the counter, we want to support arrays as big as VTK supports (64 bit index)

cryos:
It works well in my testing, maybe wrangle data is a good name too, or Data Type Conversion. It is an overloaded term that is the issue. Change Type?

psavery:
@cryos I added the component selection feature. It only appears in the dialog box if there is more than one component in the source. I'm using a QComboBox, maybe I should use a QSpinBox instead.

I think it works. In that sample I have, it does condense the data to one channel. But it looks like the channels may be equivalent in the sample.

I still have code cleanup to do...

cryos:
Most images I encounter are RGB gray scale, and so all three channels are the same value, and picking one is best. This will help common issues with data loading at least, other use cases may come up in the future. It is a pipeline, so you can string several conversions together too.

cryos:
I think we should use two rows and two columns, Convert to: and Select Component: with a default of 1 looks good.

tomviz/operators/ArrayWranglerOperator.cxx
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+private:
+  QPointer<tomviz::ArrayWranglerOperator> m_operator;
+  QComboBox* m_outputTypes;
+  QComboBox* m_componentToKeep;
+};
+} // namespace
+
+#include "ArrayWranglerOperator.moc"
```
cryos: Why are you including the moc file? Shouldn't normally be needed in Tomviz

cryos: Nevermind - private QObject derived class, I see now.

psavery:
> It works well in my testing, maybe wrangle data is a good name too, or Data Type Conversion. It is an overloaded term that is the issue. Change Type?

I might change the name to "Change Type". Does that sound good?

psavery:
Are the serialize and deserialize functions (like [these](https://github.com/OpenChemistry/tomviz/blob/master/tomviz/operators/CropOperator.cxx#L103)) for state files? I haven't added one yet to the data wrangler.

psavery:
Nevermind, I saw that they are for the state files. I added the serialize and deserialize functions.

cryos:
I don't think the commit prefixes add very much, and would rather see a succinct description of the change. The topic branch name gives that high level overview of what the branch is achieving, and repeating it with a commit prefix will just be repetitive when reviewing the shortlog.

tomviz/operators/ArrayWranglerOperator.cxx
```diff
@@ -135,11 +135,14 @@ void wrangleVtkArrayTypeUnsigned(vtkOutputArrayType* array, int nComps,
 }
 
 template <typename vtkOutputArrayType>
-void applyGenericWrangleTransform(vtkImageData* imageData, int componentToKeep,
-                                  double range[2])
+void applyGenericWrangleTransform(vtkImageData* imageData, int componentToKeep)
```
cryos: This is already in an anonymous namespace, don't repeat wrangle too much, it is already in the file name, and applyGenericTransform is already quite obvious.

psavery:
@cryos I think this is ready for review. Let me know if I have forgotten anything...

PR #1810, opened 2019-03-07
Add *.swp to gitignore for vim .swp files

psavery:


cryos:
I put things like this in my local global gitignore, I don't think we should be pushing every editor/OS temp file, same with CMakeLists.txt.user, .directory, etc.

psavery:
Okay. I'll put it in my global gitignore also.

Issue #1809, opened 2019-03-07
NPY files written as Fortran order


ercius:
**Describe the bug**
Using the save as and selecting the numpy file (.npy) option saves the data as a Fortran contiguous data set. This is related to #1808. 

The data is ordered (X,Y,num) for tilt series when it should be [num,Y,X]. Also the array.flags variable shows:
  C_CONTIGUOUS : False
  F_CONTIGUOUS : True
  OWNDATA : False
  WRITEABLE : True
  ALIGNED : True
  UPDATEIFCOPY : False

**To Reproduce**
1. Load the example tilt series
2. Save the data as a npy file
3. Load the data into python using np.load
4. array.shape and array.flags show that the data is Fortran Contiguous. 

```python
>> import numpy as np
>> ser = np.load('test.npy')
>> print(ser.shape)
(256, 256, 74)
>> print(ser.flags)
  C_CONTIGUOUS : False
  F_CONTIGUOUS : True
  OWNDATA : False
  WRITEABLE : True
  ALIGNED : True
  UPDATEIFCOPY : False
```

**Expected behavior**
The data should be transposed and properly ordered in C-ordering before being saved as a numpy array to keep with the proper conventions in numpy and python.
```python
>> out_array = np.ascontiguousarray(np.transpose(tomviz_array, [2,0,1]))
>> np.save('filename.npy', out_array)
```
Then C_CONTIGUOUS flag is True and shape = (74,256,256)
**Desktop (please complete the following information):**
 - Windows 10
 - Tomviz version: 1.5.1

**Hardware (please complete the following information):**
 - Device: Dell XPS 13

ercius:
Additional comments:
1) Im not sure if the np.transpose() should be [2,0,1] or [2,1,0] to keep the tilt axis properly oriented when importing into external numpy. That should be tested.
2) Reading npy files should also take this into account if tomviz is allowed to load arbitrary npy files. It seems that tomviz should have a spec for what it expects in terms of ndarray shape in a npy file. Im not sure if this is discussed anywhere.

cryos:
This is fixed and will be in the next release.

Issue #1808, opened 2019-03-07
EMD datasets load incorrectly into HDF5 or h5py


ercius:
**Describe the bug**
The EMD files written by tomviz can not be opened by HDF5view or h5py. They do load properly back into tomviz though. My guess is that the images are being written in Fortran ordering but h5py/HDF5 expects C-ordering. Im not sure if I can get HDF5 or h5py to understand the Fortran ordering to test this.

**To Reproduce**
1. Load the example reconstruction
2. Save as EMD file.
3. Load into HDFview and view as image or load into h5py and view as an image. Here is an example image from h5py:
![figure_1](https://user-images.githubusercontent.com/15657108/53990240-af53d380-40dc-11e9-8534-4d521f24f688.png)

**Expected behavior**
The data should be loaded and look normal as expected.

**Desktop (please complete the following information):**
 - Windows 10, Dell XPS 13
 - Tomviz version: 1.5.1

cryos:
We have code in our [Python Docker logic](https://github.com/cjh1/tomviz/blob/future/tomviz/python/tomviz/executor.py#L368) that is attempting to do this right, but it looks like maybe we are still missing something from the looks of your screenshot. Is there a Python sample to reproduce how you read/displayed this? It looks like maybe matplotlib is displaying a slice of the data?

psavery:
It looks like there is an attempt to write to row-major order [here](https://github.com/OpenChemistry/tomviz/blob/master/tomviz/EmdFormat.cxx#L283), but I guess something is missing...

Let us know, @ercius, if you have a Python sample to reproduce your image above.

psavery:
FYI, I think I figured out how to view the EMD file (see an example below), and I think the problem here is related to the discussion in #1812 in that the data needs to be transposed rather than just the axes re-ordered. I'll try to find a straightforward way to transpose vtkImageData.

```
import h5py
from matplotlib import pyplot as plt

f = h5py.File('file.emd', 'r')

emdgrp = f['/data/tomography']
data = emdgrp['data'][:]

slice = data[130] # Pick a slice to view

plt.imshow(slice)
plt.show()
``

ercius:
Yes. This is a good set of command to read in the EMD file using h5py. You can also use ncempy (pip install ncempy) to read the EMD file without having to know what is in it:
```python
>> import ncempy.io as nio
>> import matplotlib.pyplot as plt
>> emd0 = nio.emd.fileEMD('filename.emd')
[...]

PR #1812, opened 2019-03-08
Use C ordering for npy format

psavery:
This appears to save and and load files correctly, but when
loading, it prints this message:

Warning, array does not have Fortran order,
making deep copy and fixing...
...done.

I'm looking into this...

### DIFF
--- a/tomviz/python/tomviz/io/formats/numpy.py
+++ b/tomviz/python/tomviz/io/formats/numpy.py
@@ -24,6 +24,10 @@ class NumpyWriter(Writer, NumpyBase):
 
     def write(self, path, data_object):
         data = tomviz.utils.get_array(data_object)
+
+        # Switch to row major order for NPY stores
+        data = data.reshape(data.shape[::-1])
+
         with open(path, "wb") as f:
             np.save(f, data)
 
@@ -36,6 +40,9 @@ class NumpyReader(Reader, NumpyBase):
 
         if len(data.shape) != 3:
             return vtkImageData()
+
+        # NPY stores data as row major order. VTK expects column major order.
+        data = data.reshape(data.shape[::-1])
 
         image_data = vtkImageData()
         (x, y, z) = data.shape
--- a/tomviz/python/tomviz/io/formats/numpy.py
+++ b/tomviz/python/tomviz/io/formats/numpy.py
@@ -26,7 +26,7 @@ class NumpyWriter(Writer, NumpyBase):
         data = tomviz.utils.get_array(data_object)
 
         # Switch to row major order for NPY stores
-        data = data.reshape(data.shape[::-1])
+        data = np.ascontiguousarray(np.transpose(data, [2, 1, 0]))
 
         with open(path, "wb") as f:
             np.save(f, data)
@@ -42,7 +42,7 @@ class NumpyReader(Reader, NumpyBase):
             return vtkImageData()
 
         # NPY stores data as row major order. VTK expects column major order.
-        data = np.asfortranarray(data.reshape(data.shape[::-1]))
+        data = np.asfortranarray(np.transpose(data, [2, 1, 0]))
 
         image_data = vtkImageData()
         (x, y, z) = data.shape
--- a/tomviz/python/tomviz/io/formats/numpy.py
+++ b/tomviz/python/tomviz/io/formats/numpy.py
@@ -42,7 +42,7 @@ class NumpyReader(Reader, NumpyBase):
             return vtkImageData()
 
         # NPY stores data as row major order. VTK expects column major order.
-        data = data.reshape(data.shape[::-1])
+        data = np.asfortranarray(data.reshape(data.shape[::-1]))
 
         image_data = vtkImageData()
         (x, y, z) = data.shape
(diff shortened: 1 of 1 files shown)


psavery:
This fixes the issues mentioned in #1809. Pasted below, you can see the output from the steps after I made these changes. Tomviz also appears to load the file correctly (without the warning mentioned above).

I am not sure why the `asfortranarray` is needed [here](https://github.com/OpenChemistry/tomviz/pull/1812/files#diff-c207ceb83a554d92fc1c84a6a99bf34dR45) (it wasn't present in [here](https://github.com/cjh1/tomviz/blob/future/tomviz/python/tomviz/executor.py#L382)), but it seems to be needed.

```
$ python3
Python 3.6.7 (default, Oct 22 2018, 11:32:17) 
[GCC 8.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import numpy as np
>>> ser = np.load('tilt_c2.npy')
>>> print(ser.shape)
(74, 256, 256)
>>> print(ser.flags)
  C_CONTIGUOUS : True
  F_CONTIGUOUS : False
  OWNDATA : True
  WRITEABLE : True
  ALIGNED : True
  WRITEBACKIFCOPY : False
  UPDATEIFCOPY : False
```

cryos:
In the processing of data numpy defaults to C order, and so loading/processing in C order in a pure Python pipeline makes sense. When we want to render something we need to reorder for vtkImageData, so bringing this into C++ for Tomviz requires the reorder to Fortran on the way in, and to C on the way out. So I think what you see makes sense.

psavery:
Actually, I think Peter was right in that we need to use np.transpose() instead. About to submit an update...

psavery:
Just to ensure everything was working properly, I tried loading these '.npy' files in numpy and visualizing them with matplotlib. You can do so with code like follows:

```
import numpy as np
import matplotlib.pyplot as plt

tilt_series = np.load('tilt_series.npy')
slice = tilt_series[37] # Look at slice with index 37
plt.imshow(slice)
plt.show()
```

I was testing against the sample tilt series. A sample image from tomviz of a slice with index 37 is below.

tilt

Using numpy and matplotlib, the method with reshaping would produce a bad image from the .npy file:
reshape

But the transpose method produces a better image:
transpose

Where the matplotlib image has its y-axis inverted (with 0 at the top).

With the reshape method, I could save and load the file with tomviz fine. But the numpy/matplotlib images were bad. The transpose method is better. I can save and load the file with tomviz, and the numpy/matplotlib images are good.

ercius:
> With the reshape method, I could save and load the file with tomviz fine. But the numpy/matplotlib images were bad. The transpose method is better. 

The `np.reshape()` method will give you a **view** into the original array. It will not physically reorder the values. Im pretty sure `np.transpose()` will also do the same thing. You may also have to apply `np.ascontiguousarray()` to be sure its C-ordered. Always check the `array.flags` parameters to check the ordering.

psavery:
@ercius Actually, for the examples here, I got the same results with and without `np.ascontiguousarray()`. As in, with or without `np.ascontiguousarray()`, the `reshape()` method still produces the bad image above, and with or without `np.ascontiguousarray()`, the `transpose()` method still produces the good image above. So `np.ascontiguousarray()` may not be necessary here, but I added it anyways in case it is required for other examples...

I think this shows that `reshape()` and `transpose()` are doing something different. But the `transpose()` method does what we need.

I have also tested saving this and loading back into tomviz, and it seems to work fine.

psavery:
@cryos So the way this is set up right now is that the data is ordered (z,y,x) for all cases (both volumes and image stacks). Just to clarify, do we want it to be (z,y,x) for all cases, or just for image stacks (and (x,y,z) for volumes)?

cryos:
@psavery I would say it is only important for image stacks, for volumes I think there is some value in having the axes preserved if we were to compare across different visualization programs in the future.

psavery:
@ercius We discussed and decided to leave the .npy axes the same for now (no X and Z swapping), both for volumes and tilt series. However, we are planning to swap the X and Z axes for EMD tilt series (see #1816) so that the images can be scrolled through faster. So you might prefer to use EMD for tilt series when we get that merged.

psavery:
On my latest push, I tested this writer/reader several different ways:

1. I made sure I could save an npy file and load it back into tomviz, and that it appeared the same.
2. I round tripped an npy file, and using the `diff` command, I found it to be identical to the original npy file.
3. I opened the npy file with numpy (code in the discussion above). It has (X, Y, Z) ordering and `C_CONTIGUOUS : True`.
4. I viewed the file with matplotlib, and I can visualize it fine.

ercius:
Leaving the npy axes as-is works for me. This is mostly a convenience thing as long as the images can be loaded and look correct in python. Even leaving it as Fortran ordering would be ok since numpy can take care of this. 

aside: EMD datasets in #1808 have to be written as C-ordered although they do not have to be written with axes as [z,y,x] as long as the images can be viewed by the user.

cryos:
I think writing C order here is the right thing to do, and it will work out of the box for most users this way. These are merged, and it would be great to get confirmation that things work for you now.

PR #1814, opened 2019-03-11
Reorder the EMD arrays for the volume

cryos:
This is working on fixing issue #1808, where the code in this patch
ensures that a volume will be stored in the expected C order. I have
verified that this works as expected when loaded by matplotlib using the
simple Python example. Round tripping also worked with non-uniform
dimensions.

The major issue at this point is that all old EMDs will fail to load as
they were incorrectly stored in the Fortran order in the HDF5 file. I
also think that we need to treat image stacks differently in that the
slices should remain in contiguous memory so that they can be accessed.
This will require a special case for image stacks.

### DIFF
--- a/tomviz/EmdFormat.cxx
+++ b/tomviz/EmdFormat.cxx
@@ -7,7 +7,6 @@
 
 #include <vtkDataArray.h>
 #include <vtkImageData.h>
-#include <vtkImagePermute.h>
 #include <vtkNew.h>
 #include <vtkPointData.h>
 #include <vtkTrivialProducer.h>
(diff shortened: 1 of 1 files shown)


ercius:
Maybe you can add a tomviz version attribute in the file. This can than
indicate <1.5.1 are Fortran ordered and the rest are c-ordered?

cryos:
I was just talking with @cjh1 about this, and because the old files have no indication of the writer version we can't reliably know if it was Tomviz or Python or another program. We should add a writer version now, but for previously written files I think it would be a mistake to reorder them as that would break correctly ordered EMD files written elsewhere.

ercius:
Ah yes. That is true. I didn't think about that.

PR #1815, opened 2019-03-11
Added operator to transpose C and Fortran ordering

psavery:
This was added in light of the changes to the EMD
reader/writer (#1814), so that if a user loads an old tomviz
EMD file, they can transpose and use their data.

### DIFF
--- a/tomviz/CMakeLists.txt
+++ b/tomviz/CMakeLists.txt
@@ -163,6 +163,8 @@ set(SOURCES
   TomographyReconstruction.cxx
   TomographyTiltSeries.h
   TomographyTiltSeries.cxx
+  TransposeDataReaction.h
+  TransposeDataReaction.cxx
   Utilities.cxx
   Utilities.h
   Variant.cxx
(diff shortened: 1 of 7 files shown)


psavery:
This has a "Transpose Data" operator where the user then selects C or Fortran ordering via a combo box. With the new EMD reader/writer, the user can load the old EMD files and then transpose to C ordering to fix it (the old EMD files were saved in Fortran ordering, and then they will be converted to Fortran ordering a second time on reading the file, so they need to be converted back one step to view the data correctly).

Two questions:

1. Are we only interested in the "Transpose to C" operator? That is the one that transforms the old EMD format into the new one. If we are only interested in the "Transpose to C" operator, I can remove the dialog box where the user picks "C" or "Fortran" ordering, and I can just name the whole operator "Transpose to C Ordering" or something like that.

2. Since [this commit](https://github.com/OpenChemistry/tomviz/commit/9893b8122d7e5e70b45d3073b4e5f6056014ff2a#diff-aa34731822e5a2f4621038caf01d6b0f), the EMD files have been written with (Z, Y, X) axes orderings instead of (X, Y, Z). So I think EMD files written by Tomviz 1.5.0 and Tomviz 1.5.1 will have their axes swapped with respect to the previous and future versions. The Fortran/C reordering does not swap the axes back, so if we load an EMD file made by Tomviz 1.5.0/1.5.1, it will have its X and Z axes swapped. Is this fine? Maybe the user can modify the axis ordering with some operators?

Issue #1800, opened 2019-02-22
Change the text color for scale cube


jtschwar:
**Describe the bug**

I would like to be able to change the color of the text for the scale cube. It would be useful when we change the background color to white.

cryos:
This was fixed in 1.6

PR #1818, opened 2019-03-13
Added ability to modify scale cube text color

psavery:


psavery:
Fixes #1800. As far as I can tell, it works properly. The text color gets set when the user modifies it, and it is saved and loaded back in correctly via the state files.

Issue #1799, opened 2019-02-18
Tomviz can't load recent matlab (.mat) v7.3 files. 


jtschwar:
**Describe the bug**
Tomviz is unable to load the most recent version (v7.3) of matlab (.mat) files. 

**To Reproduce**
1. Load this [dataset](https://drive.google.com/file/d/1uoBpstMxp9vOmP3Lb0if462HrJvGVhpz/view?usp=sharing) into tomviz. 

'Traceback (most recent call last):
  File "/Users/jonathanschwartz/tomviz-build/lib/site-packages/tomviz/io/_internal.py", line 37, in execute_reader
    return obj.read(path)
  File "/Users/jonathanschwartz/tomviz-build/lib/site-packages/tomviz/io/formats/matlab.py", line 27, in read
    mat_dict = scipy.io.loadmat(path)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/scipy/io/matlab/mio.py", line 141, in loadmat
    MR, file_opened = mat_reader_factory(file_name, appendmat, **kwargs)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/scipy/io/matlab/mio.py", line 71, in mat_reader_factory
    raise NotImplementedError('Please use HDF reader for matlab v7.3 files')
NotImplementedError: Please use HDF reader for matlab v7.3 files
Error calling the reader'

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**
 - Operating system: macOS 10.14.3
 - Tomviz version: 1.5.1-41-g516e7464

**Hardware (please complete the following information):**
 - Device: Macbook Pro

**Additional context**

jtschwar:
Actually, after looking into the [API](https://docs.scipy.org/doc/scipy-0.19.0/reference/generated/scipy.io.loadmat.html) it looks like scipy.io does not support this file format.

psavery:
It looks like an HDF5 python library such as h5py is needed to read matlab v7.3 files instead of scipy.

Do we want to try to support matlab v7.3 files? If so, we could perhaps do a try/except for the scipy reader, and if it shows up as matlab v7.3, we can try to import h5py and read the file.

@cryos @cjh1

cryos:
For now we might add a try/except and so new files are not supported. We don't have h5py packaged yet.

cryos:
[...]

PR #1820, opened 2019-03-13
Print error when loading matlab v7.3 files

psavery:
For issue #1799.

@cryos, something like this?

It would be nice if there was a pop-up error, but I'm not sure if we can do that with the python readers.

It prints out these messages in the message dock in tomviz:
```
Tomviz does not currently support matlab v7.3 files
The file didn't contain any suitable volumetric data
```

The first line is in green and the second line is in red (and the second line is there just because we are returning an empty vtkImageData()).

Note that this fix is specific to matlab v7.3. If future versions require hdf5 as well, we'll have to add them.

Here is the error message from scipy:
`NotImplementedError: Please use HDF reader for matlab v7.3 files`

### DIFF
--- a/tomviz/python/tomviz/io/formats/matlab.py
+++ b/tomviz/python/tomviz/io/formats/matlab.py
@@ -24,7 +24,14 @@ class MatlabBase(IOBase):
 class MatlabReader(Reader, MatlabBase):
 
     def read(self, path):
-        mat_dict = scipy.io.loadmat(path)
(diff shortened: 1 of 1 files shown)


jtschwar:
@cjh1 @psavery you might want to consider recommending the user to convert the data to 7.2 matfile so that they know which file types are supported.

psavery:
@jtschwar done.

PR #1821, opened 2019-03-14
Update the basic instructions for module namespace

cryos:
The tomviz module is now a namespace, the instructions for running need
to be updated. Also, the default view on GitHub can make people miss the
document on PASSIVE.md.

### DIFF
--- a/acquisition/PASSIVE.md
+++ b/acquisition/PASSIVE.md
@@ -35,7 +35,7 @@ Once everything is installed you can start the acquisition server with the
 following commands.
 
     workon tomviz
-    python -m tomviz.acquisition.cli -a tomviz.acquisition.vendors.passive.PassiveWatchSource
+    tomviz-acquisition -a tomviz.acquisition.vendors.passive.PassiveWatchSource
 
 This will result in a process running in the terminal on the machine that has a
 directory to be watched passively.
@@ -61,7 +61,7 @@ interface to aid in development. The following (from within the 'tomviz' Python
 virtual environment) will cause an image to be written from a stack every five
 seconds:
 
-    python -m tests.mock.tiltseries.writer -p /tmp/test -d 5 -t tiff
+    tomviz-tiltseries-writer -p /tmp/test -d 5 -t tiff
 
 It supports type of 'dm3' too, the path, delay, and type can all be modified. At
 present the angle information is not being fed through, this will be added next.
--- a/acquisition/PASSIVE.md
+++ b/acquisition/PASSIVE.md
@@ -23,6 +23,7 @@ prvileges).
     mkvirtualenv tomviz
     pip install 'git+https://cjh1@bitbucket.org/cjh1/pydm3reader.git@filelike'
     pip install https://github.com/bottlepy/bottle/archive/41ed6965.zip
+    pip install Pillow requests mock diskcache
 
 At this point you will have a Python environment with the required Python tools.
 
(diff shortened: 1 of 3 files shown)


cryos:
Try these updated instructions @jtschwar

acquisition/PASSIVE.md
```diff
@@ -34,7 +33,7 @@ Once everything is installed you can start the acquisition server with the
 following commands.
 
     workon tomviz
-    tomviz-acquisition -a tomviz.acquisition.vendors.passive.PassiveWatchSource
+    python -m tomviz.acquisition.cli -a tomviz.acquisition.vendors.passive.PassiveWatchSource
```
cjh1: Do `tomviz-acquisition` not work for you?

acquisition/PASSIVE.md
```diff
@@ -60,7 +59,7 @@ interface to aid in development. The following (from within the 'tomviz' Python
 virtual environment) will cause an image to be written from a stack every five
 seconds:
 
-    tomviz-tiltseries-writer -p /tmp/test -d 5 -t tiff
+    python -m tests.mock.tiltseries.writer -p /tmp/test -d 5 -t tiff
```
cjh1: Likewise, this command works for me?

acquisition/README.md
```diff
@@ -10,7 +13,7 @@ pip install -r requirements-dev.txt
 
 ```bash
 cd <tomviz_repo>/acquisition
-python -m tomviz
+python -m tomviz.acquisition.cli
```
cjh1: This could be `tomviz-acquisition`

acquisition/PASSIVE.md
```diff
@@ -21,10 +21,11 @@ prvileges).
...
     cd tomviz/acquisition
     mkvirtualenv tomviz
-    pip install --process-dependency-links -e .
-    pip install --process-dependency-links -e .[dm3]
-    pip install --process-dependency-links -e .[tiff]
-    pip install --process-dependency-links -e .[test]
+    pip install 'git+https://cjh1@bitbucket.org/cjh1/pydm3reader.git@filelike'
+    pip install https://github.com/bottlepy/bottle/archive/41ed6965.zip
+    pip install -e .
+    pip install -e .[tiff]
```
psavery: I think this should be `tif` instead of `tiff` by the way. See [here](https://github.com/OpenChemistry/tomviz/blob/master/acquisition/setup.py#L29).

psavery: With `tiff`, I get the message `tomviz-acquisition 0.0.1 does not provide the extra 'tiff'`

psavery:
I think a few other small instructions need to be updated as well in the PASSIVE.md file, specifically in the `Connecting from the Application` section. It mentions to click on `Introspect`, but I don't think there is an `Introspect`. It also mentions that a typical filename regex is `.*`, but the file name format automatically generates the regex for us.

acquisition/PASSIVE.md
cryos: Fixed setup.py

acquisition/README.md
cryos: I think it is fine for now, and works, we need more work on this but at least it can be started form the guide now.

acquisition/PASSIVE.md
```diff
@@ -35,7 +35,7 @@ Once everything is installed you can start the acquisition server with the
 following commands.
 
     workon tomviz
-    python -m tomviz.acquisition.cli -a tomviz.acquisition.vendors.passive.PassiveWatchSource
+    tomviz.acquisition -a tomviz.acquisition.vendors.passive.PassiveWatchSource
```
psavery: Should be `tomviz-acquisition`, right?

cryos: Yes, good catch, thought I had copy-pasted. Thanks.

PR #1828, opened 2019-03-18
Allow a user to input negative angle increments

### LINKED ISSUE #1827: Cant set tilt angles with negative tilt increment
**Describe the bug**
I can not set the tilt angles for a tilt series using a negative tilt step increment.

**To Reproduce**
1. Load the example tilt series
2. Double click on the Set Tilt Angles transform
3. Set start tilt angle to positive 73 and last tilt angle to -73. Tilt step should be negative
4. Press ok or apply.
5. CLick on the child dataset named Output in the pipeline. Note that the tilt angles in the properties box are still -73 to 73.

Alternate example:
1.  Load a tilt series (or load the example tilt series and delete the Set tilt angles transform)
2. Click on Tomograhy -- Mark data as tilt series
3. Set the start and end angles such that you have a positive tilt step
4. Click ok or apply
5. Check the properties of the child dataset. All tilt angles are 0:
image

Errors will pop up in later transforms such as the Automated alignment (XCOR):
  File "Auto Tilt Image Align (XCORR)", line 20, in transform_scalars
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
Failed to execute the script.

**Expected behavior**
Allow user to set tilt angles with either a positive or negative tilt step.

**Screenshots**

 - WIndows 10
 - 1.5.1

**Hardware (please complete the following information):**
 - Device: [e.g. Macbook Pro]
 - Graphics card: [e.g. NVIDIA Quadro 6000]

**Additional context**

psavery:
Fixes #1827, where if a user started with a positive
angle and ended with a negative angle, the data would
be all zeros.

In the `if()` statement in question, delta can be negative
under two conditions:

1. The start angle is **greater** than the end angle, and the start
image is **less** than the end image.
2. The start angle is **less** than the end angle, and the start
image is **greater** than the end image.

The bug in #1827 would appear via condition number 1. I believe
the `if()` block was designed only for condition number 2, so
this fixes the issue.

### DIFF
--- a/tomviz/operators/SetTiltAnglesOperator.cxx
+++ b/tomviz/operators/SetTiltAnglesOperator.cxx
@@ -206,7 +206,7 @@ public:
         double delta =
           (endAngle->value() - startAngle->value()) / (end - start);
         double baseAngle = startAngle->value();
-        if (delta < 0) {
+        if (end < start) {
           int temp = start;
           start = end;
           end = temp;


psavery:
I still want to test this just a little more, but I believe it fixes the problem.

psavery:
Indeed, I believe this fixes the issues. For both of the two conditions listed above, the output tilt angles started positive and ended negative as expected, and the cross correlation image alignment worked fine.

I also noticed that the bug would occur if both the start angle and the start image were greater than the end angle and the end image, respectively. This fixes that as well.

Issue #1823, opened 2019-03-14
Weighted Backprojection size incorrectly calculated


ercius:
**Describe the bug**
The weighted back-projection (WBP) GUI includes a "Reconstruction size" option. This value should default to the current output of the pipeline before being reconstructed. However, the default value is linked to the original data set size before any operations.

**To Reproduce**
1. Load example tilt series
  Note: Tilt series images are 256x256
2. Select Tomography -- Bin Tilt Image x2
  Note: Tilt series images (pipeline output) are now 128x128
3. Select Tomography -- Weighted Backprojection
4. The size of the reconstruction is default set to 256
![image](https://user-images.githubusercontent.com/15657108/54392375-bb074300-4664-11e9-82e9-6f137472b677.png)
5. The final reconstruction size is 128x256x256.
  Note: The reconstruction looks right in this software because VTK shows the volume with non-square pixel sizes. I.e. The reconstruction has "Units and size" of: 128x256x256 nm

**Expected behavior**
The Reconstruction size should default to the input image size (after all pipeline transformations) along the Y direction where X is the tilt axis (horizontal) and Z is the tilt rotation array axis. The expected final reconstruction should be 128x128x128 pixels for the example tilt series after binning x2. The default behavior produces 128x256x256.

More information:
Non-square tilt series images will have the final reconstruction shape as [X, Y, Y]. You can test this by cropping a square tilt series before reconstruction along **only** the tilt axis (X-axis) direction. If I crop before reconstruction as:
![image](https://user-images.githubusercontent.com/15657108/54392764-abd4c500-4665-11e9-8c17-eecb28498b7f.png) 
and then reconstruct then I get a final reconstruction size of 54x128x128 if I set the *Reconstruction Size* to 128 in the WBP GUI. This is the expected behavior, but **not** the default behavior.
![image](https://user-images.githubusercontent.com/15657108/54392946-20a7ff00-4666-11e9-87b0-908945868110.png)

Windows 10, Dell XPS 13 laptop.

psavery:
Note to devs: the code block producing the default value is [here](https://github.com/OpenChemistry/tomviz/blob/master/tomviz/InterfaceBuilder.cxx#L201) (num-voxels-y).

psavery:
[...]

PR #1829, opened 2019-03-18
Use child data for operator defaults

psavery:
This is done in order to generate better default values
for operators such as the weighted back-projection.


psavery:
With this change, I get the correct default value for the weighted back-projection (mentioned in issue #1823). However, I want to look at it a little more and test it to make sure it is an appropriate fix.

psavery:
As far as I can tell, this works as intended, and fixes #1823.

I followed that pointer variable to make sure it doesn't affect functions elsewhere. It looks like it is only used in `InterfaceBuilder` for `GetExtent()` [here](https://github.com/OpenChemistry/tomviz/blob/master/tomviz/InterfaceBuilder.cxx#L202), which is what we were trying to fix. So I think this fix does not affect other functionality.

Issue #1797, opened 2019-02-05
Tilt axis alignment (manual) pixel/physical projection number


ercius:
**Describe the bug**
The projection number is also being affected by the pixel/physical drop down box in the GUI. The projection number should always be calibrated in "pixels." 

**To Reproduce**
1. Load example tilt series data set
2. Change Units and size to have a pixel size of 0.1 nm:
    1. Note that this is similar to what happens when you load an MRC file where the pixel size has been set to 1 as default.
![image](https://user-images.githubusercontent.com/15657108/52310676-143ed100-2959-11e9-8953-cc8df5ce2af6.png)
3. Open the Tilt axis alignment (manual) GUI
    1. Notice that "length unit" is set to Pixel and projection number is 37
4. Change Length Unit to physical
    1. The projection number is 3.7 now which is confusing
![image](https://user-images.githubusercontent.com/15657108/52310943-02116280-295a-11e9-9f24-ba5c7f151e7f.png)
    
**Expected behavior**
Projection number should always be in "pixel" number.

**Desktop (please complete the following information):**
Windows 10, Dell XPS13 laptop, Intel GPU

psavery:
I have a fix for this, but one question first.

@cryos, do you think that the slice numbers should be in pixels as well? It seems kind of strange to me that a `Slice No.` can be a decimal number. See image below. Slice numbers with the slice widget are always in pixels.

![image](https://user-images.githubusercontent.com/9558430/54715625-3f0c6f80-4b2a-11e9-8d20-57325c32a0cd.png)

cryos:
It is no longer a slice number if it is in physical units, it is a measurement. I think we should offer slice numbers **and** physical units, and let one set the other. I think having the slice number lets us see the actual number of voxels, but where appropriate turning that into the physical unit is also useful.

ercius:
In my opinion I would simplify this dialog and only display everything in
Pixels. Remove the physical option. There is not much of a reason for
physical measurements here.

cryos:
@psavery do as @ercius suggests then, I thought that people wanted physical units but I would rather keep the dialog simple if there is no strong need for physical units.

psavery:
@cryos Will do.

cryos:
This is fixed.

PR #1830, opened 2019-03-21
Always use pixels for tilt axis align proj. num

psavery:

### DIFF
--- a/tomviz/RotateAlignWidget.cxx
+++ b/tomviz/RotateAlignWidget.cxx
@@ -499,7 +499,7 @@ void RotateAlignWidget::onProjectionNumberChanged(int val)
   this->Internals->Ui.sliceView->GetRenderWindow()->Render();
 }
 
-void RotateAlignWidget::onRotationShiftChanged(double val)
+void RotateAlignWidget::onRotationShiftChanged(int val)
 {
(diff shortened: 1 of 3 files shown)


psavery:
Fixes first part of #1797. 

Note that I removed setting the single step number of the spinbox because it should always be 1 now (I think). The value should always be an `int` now, but I am still using `tomviz::DoubleSpinBox` instead of `QSpinBox`, because I thought there might be some overridden behavior by `tomviz::DoubleSpinBox` that we want (looks like there is some mouse press and mouse release overriding). Let me know if I can swap it out for QSpinBox instead.

psavery:
I'm now going to add another row to the widget so that slice numbers and physical numbers for the bottom three images are always displayed.

psavery:
Here's my work so far on displaying both slice numbers and physical units at the same time. Let me know if you have any suggestions.

Screenshot from 2019-03-21 09-37-12

psavery:
I removed the physical units entirely. An example image is below.

Screenshot from 2019-03-21 10-34-27

I want to fix one small bug I found as well... it appears that the top-left image starts at the 0 projection number rather than the one in the spinbox, but the image updates immediately after changing the spinbox. This bug was present before this PR.

Issue #1751, opened 2018-11-30
Calibrate example data sets


ercius:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you would like**
Can you add pixel sizes to the example data sets? This would make issues where the pixel and real space coordinates are confused more obvious during development and testing. i.e. #1749 

**Describe alternatives you have considered**
Or dont....

psavery:
I took the Co2P tilt series example from the Sci. data paper, manipulated it a little bit, and tried to calibrate the physical units. Attached is a sample EMD file (~4.8 megabytes unzipped, but github only lets me upload it zipped, I attached zip and gzip copies). I put the operations and my math below.

Original dimensions in pixels (pixel size is ~0.71 nm, according to the paper):
1157 x 1157 x 76

**Resample Operation:** 0.25 x 0.25 x 1.0
New dimensions in pixels: 289 x 289 x 76
New pixel size: 0.71 nm * 1157 px / 289 px = 2.84 nm per pixel

**Crop Operation:** X: 5 - 260, Y: 0 - 255, Z: 0 - 75
New dimensions in pixels: 256 x 256 x 76

**Convert Type Operation:** to UInt8 (to reduce final file size):
Dimensions unchanged

New physical dimensions (calculated via 2.84 nm per pixel * 256 pixels):
727.67 nm by 727.67 nm by 76 slices

Please let me know if any of the math looks wrong, or I should change any of the operations (maybe I could center the image more or something). 

[tilt_sample.emd.zip](https://github.com/OpenChemistry/tomviz/files/2992819/tilt_sample.emd.zip)
[tilt_sample.emd.gz](https://github.com/OpenChemistry/tomviz/files/2992820/tilt_sample.emd.gz)

psavery:
I shifted the volume, and I think these are centered just a little better.

[tilt_sample2.emd.zip](https://github.com/OpenChemistry/tomviz/files/2993091/tilt_sample2.emd.zip)
[tilt_sample2.emd.gz](https://github.com/OpenChemistry/tomviz/files/2993092/tilt_sample2.emd.gz)

psavery:
I made one more change to the tilt sample (I just rounded the physical units to two decimal places). Here is the new one.

I noticed that the Y-axis of the old tomviz tilt series sample appears to be flipped relative to the Y-axis of this new tilt sample (and the original file). Let me know if I need to flip the Y-axis to make it match.

[...]

PR #1831, opened 2019-03-22
Ensure tilt angles array is not sized too large

psavery:


psavery:
This fixes the issue I mention in #1751.

@cryos, what do you think of this fix?

cryos:
This doesn't make sense to me, we would never want to resize the tilt angles array to such a large number, where is that number coming from? Isn't vtkIdType a 64 bit unsigned int, and you are then comparing to a signed 32 bit int. What number is being generated, and where is it coming from?

tomviz/DataSource.cxx
```diff
@@ -42,18 +42,22 @@
...
 namespace {
-void createOrResizeTiltAnglesArray(vtkDataObject* data)
+void createOrResizeTiltAnglesArray(
+  vtkDataObject* data,
+  vtkIdType maxSize = std::numeric_limits<vtkIdType>::max())
 {
   auto fd = data->GetFieldData();
   if (!fd->HasArray("tilt_angles")) {
     int* extent = vtkImageData::SafeDownCast(data)->GetExtent();
     int numTiltAngles = extent[5] - extent[4] + 1;
```
cryos: I am guessing one of these numbers is coming out off, any realistic data set should have 1 - maybe a few thousand projections.

psavery:
@cryos Sorry, I submitted this PR in a rush before I left the office. I should have written some explanation to go with it.

What's happening in the issue is that, when the angles get set in the EMD reader [here](https://github.com/OpenChemistry/tomviz/blob/master/tomviz/EmdFormat.cxx#L664), the allocation for the angle array that occurs [here](https://github.com/OpenChemistry/tomviz/blob/master/tomviz/DataSource.cxx#L1168) is too large (for the tilt series example, the allocation is for 255 elements, when the number of tilt angles is only 74). That is because the allocation is made based upon the extent [here](https://github.com/OpenChemistry/tomviz/blob/master/tomviz/DataSource.cxx#L53) rather than the number of angles given in `setTiltAngles()`. This resulted in a lot of tilt angles of `0` after the first 74 tilt angles, so I wanted to provide a way for `setTiltAngles()` to communicate to `createOrResizeTiltAnglesArray()` how much it should allocate, that's what the `maxSize` parameter is for.

If `maxSize` was set by the caller, it would cause the angle array to allocate that many elements (74 in our case). And if `maxSize` was left to its default state, it would have no impact on the function (because `numTiltAngles < maxSize` would always be true). I used vtkIdType here because I looked up the type that `SetNumberOfTuples()` takes (I didn't actually notice that `numTiltAngles` was an int; I could change `maxSize` and `numTuples` to be ints instead).

So yeah, basically this PR was an attempt to provide a way to supply `createOrResizeTiltAnglesArray()` with the actual array size that was needed, so it wouldn't allocate too much space for the array.

An alternative one-line fix for this bug is to put `tiltAngles->SetNumberOfTuples(angles.size());` right after [this line](https://github.com/OpenChemistry/tomviz/blob/master/tomviz/DataSource.cxx#L1170). The only downside is that an array too large will usually be generated, and then it will be sized down immediately after. I could do this fix instead, let me know.

cryos:
createOrResizeTiltAnglesArray is a private function, setting std::numeric_limits<>::max() makes no sense, and the type used is larger than what can be accommodated by vtkImageData for extents (signed int). The number of slices in z should always match the number of tilt angles, I would like to understand why the vtkImageData passed in from EmdFormat is giving 255 for extents when it looks like it should have the correct number of slices.

We should be doing a sanity check, probably in the reader and/or the static set tilt angles method, to confirm the number of angles is equal to the number of slices in z before attempting to set them. I don't have time to look at this right now, but adding some of these sanity checks would be better.

psavery:
@cryos Okay. I think I have a much better fix for this. I'll submit the PR later tonight.

cryos:
It can wait until Monday, but I think I see the real issue as I was just taking a look - line 681 in EmdFormat.cxx performs the permute to swap axes long after the angles have been set. That block just needs moving up before the angles are set so that the number of slices is the same as the number of angles. If they ever differ there are likely to be other issues.

cryos:
PR #1832 works well here with your new tilt series. The logic was a little odd, I think this is simpler but could still do with a little more in future.

psavery:
@cryos Thanks for taking a look at this! Yeah, that change in the EMD reader was similar to what I was going to try this evening. Sorry I didn't notice this issue sooner.

PR #1832, opened 2019-03-23
Permute the axes before setting the angles

cryos:
The vtkImageData should be permuted before setting the angles, simplify
the logic a little as the reader implicitly knows if the data has tilt
angles if the array is kept in scope a little longer.

### DIFF
--- a/tomviz/DataSource.cxx
+++ b/tomviz/DataSource.cxx
@@ -48,9 +48,9 @@ namespace {
 void createOrResizeTiltAnglesArray(vtkDataObject* data)
 {
   auto fd = data->GetFieldData();
+  int* extent = vtkImageData::SafeDownCast(data)->GetExtent();
+  int numTiltAngles = extent[5] - extent[4] + 1;
   if (!fd->HasArray("tilt_angles")) {
-    int* extent = vtkImageData::SafeDownCast(data)->GetExtent();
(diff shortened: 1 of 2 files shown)


psavery:
@cryos I think this works fine in my tests. Thanks for the PR! I am able to open up the new tilt series sample and run reconstructions correctly now, it seems.

One question: in the old tilt series example, after we marked the data as a tilt series, the Z in `Units and Size` in the data properties panel would be set to the number of tilt angles (74, in our case). In our new example with the EMD reader, this gets set to 148, I think because of the spacing. Do we want it to display 74 for the Z? If so, I think we can fix it by putting something like `image->GetSpacing()[2] = 1.0;` right after [here](https://github.com/OpenChemistry/tomviz/pull/1832/files#diff-aa34731822e5a2f4621038caf01d6b0fR680) in the EMD reader.

psavery:
Actually, on a second look, it looks like "Mark Data as Tilt Series" doesn't affect the units. But I am still wondering if we should get the EMD reader to set the spacing to 1 so that the Z units match the number of tilt angles. If not, it's fine...

cryos:
The spacing doesn't really have much meaning for tilts. I think we should do it after the release, but the dialog should be better about tilt series not being a distance whereas volumes are. Leaving any tilt series at the default spacing of 1 is probably the most reasonable thing to do, that dialog can be improved.

Issue #1833, opened 2019-03-24
Tilt Series State Lost when Operating on Tilt Series Loaded from EMD Files


psavery:
**Describe the bug**
If a file that explicitly contains a tilt series is opened up (such as an EMD file), and then an operation is performed such as `Bin Tilt Images 2x`, the tilt series state is forgotten. Many operations seem to do this, including `Bin Tilt Images 2x` (pre-processing operation) and `Invert Data` (data transform).

**To Reproduce**
1. Unzip the attached file, and open it with tomviz. The data should already be recognized as a tilt series.
[tilt_sample4.emd.gz](https://github.com/OpenChemistry/tomviz/files/3000707/tilt_sample4.emd.gz)
2. Perform a `Bin Tilt Images 2x` operation.
3. Notice that the tilt series state is forgotten (for instance, `Mark Data As Tilt Series` is no longer grayed out).

**Expected behavior**
Do not forget the tilt series state after these operations.

**Desktop (please complete the following information):**
 - Operating system: Ubuntu 18.04
 - Tomviz version: 1.5.1-135-g56e21824

psavery:
This bug might only occur for the EMD reader. But it can be looked out for in other readers in the future.

cryos:
I am closing this for now, only the EMD reader brings in angles, we should ensure this works with future readers.

PR #1834, opened 2019-03-25
Set tilt series type field data for emd reader

psavery:
Previously, when reading an EMD file that contains a tilt series,
the data source type field data was not set in the vtkDataObject.
This would cause tomviz to forget that it contains tilt series
data after an operation was performed on it.

This fixes the issue.

### DIFF
--- a/tomviz/EmdFormat.cxx
+++ b/tomviz/EmdFormat.cxx
@@ -11,6 +11,7 @@
 #include <vtkNew.h>
 #include <vtkPointData.h>
 #include <vtkTrivialProducer.h>
+#include <vtkTypeInt8Array.h>
 
 #include <vtkSMSourceProxy.h>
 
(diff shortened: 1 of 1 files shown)


psavery:
This fixes issue #1833.

tomviz/EmdFormat.cxx
```diff
@@ -678,6 +679,16 @@ bool EmdFormat::read(const std::string& fileName, vtkImageData* image)
     permute->Update();
     image->ShallowCopy(permute->GetOutput());
     DataSource::setTiltAngles(image, angles);
+
+    // Now set the field data to preserve the tilt series state
+    auto typeArray = vtkSmartPointer<vtkTypeInt8Array>::New();
```
cryos: `vtkNew<vtkTypeInt8Array> typeArray;` would be a little more compact

PR #1838, opened 2019-04-04
Open up EMD files with units etc in the file

cryos:
This addresses issue #1751 in conjunction with the superbuild update.

### DIFF
--- a/tomviz/MainWindow.cxx
+++ b/tomviz/MainWindow.cxx
@@ -566,7 +566,7 @@ void MainWindow::openFiles(int argc, char** argv)
 void MainWindow::openTilt()
 {
   QString path = QApplication::applicationDirPath() + "/../share/tomviz/Data";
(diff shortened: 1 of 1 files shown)


psavery:
I took a quick look at the nightly. It looks to me that the EMD samples are working fairly well!

One question: I noticed that if you open up the tilt sample via `Sample Data` -> `Star Nanoparticle (Tilt Series)`, it automatically runs the `setTiltAngles` operator on it [here](https://github.com/cryos/tomviz/blob/afb08fc3fc9fc8e8d6ef9db8478492abf09c99be/tomviz/MainWindow.cxx#L573), and it sets the tilt angles from -73 to 73 with a spacing of 2. This is different than the tilt angles that we set in the emd file (angles are -75 to 75 with a spacing of 2, leaving out angles 23 and 25). Should we modify this?

cryos:
Yes, you can delete the operator and things work as expected. I merged a branch to stop setting the angles and it is building now.

PR #1841, opened 2019-04-05
Upgrade ParaView to the Latest Version

psavery:
This is a work in progress.


CMakeLists.txt
```diff
@@ -65,10 +65,7 @@ if(NOT PARAVIEW_ENABLE_PYTHON)
     "Tomviz requires PARAVIEW_ENABLE_PYTHON to be enabled. "
     "Please rebuild ParaView with PARAVIEW_ENABLE_PYTHON set to TRUE.")
 endif()
-if(NOT PARAVIEW_QT_VERSION STREQUAL "5")
-  message(FATAL_ERROR
-    "Tomviz requires PARAVIEW_QT_VERSION to be 5, please rebuild ParaView.")
-endif()
+# TODO: is there a way in the latest ParaView to check Qt version used?
```
cryos: We should just remove this, ParaView doesn't build against old Qt 4 anymore.

CMakeLists.txt
```diff
@@ -65,11 +65,24 @@ if(NOT PARAVIEW_ENABLE_PYTHON)
     "Tomviz requires PARAVIEW_ENABLE_PYTHON to be enabled. "
     "Please rebuild ParaView with PARAVIEW_ENABLE_PYTHON set to TRUE.")
 endif()
-if(NOT PARAVIEW_QT_VERSION STREQUAL "5")
-  message(FATAL_ERROR
-    "Tomviz requires PARAVIEW_QT_VERSION to be 5, please rebuild ParaView.")
+
+load_cache(${ParaView_DIR}
```
mathstuf: This only works in the superbuild by accident since we use ParaView's build tree there. An install tree will be…well, it doesn't have `CMakeCache.txt`. [This MR](https://gitlab.kitware.com/vtk/vtk/merge_requests/5373) should be used in the future.

psavery: Our past builds of tomviz have also relied on `load_cache`. I believe the build instructions require linking to a build tree, not an install tree. But we can look into using the variable instead when that gets merged.

cryos: Yes, we haven't worried about building against an installed ParaView. I think there are some other concerns if we need to do that, such as the use of third party headers/libraries too.

cryos:
It is worth noting at a high level that we need some fairly specific flags enabled for Tomviz to work, and I don't know how they were mapped. They mainly revolve around enabling Python and thread safety for Python in ParaView/VTK.

CMakeLists.txt
mathstuf: > <using build tree>

This is likely because on macOS, ParaView didn't install a useful SDK. It now does.

> I think there are some other concerns if we need to do that, such as the use of third party headers/libraries too.

VTK/ParaView's third party libraries no longer conflict with anything (headers, libraries, or symbols), so that concern should be resolved.

mathstuf:
> It is worth noting at a high level that we need some fairly specific flags enabled for Tomviz to work, and I don't know how they were mapped. They mainly revolve around enabling Python and thread safety for Python in ParaView/VTK.

If these didn't get ported up, we should look at adding them back. Enabling Python is now guaranteed by `VTK_WRAP_PYTHON` or `PARAVIEW_ENABLE_PYTHON`. Not sure about the threading option. That may have gone away (though I'd be more interested in knowing if it is still actually necessary before adding it back if missing).

mathstuf:
The classes in the `TomvizExtensions` module should probably have `_EXPORT` macros on them for Windows.

psavery:
@mathstuf Done. Thanks for taking a look, Ben!

tomviz/CMakeLists.txt
```diff
@@ -552,7 +551,7 @@ target_link_libraries(tomviz PRIVATE tomvizlib ${OPENGL_LIBRARIES})
 
 target_link_libraries(tomvizlib
   PUBLIC
-    pqApplicationComponents
+    ParaView::pqApplicationComponents
     vtkPVServerManagerRendering
```
mathstuf: These will also need updated.

tomviz/pvextensions/Plugin/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+paraview_add_plugin(TomvizExtensions
+  VERSION "1.0"
+  SERVER_MANAGER_XML TomvizExtensions/TomvizExtensions.xml
```
mathstuf: The `paraview_add_server_manager_xmls` makes this unnecessary.

tomviz/CMakeLists.txt
```diff
@@ -4,7 +4,6 @@ include_directories(SYSTEM
   ${Qt5Network_INCLUDE_DIRS}
```
mathstuf: Using `Qt5::Network` below should make these include directories unnecessary.

psavery:
@mathstuf Done and done. Thanks again!

psavery:
Good news! It now compiles, and the few things I've tested (opening data, setting tilt angles, running a reconstruction, etc.) seem to work!

I do get several compile time and runtime deprecated function warnings, though. Some of them can be seen below.

Compile time warnings:
```
tomviz/acquisition/AcquisitionWidget.cxx:48:36: warning: ‘QVTKInteractor* QVTKOpenGLWidget::GetInteractor()’ is deprecated [-Wdeprecated-declarations]
   m_ui->imageWidget->GetInteractor()->SetInteractorStyle(

tomviz/acquisition/AcquisitionWidget.cxx:221:38: warning: ‘vtkRenderWindow* QVTKOpenGLWidget::GetRenderWindow()’ is deprecated [-Wdeprecated-declarations]
   m_ui->imageWidget->GetRenderWindow()->Render();
```

Runtime warnings:
```
(   0.093s) [paraview        ]   QVTKOpenGLWidget.cxx:141   WARN| QVTKOpenGLWidget::SetRenderWindow was deprecated for VTK 8.3 and will be removed in a future version.  Use QVTKOpenGLWidget::setRenderWindow instead.
```

psavery:
I am also getting these warnings:

```
tomviz/tomvizPythonConfig.h:17:0: warning: "PARAVIEW_VERSION_MAJOR" redefined
 #define PARAVIEW_VERSION_MAJOR ""
/home/patrick/src/paraview/build/ParaViewCore/Core/vtkPVConfig.h:35:0: note: this is the location of the previous definition
 #define PARAVIEW_VERSION_MAJOR 5

tomviz/tomvizPythonConfig.h:18:0: warning: "PARAVIEW_VERSION_MINOR" redefined
 #define PARAVIEW_VERSION_MINOR ""
/home/patrick/src/paraview/build/ParaViewCore/Core/vtkPVConfig.h:36:0: note: this is the location of the previous definition
 #define PARAVIEW_VERSION_MINOR 6
```

That config file is trying to define them via `@PARAVIEW_VERSION_MAJOR@` and `@PARAVIEW_VERSION_MINOR@`. Maybe we shouldn't `#define` those since it looks like they are already defined.

psavery:
All new compile warnings are now gone! And most of the program runs as expected. But when I run the contour module, this appears:

```
[paraview        ]vtkSMFieldDataDomain.cx:130    ERR| vtkSMFieldDataDomain (0x55f1e3827620): `vtkSMFieldDataDomain` is being used on a `vtkSMStringVectorProperty`. This is no longer needed or supported. Simply remove it from your XML configuration for property SelectInputScalars
```

I'll be looking into this next.

psavery:
@cryos I think this is ready for a review.

When I built ParaView, I used the same cmake flags mentioned in the [build instructions](https://github.com/OpenChemistry/tomviz/blob/master/BUILDING.md#initial-build), and I didn't seem to encounter any issues.

For my Ubuntu 18.04 computer at least, it looks like tomviz now builds and runs. I haven't tested every operation in tomviz, but the ones that I have tested seem to work.

We may also consider looking into getting the ParaView python information from the Python3::Python target, instead of from the ParaView build cache.

Issue #1842, opened 2019-04-17
Tilt axis (manual) uses untransformed data


ercius:
The transformed data during pre-processing (e.g. sqrt, reinterpret signed to unsigned, etc.) is not used in the Tilt Axis (manual) GUI.

1) Load a data set (star tilt series)
2) Apply an obvious transform that modifies the intensities. Sqrt in the default custom transform works.
3) Start the Tilt Axis Alignment (manual)
4) The histogram changes to the old original values. The image on the screen is hard to see (sometimes all black or white) because the colormap is confised as to which data is being shown (see screenshot).
![image](https://user-images.githubusercontent.com/15657108/56311590-4744d400-6103-11e9-8ddd-28073e2143c2.png)

This transform should use the unsigned integers. Im not sure if it's just a display issue or if the transform is also applied to the original data. I think it applies the rotate and shift to the transformed data, but that should be verified.

tomviz 1.6.0, WIndows 10, DEL XPS 13 laptop, intel graphics card.

psavery:
I see that it looks like the histogram also gets reset for some other operators with edit dialogs as well, such as `Image Alignment (Manual)`. I'll take a look at it.

psavery:
@ercius I discussed in #1843 that it appeared that the parent's color map was used instead of the child's color map for the Tilt Axis Alignment (manual) operator. I fixed it there.

I also discussed in that PR that it appears the reason the histogram is changing in the background is just because the parent becomes the active data source (same effect as left clicking on the top row in the pipeline diagram). I don't think it's actually modifying any values, so it might be okay to leave it.

Can you check on a nightly build tomorrow or another day, and confirm that everything is working properly?

ercius:
Thanks for confirming this issue. I'll check it fixed when I get a chance.

cryos:
Is this fixed? It would be good to ensure we have it working in master before our next release.

ercius:
This is working 50%. Only the histogram in the main window is going back to use the original data's values. The GUI uses the transformed  data.

psavery:
As mentioned above, I think the histogram is changing because the active data source is being switched to the parent rather than the child, and this happens for all operators. I can try to see if I can fix it...

psavery:
The PR linked above is a good fix for this, I think. It should fix the rest of the issue.

PR #1843, opened 2019-04-17
Fix Rotate Align Widget's Color Map

psavery:
In the Rotate Align Widget, use a child data source's color
map when it is available so that the color map will match
that of the view.

### DIFF
--- a/tomviz/RotateAlignWidget.cxx
+++ b/tomviz/RotateAlignWidget.cxx
@@ -394,10 +394,15 @@ RotateAlignWidget::RotateAlignWidget(Operator* op,
   this->Internals->mainSliceMapper->SetInputData(this->Internals->m_image);
   this->Internals->mainSliceMapper->Update();
 
-  DataSource* ds = op->dataSource();
-  if (!ds) {
+  // Use a child data source if one is available so the color map will match
+  DataSource* ds;
+  if (op->childDataSource())
+    ds = op->childDataSource();
+  else if (op->dataSource())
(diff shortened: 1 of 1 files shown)


psavery:
This is a WIP to fix #1842. You can see the results of this PR in the images attached.

It looks to me that we just needed to make sure we used the child data source's color map in the rotate align widget. In this PR's changes, the `ds` variable that gets set is not used for anything other than the color map, so I don't think modifying it (as I have done) will have any other effect.

Note in the images that the histogram in the background still matches that of the parent data source. But I think that is just because the active data source is being changed, not because the histogram is being modified. The parent active data source gets loaded [here](https://github.com/OpenChemistry/tomviz/blob/0f095c57c127cba91acd9691a8e4fc2794950512/tomviz/AddPythonTransformReaction.cxx#L171) and set [here](https://github.com/OpenChemistry/tomviz/blob/0f095c57c127cba91acd9691a8e4fc2794950512/tomviz/operators/EditOperatorDialog.cxx#L92), I believe.

**Rotate align widget after performing a `Square Root Data`:**
Screenshot from 2019-04-17 18-47-27

**Rotate align widget after performing a `Square Root Data` followed by an `Invert Data`:**
Screenshot from 2019-04-17 18-47-52

**After setting 25 and 25 to the tilt and rotation alignment axes, respectively, in the image above:**
Screenshot from 2019-04-17 18-49-30

Maybe someone can confirm that everything looks as it should?

PR #1844, opened 2019-04-18
H5CPP Stride and Data Exchange

psavery:
This provides a simple reader for the Data Exchange format. It currently assumes all `.h5` files are of the Data Exchange format.

It also provides a stride option to h5cpp, because it was needed to read the large Data Exchange files.

@cryos The stride number is currently hard-coded to 2 [here](https://github.com/OpenChemistry/tomviz/compare/master...psavery:h5cpp-stride-data-exchange?expand=1#diff-acf855f951637487160f365fb8982c57R64). It applies a stride of `2` to all directions. I can load in the data exchange file sample. A sample image can be seen below.

Screenshot from 2019-04-18 12-01-11

### DIFF
--- a/tomviz/LoadDataReaction.cxx
+++ b/tomviz/LoadDataReaction.cxx
@@ -4,6 +4,7 @@
 #include "LoadDataReaction.h"
 
 #include "ActiveObjects.h"
+#include "DataExchangeFormat.h"
 #include "DataSource.h"
 #include "EmdFormat.h"
 #include "FileFormatManager.h"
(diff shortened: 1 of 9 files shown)


cryos:
This works, we can't keep the hardwired stride but it is good to get up and running. We need to write at least a basic version of the file too. I am testing the move to ParaView master right now, so this will have to wait a little, but if you could add a basic writer, and maybe only stride if the dimensions are over 1100 in any direction for now perhaps - this would let us keep moving forward while we think about UI/hooking things up for upfront subsampling.

psavery:
@cryos I modified it as requested, and I added a basic Data Exchange writer. It currently just writes the data (named "data") to the group "/exchange". No attributes or metadata are currently written.

I tested it by saving a file, opening it up, and then saving it again. h5diff reports them to be identical.

tomviz/LoadDataReaction.cxx
```diff
@@ -115,6 +116,7 @@ QList<DataSource*> LoadDataReaction::loadData()
           << "JPeg Image files (*.jpg *.jpeg)"
           << "PNG Image files (*.png)"
           << "TIFF Image files (*.tiff *.tif)"
+          << "Data Exchange files (*.h5)"
```
cryos: This is just the HDF5 generic extension, we probably want to avoid claiming it is just Data Exchange.

PR #1849, opened 2019-04-25
Add the plugin back in and force it to be static

psavery:
A recent ParaView commit allows the plugin to be forced
to be static. Utilize that so we can link to the plugin.

Hopefully, the dashboards will pass with this.


psavery:
I'm gonna give this a try on the dashboards first, along with [this one](https://github.com/OpenChemistry/tomviz-superbuild/pull/277).

PR #1852, opened 2019-05-08
Add Generic HDF5 Volume Reader

psavery:
This PR depends on #1844, and this PR only has one extra commit.

This uses the current `DataExchangeFormat` class as a generic HDF5 volume reader. A generic HDF5 reader like this might be able to replace the `DataExchangeFormat`, but the reader would likely need to be renamed.

If you open a file that contains just one 3D dataset, it automatically loads it in as a volume.

If you open a file that contains multiple 3D datasets, the following dialog pops up, from which you can select a volume like so:

Screenshot from 2019-05-08 10-30-31

image

Screenshot from 2019-05-08 10-31-40

### DIFF
--- a/tomviz/CMakeLists.txt
+++ b/tomviz/CMakeLists.txt
@@ -67,6 +67,8 @@ set(SOURCES
   ExportDataReaction.h
   FileFormatManager.cxx
   FileFormatManager.h
+  GenericHDF5Format.cxx
+  GenericHDF5Format.h
   GradientOpacityWidget.h
   GradientOpacityWidget.cxx
   HistogramManager.h
--- a/tomviz/DataExchangeFormat.cxx
+++ b/tomviz/DataExchangeFormat.cxx
@@ -4,6 +4,7 @@
 #include "DataExchangeFormat.h"
 
 #include "DataSource.h"
+#include "GenericHDF5Format.h"
 
 #include <h5cpp/h5readwrite.h>
 #include <h5cpp/h5vtktypemaps.h>
@@ -33,18 +34,6 @@ void ReorderArrayC(T* in, T* out, int dim[3])
   }
 }
 
-template <typename T>
-void ReorderArrayF(T* in, T* out, int dim[3])
-{
-  for (int i = 0; i < dim[0]; ++i) {
-    for (int j = 0; j < dim[1]; ++j) {
-      for (int k = 0; k < dim[2]; ++k) {
-        out[(k * dim[1] + j) * dim[0] + i] = in[(i * dim[1] + j) * dim[2] + k];
-      }
-    }
-  }
-}
-
 bool DataExchangeFormat::read(const std::string& fileName, vtkImageData* image)
 {
   using h5::H5ReadWrite;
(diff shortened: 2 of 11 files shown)


cryos:
Avoid merging master in to branches where possible, it is best to just do it once at the end if there is a conflict we need to resolve. I still want to keep the Data Exchange format for reading and writing, and think we should probably merge that as-is and develop a more generic reader that will not expect any structure.

I would say the dialog can be a little less wordy too, probably load the volume if there is only one as you already do, otherwise "Choose a volume to load:" then a combobox with names.

psavery:
I force-pushed this to get rid of the history. I put those previous additions in a `GenericHDF5Format` class. The `GenericHDF5Format` is now used by default for reading `.h5` files, and if the file contains a dataset at `/exchange/data`, it assumes it is a `DataExchangeFormat` file and reads it with the data exchange format reader.

The `GenericHDF5Format` has the same functionality as above where it will look for all datasets that have 3 dimensions. If there is only one, it loads it automatically. If there are more than one, it asks the user to choose:
Screenshot from 2019-05-08 15-16-00

There is no writer for the `GenericHDF5Format`. If someone tries to write an `h5` file, it writes it in `DataExchangeFormat`.

psavery:
The EMD format, DataExchange format, and GenericHDF5 formats all used the same copy-pasted code for reading in volumes. I put that copy-pasted code into a function in GenericHDF5 format that they all call now (details can be seen in the commit message) to reduce the amount of copy-pasted code.

I probably only need to fix one more thing for this branch. If you open an "h5" file that has multiple datasets in it, the pasted warning gets printed to the terminal. This is not really a problem since it is just warning us that `/exchange/data` wasn't found, which is what we are expecting to happen. But I should probably eliminate unnecessary warnings from the terminal.
```
HDF5-DIAG: Error detected in HDF5 (1.10.3) thread 0:
  #000: ../VTK/ThirdParty/hdf5/vtkhdf5/src/H5Odeprec.c line 147 in vtkhdf5_H5Oget_info_by_name1(): can't get info for object: '/exchange/data'
    major: Object header
    minor: Can't get value
  #001: ../VTK/ThirdParty/hdf5/vtkhdf5/src/H5Oint.c line 2376 in H5O__get_info_by_name(): can't retrieve object info
    major: Object header
    minor: Can't get value
  #002: ../VTK/ThirdParty/hdf5/vtkhdf5/src/H5Gloc.c line 731 in vtkhdf5_H5G_loc_info(): can't find object
    major: Symbol table
    minor: Object not found
  #003: ../VTK/ThirdParty/hdf5/vtkhdf5/src/H5Gtraverse.c line 851 in vtkhdf5_H5G_traverse(): internal path traversal failed
    major: Symbol table
    minor: Object not found
  #004: ../VTK/ThirdParty/hdf5/vtkhdf5/src/H5Gtraverse.c line 741 in H5G__traverse_real(): component not found
    major: Symbol table
    minor: Object not found
Failed to get H5O info by name
```

psavery:
That error printing was fixed in the last commit.

PR #1855, opened 2019-05-09
For HDF5, choose stride for large dataset

psavery:
This pops up a small dialog when the dataset is large where the user can choose a stride. In this case (for HDF5), if any of the dimensions are greater than 1100, the dialog will appear.

This isn't a comprehensive solution for other formats including TIFF and RAW, yet. But maybe we can discuss the dialog and add a common function they will all call when choosing a stride.

image

### DIFF
--- a/tomviz/GenericHDF5Format.cxx
+++ b/tomviz/GenericHDF5Format.cxx
@@ -11,7 +11,9 @@
 #include <QComboBox>
 #include <QDialog>
 #include <QDialogButtonBox>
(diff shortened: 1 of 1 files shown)


psavery:
A note: we will be moving towards automatically striding for large datasets, and only reading the whole dataset if the user asks for it.

psavery:
Closing this in favor of more general solutions that will hopefully come from PR #1857.

PR #1856, opened 2019-05-10
Add voxels and size to data properties panel

psavery:
The voxels are obtained via `vtkPVDataInformation::GetNumberOfPoints()`
The size is obtained via `vtkPVDataInformation::GetMemorySize()`

You can see some examples attached below.

I'm not sure if the `Units and Size` is the best place to put these. Maybe they should be moved elsewhere and formatted better. Let me know.

Screenshot from 2019-05-10 12-42-13

Float example:
===========
Screenshot from 2019-05-10 12-43-00

### DIFF
--- a/tomviz/DataPropertiesPanel.cxx
+++ b/tomviz/DataPropertiesPanel.cxx
@@ -169,7 +169,7 @@ QString getMemSizeString(vtkSMSourceProxy* proxy)
   // GetMemorySize() returns kilobytes
   size_t memSize = info->GetMemorySize() * 1000;
 
-  return "Size: " + getSizeNearestThousand(memSize, true);
+  return "Memory: " + getSizeNearestThousand(memSize, true);
 }
 
 } // namespace
--- a/tomviz/DataPropertiesPanel.cxx
+++ b/tomviz/DataPropertiesPanel.cxx
@@ -132,14 +132,14 @@ QString getDataDimensionsString(vtkSMSourceProxy* proxy)
 }
 
 template <typename T>
-QString getSizeNearestThousand(T num)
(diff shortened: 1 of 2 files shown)


tomviz/DataPropertiesPanel.cxx
```diff
@@ -131,6 +131,43 @@ QString getDataDimensionsString(vtkSMSourceProxy* proxy)
...
+  QString ret;
+  if (num < 1e3)
+    ret = QString::number(num) + " B";
+  else if (num < 1e6)
+    ret = QString::number(num / 1e3, format, prec) + " K";
+  else if (num < 1e9)
+    ret = QString::number(num / 1e6, format, prec) + " M";
+  else if (num < 1e12)
+    ret = QString::number(num / 1e9, format, prec) + " G";
+  else
```
cryos: We could add T, and why not just leave the final statement as / 1e12 etc and then let it be bigger if it somehow loads?

cryos:
I think we should replace size (ambiguous/overloaded) with memory and make it clearer by following M etc with a B, i.e. MB, GB, etc.

tomviz/DataPropertiesPanel.cxx
psavery: Done.

psavery:
Done. Here is what they look like now.

Screenshot from 2019-05-10 13-17-57
Screenshot from 2019-05-10 13-18-22

cryos:
You know, I would put those two right underneath dimensions, otherwise I think this is good.

psavery:
Okay.

psavery:
Here's what they look like now:
Screenshot from 2019-05-10 15-33-11

Screenshot from 2019-05-10 15-33-36

PR #1857, opened 2019-05-11
Show original voxels and memory

psavery:
If a dataset was subsampled by striding, this PR causes the original (not subsampled) voxels and memory size to be displayed below the subsampled voxels and memory. This does not currently allow the data to be reloaded without subsampling, but it is intended to work towards that.

This is accomplished primarily due to [this commit](https://github.com/OpenChemistry/tomviz/commit/b922ac89d8b0e15de7979c2c66e7cd0a94069b82), where the stride can be saved in the vtkFieldData in the vtkDataObject. The next commit saves the stride in the GenericHDF5Reader class, and then the next commit reads the stride in the DataPropertiesPanel.

Some example images can be seen below.

Example with stride == 2
===================

Screenshot from 2019-05-11 13-04-49

Example with no subsampling
=======================
Screenshot from 2019-05-11 13-05-31

Depends on: #1856

### DIFF
--- a/tomviz/DataPropertiesPanel.cxx
+++ b/tomviz/DataPropertiesPanel.cxx
@@ -169,7 +169,7 @@ QString getMemSizeString(vtkSMSourceProxy* proxy)
   // GetMemorySize() returns kilobytes
   size_t memSize = info->GetMemorySize() * 1000;
 
-  return "Size: " + getSizeNearestThousand(memSize, true);
+  return "Memory: " + getSizeNearestThousand(memSize, true);
 }
 
 } // namespace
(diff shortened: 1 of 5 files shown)


psavery:
I hope this isn't becoming too cluttered. I wonder if it would be good to have a button or an option somewhere to display additional information about the data if the user wants more info, and then put the original voxels and memory in there?

Maybe right-click the dataset in the pipeline and click "Properties" or something like that?

cryos:
Let's take a look at layout so that this doesn't get too crowded, I think it is good but we could keep tings more compact hopefully.

psavery:
If we want to finish this, we will need to refactor it to take into account volume subsample selections, and we should use the API for getting stride/volume subsampling that was added in #1924

Issue #1858, opened 2019-05-15
Hard crash with NPY files and state files


ercius:
**Describe the bug**
NPY files are not treated the same as other types of data files. One example is that they can not be used with state files. If you load a numpy (NPY) file and then save a state file then that state file can not be loaded back into Tomviz. Tomviz hard crashes

**To Reproduce**
1. Create a numpy file using python and numpy with a 3D array of any size or data type.
```python
import numpy as np
np.save('output.npy',np.ones((10,10,10)))
```
2. Load the npy file into tomviz
3. Save a state file
4. Load the state file
5. Tomviz crashes without any output

**Expected behavior**
No crashing! :)

**Screenshots**

**Desktop (please complete the following information):**
 - Operating system: Window 10
 - Tomviz version: 1.6.0 (a separate user also reported this occurred in 1.5.1)

**Hardware (please complete the following information):**
 - Device: Dell XPS 13
 - Graphics card: Intel

psavery:
Hmm. Well, on Ubuntu 18.04, at least, it looks like it is crashing because it is attempting to read the npy file with a ParaView reader instead of with our python reader.

The options [here](https://github.com/OpenChemistry/tomviz/blob/60b1e8bccfcb926b6f86388754f6396ebe8ff2f7/tomviz/LoadDataReaction.cxx#L233) do contain `reader`, (because of [here](https://github.com/OpenChemistry/tomviz/blob/60b1e8bccfcb926b6f86388754f6396ebe8ff2f7/tomviz/modules/ModuleManager.cxx#L974)), but `name` is empty, resulting in a crash.

If I force it to run through [this block](https://github.com/OpenChemistry/tomviz/blob/60b1e8bccfcb926b6f86388754f6396ebe8ff2f7/tomviz/LoadDataReaction.cxx#L256) instead, it seems to load fine for me. So maybe we need to modify that logic a little bit.

PR #1860, opened 2019-05-16
Give python readers precedence over json options

psavery:
Currently, the two python readers that we have are for
".npy" files and ".mat" files. If a state file is loaded
that contains data from one of these two formats, it
attempts to read the files with ParaView (via the
"reader" in the json options) instead of with the
python readers.

Give the python readers precedence over the "reader"
in the json options to fix this issue.

Fixes: #1858


psavery:
By the way, you have to load the state file by first loading tomviz and then doing "File" -> "Load State".

If you load it by doing something like `tomviz statefile.tvsm`, you will get this message:
```
The Python reader for this file type hasn't loaded yet. Please try again in a few seconds
```

PR #1863, opened 2019-06-04
Updated custom range selector for float

bnmajor:
Updated custom range selector to offer six decimal precision for float values.

### DIFF
--- a/tomviz/HistogramWidget.cxx
+++ b/tomviz/HistogramWidget.cxx
@@ -37,6 +37,7 @@
 #include <vtkSMTransferFunctionPresets.h>
 #include <vtkSMTransferFunctionProxy.h>
 #include <vtkSMViewProxy.h>
+#include <vtkType.h>
 
 #include <QCheckBox>
 #include <QColorDialog>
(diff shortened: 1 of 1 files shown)


tomviz/HistogramWidget.cxx
```diff
@@ -401,6 +402,15 @@ void HistogramWidget::onCustomRangeClicked()
   double maxRange[2];
   activeDataSource->getRange(maxRange);
 
+  // Get the type of the active scalar
+  auto scalar = activeDataSource->activeScalars();
+  auto array = activeDataSource->getScalarsArray(scalar);
+  auto dataType = array->GetDataType();
+  int precision = 2;
```
cryos: So, I am not sure I understand why we would use precision of 2 for integral types, and six for float. We also support doubles (although discourage it). I would prefer to keep it simple and offer up to 6 for all, or check for double/float offer 6 and offer 0 decimal places for integral types.

cjh1: From the issue it sounded like we want 6, just for float ( and yes double ). 2 is the current default.

cryos: I think the default was more of an oversight, let's use 0 as our default here (for integral types), and set it to 6 for float and double.

cjh1: Sounds like a good plan.

cryos:
Looks great, thanks!

Issue #1846, opened 2019-04-19
Tomviz is no longer able to build on Mac


jtschwar:
**Describe the bug**

Now whenever I try to build tomviz I get this error message:

CMake Error at CMakeLists.txt:74 (message):
  Failed to get PARAVIEW_PYTHON_VERSION

-- Configuring incomplete, errors occurred!
See also "/Users/hovdengroup/tomviz-build/CMakeFiles/CMakeOutput.log".

I can provide my cmakeoutput.log if necessary

**To Reproduce**
1. Go to: tomviz-build directory
2. Click on: run: cmake --build .
3. See error

CMake Error at CMakeLists.txt:74 (message):
  Failed to get PARAVIEW_PYTHON_VERSION

-- Configuring incomplete, errors occurred!
See also "/Users/hovdengroup/tomviz-build/CMakeFiles/CMakeOutput.log".

I can provide my cmakeoutput.log if necessary

********Expected behavior***********

**Screenshots**

**Desktop (please complete the following information):**
 - Operating system: macOS Mojave (10.14)
 - Tomviz version: 1.6

**Hardware (please complete the following information):**
 - Device: Macbook Pro and iMac

**Additional context**

psavery:
@jtschwar Yeah, we are currently in the process of upgrading tomviz to use the latest ParaView. This will probably be a work in progress for the next week.

It may work on the mac if you build the master branch of ParaView and use that instead of the older version of ParaView. But we will have greater certainty and perhaps clearer instructions sometime next week.

cryos:
You can develop on the 1.6.0 tag if you want to continue working, this will be resolved soon but if I was you I would just stick to the 1.6.0 tag which is pretty close to master without the changes for ParaView master.

jtschwar:
Sounds good, thanks!

jtschwar:
Checking in on this issue. Still having trouble building the latest tomviz.

cryos:
Yes, you should be able to build against ParaView master now, but we are still iterating on getting everything working. You can see [working builds on the buildbot](https://open.cdash.org/index.php?project=TomViz) machines for all platforms. We have disabled the ParaView plugin for now, and so slice and contour modules are not working in the short term. I am working on solving that. I would still develop on the 1.6.0 tag if I was you, there are no major changes to Tomviz code that would prevent from merging your work.

`git checkout 1.6.0 && git checkout -b working-brach`

[...]

PR #1865, opened 2019-06-05
Disable all ParaView plugins by default

cryos:


cryos:
The new build flag for ParaView, I tried merging master into release and pushing, but would appreciate one of you verifying that ParaView release tip builds OK, and that Tomviz works as expected with that build @cjh1 @bnmajor

bnmajor:
I was able to build ParaView with the new build flag and Tomviz worked perfectly.

cryos:
Thank you for verifying this should fix #1846

PR #1868, opened 2019-06-14
Color Preset currently displays a smaller dialog box with a bas…

bnmajor:
Currently displays filler data and has no functionality. The initial size could easily be adjusted if it should be smaller, and I'm of course open to improvements for the model or approach to anything.

### DIFF
--- a/tomviz/HistogramWidget.cxx
+++ b/tomviz/HistogramWidget.cxx
@@ -8,6 +8,7 @@
 #include "DoubleSliderWidget.h"
 #include "ModuleContour.h"
 #include "ModuleManager.h"
+#include "PresetDialog.h"
 #include "QVTKGLWidget.h"
 #include "Utilities.h"
 
@@ -516,7 +517,11 @@ void HistogramWidget::onSaveToPresetClicked()
 
 void HistogramWidget::onPresetClicked()
 {
-  showPresetDialog(this, nullptr);
+  if (m_presetDialog == nullptr) {
+    m_presetDialog = new PresetDialog(this);
+  }
+  m_presetDialog->show();
+  //showPresetDialog(this, nullptr);
 }
 
 void HistogramWidget::applyCurrentPreset()
--- a/tomviz/HistogramWidget.cxx
+++ b/tomviz/HistogramWidget.cxx
@@ -525,11 +525,8 @@ void HistogramWidget::applyCurrentPreset()
     return;
   }
 
-  vtkSMProxy* sof =                                                                                
-      vtkSMPropertyHelper(lut, "ScalarOpacityFunction", true).GetAsProxy();
-  
   QString result = m_presetDialog->getName();
-  vtkSMTransferFunctionProxy::ApplyPreset(lut, result.toStdString().c_str(), true);
+  vtkSMTransferFunctionProxy::ApplyPreset(lut, result.toLatin1().data(), true);
 
   renderViews();                                                                                   
   emit colorMapUpdated();
--- a/tomviz/HistogramWidget.cxx
+++ b/tomviz/HistogramWidget.cxx
@@ -520,6 +520,8 @@ void HistogramWidget::onPresetClicked()
   if (m_presetDialog == nullptr) {
     m_presetDialog = new PresetDialog(this);
   }
+  QObject::connect(m_presetDialog, SIGNAL(applyPreset(const Json::Value&)),
+      this, SLOT(applyCurrentPreset()));
   m_presetDialog->show();
   //showPresetDialog(this, nullptr);
 }
--- a/tomviz/HistogramWidget.cxx
+++ b/tomviz/HistogramWidget.cxx
@@ -445,11 +445,13 @@ void HistogramWidget::onCustomRangeClicked()
 
   // Make sure the bottom isn't higher than the top, and the
   // top isn't higher than the bottom.
-  connect(&bottom, static_cast<void (QDoubleSpinBox::*)(double)>(
-                     &QDoubleSpinBox::valueChanged),
+  connect(&bottom,
+          static_cast<void (QDoubleSpinBox::*)(double)>(
+            &QDoubleSpinBox::valueChanged),
           &top, &QDoubleSpinBox::setMinimum);
-  connect(&top, static_cast<void (QDoubleSpinBox::*)(double)>(
-                  &QDoubleSpinBox::valueChanged),
+  connect(&top,
+          static_cast<void (QDoubleSpinBox::*)(double)>(
+            &QDoubleSpinBox::valueChanged),
           &bottom, &QDoubleSpinBox::setMaximum);
 
   QDialogButtonBox buttonBox;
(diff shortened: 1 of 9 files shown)


cryos:
Our initial set of colors is https://github.com/OpenChemistry/tomviz/blob/master/tomviz/resources/matplotlib_cmaps.json where we want to get them in the model, and generate previews. I think @psavery may be able to help a little here as he has helped pick through some of these API calls in the past for me.

tomviz/PresetDialog.cxx
```diff
@@ -0,0 +1,39 @@
+/* This source file is part of the Tomviz project, https://tomviz.org/. 
+   It is released under the 3-Clause BSD License, see "LICENSE". */
+
+#include "PresetDialog.h"
+#include "PresetModel.h"
+#include "ui_PresetDialog.h"
+
+#include <QTableView>
+#include <QtWidgets>
```
cryos: We avoid include entire modules, like QtWidgets. This can really slow down compilation. Prefer to forward declare what we can in the header, and then include the headers for the classes used.

tomviz/HistogramWidget.h
```diff
@@ -76,6 +77,8 @@ public slots:
   vtkWeakPointer<vtkTable> m_inputData;
 
   QVTKGLWidget* m_qvtk;
+
+  QWidget* m_presetDialog = nullptr;
```
psavery: This might should be a `PresetDialog*` instead of a `QWidget*`, so that you'd be able to call derived class functions if any get added.

tomviz/PresetDialog.cxx
```diff
@@ -0,0 +1,39 @@
...
+  // make columns fit what is in them
+  view->resizeColumnsToContents();
+  // make columns stetch to fit table size
+  view->horizontalHeader()->setSectionResizeMode(QHeaderView::Stretch);
+
+  // create a grid to center the column
+  QVBoxLayout *layout = new QVBoxLayout;
+  layout->addWidget(view);
+  layout->addWidget(m_ui->buttonBox);
+  this->setLayout(layout);
```
psavery: While widgets like the QTableView and QVBoxLayout *can* be added in programmatically, we typically would add them in via the Qt Designer GUI. I can help show how if needed.

cryos: I would rather keep it programmatic for now, Patrick is right that we can use UI files too.

tomviz/PresetModel.h
```diff
@@ -0,0 +1,38 @@
...
+  QVariant headerData(int section, Qt::Orientation orientation, int role) const override;
+
+ signals:
+  void getPreset(const Json::Value& preset);
+  
+ public slots:
+  void handleClick(const QModelIndex &index);
+
+ private:
+  vtkNew<vtkSMTransferFunctionPresets> Presets;
```
psavery: We usually name member variables with the `m_` prefix to better distinguish them from local variables.

tomviz/HistogramWidget.cxx
```diff
@@ -516,7 +517,13 @@ void HistogramWidget::onSaveToPresetClicked()
 
 void HistogramWidget::onPresetClicked()
 {
-  showPresetDialog(this, nullptr);
+  if (m_presetDialog == nullptr) {
+    m_presetDialog = new PresetDialog(this);
+  }
+  QObject::connect(m_presetDialog, SIGNAL(applyPreset(const Json::Value&)),
+      this, SLOT(applyCurrentPreset()));
```
psavery: This would probably be better placed within the `if()` block above. Otherwise, an extra connection will be made every time `onPresetClicked()` is called, and `applyCurrentPreset()` will be called the number of times it was connected (so if `onPresetClicked()` gets called 3 times, for instance, the third time it is called, `applyCurrentPreset()` will be called 3 times for a single `applyPreset` signal).

Also, this is using the old signal and slot syntax in Qt. You don't necessarily need to change it to the new syntax, but I think it can be useful for a few reasons, including that the new syntax will give a compile error when improperly connected, whereas the old syntax gives a runtime error. See [here](https://wiki.qt.io/New_Signal_Slot_Syntax) for more information. I think this would be changed to this if you went with the new syntax:
```
QObject::connect(m_presetDialog, &PresetDialog::applyPreset, this, &HistogramWidget::applyCurrentPreset);`
```

tomviz/PresetModel.h
bnmajor: Yes, thank you! I I messed that up in more than one place, they are all corrected now.

tomviz/PresetModel.cxx
```diff
@@ -0,0 +1,78 @@
...
+
+namespace tomviz {
+
+// ctor
+PresetModel::PresetModel(QObject *parent)
+  : QAbstractTableModel(parent)
+{
+  //  vtkNew<vtkSMTransferFunctionPresets> Presets;
+  auto begin = Presets->GetNumberOfPresets();
+  QString file = "../tomviz/tomviz/resources/matplotlib_cmaps.json";
```
psavery: You might be able to either copy-paste or re-use [this function](https://github.com/OpenChemistry/tomviz/blob/6d07b4cb05b9a18e794fb27d0dc75f5e80bc7e7f/tomviz/Behaviors.cxx#L113) to make the json file more easily obtainable.

tomviz/HistogramWidget.cxx
cryos: I think we should be happy with it either way in Tomviz, and not devote too much time/effort to enforcing one or the other style. Happy to prefer the new signatures, but I wouldn't worry about changing it either. It is a little nicer for failing more predictably using the compiler to create connections.

tomviz/PresetDialog.h
```diff
@@ -0,0 +1,36 @@
...
+
+namespace tomviz {
+
+class PresetDialog : public QDialog
+{
+  Q_OBJECT
+
+public:
+  explicit PresetDialog(QWidget* parent);
+  QString getName();
```
cryos: We use the Qt naming convention - setter is called `setName(const QString& name)` and getter is called `name()`. Also prefer declaring the constructor, then the destructor, and then the other members. This may benefit from a little more detail, such as `presetName()` rather than just name.

tomviz/PresetDialog.h
```diff
@@ -0,0 +1,36 @@
+/* This source file is part of the Tomviz project, https://tomviz.org/.
+  It is released under the 3-Clause BSD License, see "LICENSE". */
+
+#ifndef tomvizPresetDialog_h
+#define tomvizPresetDialog_h
+
+#include "PresetModel.h"
```
cryos: This should be forward declared as you are just using a pointer in the header.

tomviz/PresetDialog.cxx
```diff
@@ -0,0 +1,45 @@
...
+          &PresetDialog::applyPreset);
+  connect(m_model, &PresetModel::applyPreset, this, &PresetDialog::applyPreset);
+}
+
+QString PresetDialog::getName()
+{
+  return m_model->getName();
+}
+
+PresetDialog::~PresetDialog() = default;
```
cryos: Move this up to just below the constructor.

tomviz/PresetModel.cxx
cryos: As Patrick pointed out, this needs to be relative to the application directory, and doesn't work for mo locally. It would be good to adapt the code he pointed out, this wouldn't work in an installed application either.

tomviz/PresetModel.cxx
```diff
@@ -0,0 +1,84 @@
...
+
+#include <pqPresetToPixmap.h>
+
+#include <QPair>
+#include <QPixmap>
+#include <QSize>
+
+namespace tomviz {
+
+// ctor
```
cryos: Remove this comment, doesn't add anything.

cryos:
It works on my machine now too, and it looks pretty good. We should change the name of the dialog to color presets perhaps. Maybe we skip the use of the header in the end too as it would be in the dialog title, and could hide the header row altogether. I think this is about ready to merge and test out on the build bots.

PR #1872, opened 2019-07-01
Move slider widget precision to source file

psavery:
I upgraded my local ParaView to master and re-compiled tomviz.

As of [this commit](https://gitlab.kitware.com/paraview/paraview/commit/e98490b3545cf0c652e065494362095f711a1c5b) in ParaView, the slider widget precision was removed from "vtkPVConfig.h.in" and put in the slider widget source file.

We need to do the same, or we get a compilation error that says that `DEFAULT_DOUBLE_PRECISION_VALUE` is not defined.

### DIFF
--- a/tomviz/DoubleSliderWidget.cxx
+++ b/tomviz/DoubleSliderWidget.cxx
@@ -11,6 +11,8 @@
 #include <QHBoxLayout>
 #include <QSlider>
 
+#define DEFAULT_DOUBLE_PRECISION_VALUE 16
+
 namespace tomviz {
 
 DoubleSliderWidget::DoubleSliderWidget(bool showLineEdit, QWidget* p)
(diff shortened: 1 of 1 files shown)


tomviz/DoubleSliderWidget.cxx
```diff
@@ -11,6 +11,8 @@
 #include <QHBoxLayout>
 #include <QSlider>
 
+#define DEFAULT_DOUBLE_PRECISION_VALUE 16
```
cryos: I think we should remove the vtkPVConfig.h include as I think that is only there for this variable. Also, there is only one call site, let's just put the number in there - it feels over-engineered to use a define for a single call site. I wonder if 8 might also be enough, but happy to go either way there.

cryos:
It is odd that I didn't see a compile failure as that is a pretty old commit. Maybe I have an old header hanging around.

psavery:
Looks like that even though the commit was made a while back, the [MR](https://gitlab.kitware.com/paraview/paraview/merge_requests/2989) was just merged a couple of weeks ago. Maybe that's why.

tomviz/DoubleSliderWidget.cxx
psavery: Done and done. I also changed the precision to 8.

PR #1875, opened 2019-07-02
Update python paths for python 3.7

psavery:
Only Linux has been done so far, but others will be added if needed.


psavery:
Yeah, I didn't see any changes needed for Mac as of yet. Windows might not need any changes either. If they are both good, then this will be ready.

PR #1876, opened 2019-07-02
Create new presets and save them

bnmajor:
Started getting paraview errors about possibly invalid preset right before commiting, still seems to work. Does not check for duplicate names. Looking for input on improved design and best practices.

### DIFF
--- a/tomviz/PresetModel.cxx
+++ b/tomviz/PresetModel.cxx
@@ -2,33 +2,38 @@
    It is released under the 3-Clause BSD License, see "LICENSE". */
 
 #include "PresetModel.h"
+#include "Utilities.h"
 
 #include <pqPresetToPixmap.h>
+#include <pqSettings.h>
 
+#include <QJsonArray>
+#include <QJsonObject>
 #include <QPair>
 #include <QPixmap>
 #include <QSize>
+
+#include <vtkSMProxy.h>
+#include <vtkScalarsToColors.h>
+#include <vtk_jsoncpp.h>
 
 namespace tomviz {
 
 PresetModel::PresetModel(QObject* parent) : QAbstractTableModel(parent)
 {
-  vtkNew<vtkSMTransferFunctionPresets> Presets;
-  auto begin = Presets->GetNumberOfPresets();
-  QString file = getMatplotlibColorMapFile();
-  Presets->ImportPresets(file.toStdString().c_str());
-
-  pqPresetToPixmap PixMapRenderer;
-  for (auto i = begin; i < Presets->GetNumberOfPresets(); ++i) {
-    m_Pixmaps.push_back(QPair<QString, QPixmap>(
-      static_cast<QString>(Presets->GetPresetName(i)),
-      PixMapRenderer.render(Presets->GetPreset(i), QSize(135, 20))));
+  auto settings = pqApplicationCore::instance()->settings();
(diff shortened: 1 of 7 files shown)


cryos:
I know this is draft, so maybe you are going to fix things anyway. Generally no two commits would have the same subject line, it looks like they are duplicate commits. If they should have the same subject you should just squash them together. If they are distinct you should add a distinct subject line. Generally avoid merging into topic branches too, they should be simple linear topics with a branch point from master, commits, and then a merge back into master. Sometimes it is easier to merge upstream master right before merging to resolve conflicts, but otherwise just merging changes is good. Just going to build and take a look at running code now.

cryos:
On my machine Viridis has an odd color issue, with no color point in our GUI, but it shows up in the slice too:
image

tomviz/PresetModel.cxx
```diff
@@ -40,59 +45,109 @@ QVariant PresetModel::data(const QModelIndex& index, int role) const
...
 }
 
-QString PresetModel::getMatplotlibColorMapFile()
+void PresetModel::addNewPreset(const QJsonObject& newPreset)
+{
+  m_Presets.push_back(newPreset);
+  render(newPreset);
+  updateRow(m_Presets.size()-1);
+  saveSettings();
+  beginResetModel();
```
cryos: This is the sledge hammer we may want to avoid in the future.

tomviz/PresetModel.cxx
```diff
@@ -40,59 +45,109 @@ QVariant PresetModel::data(const QModelIndex& index, int role) const
...
+}
+
+QPixmap PresetModel::render(const QJsonObject& newPreset) const
+{
+  QJsonDocument doc(newPreset);
+  QString preset(doc.toJson(QJsonDocument::Compact));
+
+  Json::Value colors;
+  Json::Reader reader;
+  reader.parse(preset.toStdString().c_str(), colors);
```
cryos: Prefer `toLatin1().data()` over `toStdString().c_str()`

tomviz/PresetModel.cxx
```diff
@@ -45,7 +45,7 @@ QVariant PresetModel::data(const QModelIndex& index, int role) const
 {
   switch (role) {
     case Qt::DisplayRole:
-      return m_Presets[index.row()].toObject().value("name");
+      return m_Presets[index.row()].toObject().value("Name");
```
cryos: Please stick to lower case first letter for first word

tomviz/PresetModel.cxx
```diff
@@ -144,8 +144,11 @@ void PresetModel::loadFromFile()
   QJsonArray objects = doc.array();
   for (auto value : objects) {
     QJsonObject obj = value.toObject();
-    QJsonObject nextDefault{ { "name", obj["Name"] },
-                             { "RGBPoints", obj["RGBPoints"] } };
+    QJsonObject nextDefault{
+      { "Name", obj["Name"] },
+      { "ColorSpace", obj["ColorSpace"] },
```
cryos: Same here. It would be nice to keep the state file and color JSON object using the same case/names so that we only have one convention to worry about in Tomviz.

cryos:
This looks close, we can go ahead and merge but one major issue (follow up branch is OK here) is that the default color map on loading data is now one that isn't even shown in the dialog. I think we probably just want to adjust how we create color maps on load.

PR #1891, opened 2019-07-11
Preset defaults

bnmajor:
Users can now edit preset names and reset preset options back to defaults. Currently double clicking to apply and preview a preset also enables editing the name and I'm not sure if the pixmap and name should be two columns or editing should be a context menu option, or if it's not much of an issue.


cjh1:
@cryos @bnmajor Testing this branch and it works well for me.

PR #1897, opened 2019-07-16
Increase allowed range for data transformation

bnmajor:
Add constant data transformation is now only limited by the largest and smallest possible values given its type.


cryos:
@bnmajor please reference the issues if your pull request is addressing one. We may want to think a little harder about this, the early version used numeric limits but it causes input boxes to be too large on default layouts.

PR #1907, opened 2019-07-19
Ensure physical size is correctly set for child datasets.

cjh1:

### DIFF
--- a/tomviz/python/tomviz/utils.py
+++ b/tomviz/python/tomviz/utils.py
@@ -165,6 +165,10 @@ def make_child_dataset(reference_dataset):
     from vtk import vtkImageData
     new_child = vtkImageData()
     new_child.CopyStructure(reference_dataset)
+    input_spacing = reference_dataset.GetSpacing()
+    # For a reconstruction we copy the X spacing from the input dataset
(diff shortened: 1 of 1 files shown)


tomviz/python/tomviz/utils.py
```diff
@@ -329,6 +333,28 @@ def mark_as_tiltseries(dataobject):
...
+        lengths.append(y)
+    if z is not None:
+        axes.append(2)
+        lengths.append(z)
+
+    extent = dataobject.GetExtent()
+    spacing = list(dataobject.GetSpacing())
+    for axis, new_length in zip(axes, lengths):
+        spacing[axis] = \
+            new_length / (extent[2 * axis + 1] - extent[2 * axis] + 1)
```
cryos: Here you are calculating the spacing from the length, why not just have the API set the voxel size? I think that is what most internal code does. If we want to set the voxel size I think this is off by one - if I say the length is 2nm, and the cube has three points it only has two voxels (from 0 to 1 and from 1 to 2 or am I missing something?)

cryos: Sorry, we do treat these as pixels, and so they effectively have an extra half on each side when we draw the slice/volume. That means that the + 1 you used / copied is correct in the way we are defining voxels having a value.

PR #1915, opened 2019-08-02
Allow user to choose stride for big HDF5 data

psavery:
Before, the stride would automatically be 2. Now, the user can
choose the stride if a big dataset is detected.

Below is an example image of the dialog.

image

### DIFF
--- a/tomviz/GenericHDF5Format.cxx
+++ b/tomviz/GenericHDF5Format.cxx
@@ -11,6 +11,7 @@
 #include <QComboBox>
 #include <QDialog>
 #include <QDialogButtonBox>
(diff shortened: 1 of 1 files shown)


psavery:
For re-running the data using a different stride/volume subsample, if we make a dialog for it, we might replace this input dialog with it.

cryos:
This is a good start, we need to generalize a little for Data Exchange and other HDF5-based formats.

psavery:
Yeah, Data Exchange and the EMD formats both call this modified function (`GenericHDF5Format::readVolume()`). So I think it should work for them as well.

PR #1916, opened 2019-08-02
For HDF5 formats, allow reload and resample

psavery:
For HDF5 formats, there is now a new context menu item for the
original data source in the pipeline view called "Reload and Resample".
    
If selected, it will prompt the user to choose a stride, and it
will then reload the dataset using the stride. It reloads the dataset
by writing over the previous vtkImageData. No new DataSource gets
created.
    
It seems to work for the most part, except for a few quirks:
    
~~1. The volume module disappears when it is reloaded and resampled.~~
2. The slice widget size does not update with the new volume size.
3. The camera does not seem to update with the new volume size. The "Reset Camera" button doesn't zoom correctly.
    
I am looking through these. Let me know if there is a more preferable
way to do anything I've done so far.

### DIFF
--- a/tomviz/DataExchangeFormat.cxx
+++ b/tomviz/DataExchangeFormat.cxx
@@ -34,7 +34,8 @@ void ReorderArrayC(T* in, T* out, int dim[3])
   }
 }
 
-bool DataExchangeFormat::read(const std::string& fileName, vtkImageData* image)
+bool DataExchangeFormat::read(const std::string& fileName, vtkImageData* image,
+                              bool checkSize, int stride)
 {
   using h5::H5ReadWrite;
   H5ReadWrite::OpenMode mode = H5ReadWrite::OpenMode::ReadOnly;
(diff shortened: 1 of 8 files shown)


psavery:
Nevermind about number 1. There was no volume module to begin with...

psavery:
It looks like the "Reset Camera" button works properly if there is no slice module. The slide module must not be updating.

psavery:
The last two issues above appear more to be issues with the slice widget and "Reset Camera" in conjunction with the slice widget when data gets resized. I posted issue #1917 to track it. If I delete the slice module and start with a volume module instead, resizing the data does not appear to cause "Reset Camera" issues.

psavery:
There is one more issue I noticed that I will mention: if a dataset is loaded, and an operator (such as "Square Root Data") was performed on the dataset, and then you "Reload and Resample" the original dataset, it does not seem to update everything. It looks like the colormap for the original dataset changes, and the "Units and Size" change, but the rest of the "Data Properties Panel" do not update, and the rendered view does not change either.

Maybe I need to make sure it re-runs the pipeline when the original dataset gets resized. Not sure...

cryos:
We will need to trigger re-running the pipeline when the original data set is resampled.

psavery:
For now, I just disabled the ability to resample the original dataset if the original dataset has children. It's something I think can we can fix in a future PR.

PR #1921, opened 2019-08-18
Cast memsize int to size_t to prevent overflow

psavery:
For big numbers, there was an integer overflow when it was
multiplied by 1000. This resulted in a really big number when
it was converted to size_t. Fix the overflow.

Before (data type is uint16_t):

800_800_2048_memory_glitch

After:

Screenshot from 2019-08-18 15-52-02


cryos:
I think you mean overflowed, you are worried about the number wrapping round.

psavery:
@cryos Yes, that's right. I mistyped. Meant to say "overflow" instead...

psavery:
Looks like I mistyped it in many places...

psavery:
I fixed it everywhere except for the branch name. I guess I would need to make a new PR to fix the branch name. Should I do that?

PR #1922, opened 2019-08-18
Add ability for user to select HDF5 subvolume

psavery:
This adds a simple dialog that allows a user to select a subvolume and
a stride. The dialog also shows the user an estimated memory usage
(which is quite accurate) at the bottom.

When reading a volume from an HDF5 file for the first time, the dialog
will automatically appear if one of the dimensions is greater than
1200, so that the user can subsample if needed.

In addition, the dialog will appear if the user right-clicks the
dataset in the pipeline view and selects "Reload and Resample".

Note that the previous stride and subvolume choices are not saved,
and if the user chooses to "Reload and Resample", it will always
start with the default settings.

A sample can be seen below.

image

### DIFF
--- a/tomviz/CMakeLists.txt
+++ b/tomviz/CMakeLists.txt
@@ -73,6 +73,8 @@ set(SOURCES
   GenericHDF5Format.h
   GradientOpacityWidget.h
   GradientOpacityWidget.cxx
+  Hdf5SubsampleWidget.h
+  Hdf5SubsampleWidget.cxx
   HistogramManager.h
   HistogramManager.cxx
   HistogramWidget.h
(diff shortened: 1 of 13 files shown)


psavery:
It now remembers the previous settings that were used, and it displays them if the user tries to subsample again.

cjh1:
@psavery Will this also work for EMD?

psavery:
@cjh1 Yep! It will. It will work for `DataExchange`, `EMD`, and the generic HDF5 format too.

cjh1:
@psavery Awesome! That is really good.

cjh1:
@psavery Just a few clang-format issues, then we can merge.

PR #1928, opened 2019-08-22
Add support for running external pipeline on multiple datasets

cjh1:

### DIFF
--- a/tomviz/python/tomviz/cli/__init__.py
+++ b/tomviz/python/tomviz/cli/__init__.py
@@ -1,8 +1,12 @@
 import click
 import json
 import os
+import logging
+from pathlib import Path
 
 from tomviz import executor
+
+logger = logging.getLogger('tomviz')
 
 
 def _extract_pipeline(state):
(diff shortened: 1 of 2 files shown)


tomviz/python/tomviz/cli/__init__.py
```diff
@@ -59,23 +65,46 @@ def main(data_file_path, state_file_path, output_file_path, progress_method,
...
-        data_file_path = filenames[0]
+        data_path = filenames[0]
         # fileName is relative to the state file location, so convert to
         # absolute path.
-        data_file_path = os.path.abspath(
-            os.path.join(os.path.dirname(state_file_path), data_file_path))
-        if not data_file_path.lower().endswith('.emd'):
+        data_path = os.path.abspath(
+            os.path.join(os.path.dirname(state_file_path), data_path))
+        if not data_path.lower().endswith('.emd'):
```
psavery: So I noticed that when it is reading from the state file, it requires the file to be EMD. But below ([this block](https://github.com/OpenChemistry/tomviz/pull/1928/files#diff-f4a51f1c4f8109ed11a0e42e6aac1c32R96)), there doesn't seem to be an EMD requirement. Is that intentional?

cjh1: Well, in the case of a directory being provided we glob for *.emd, in the other case we would try to open the file as an hdf5 file which, will probably raise an exception. i can add in another check for the file ending if you think this is useful.

psavery: No, I was just wondering. It's up to you. Looks good otherwise.

cjh1: I moved the check further down, so now it should catch both cases.

PR #1939, opened 2019-08-30
Fix up operator test

cjh1:
Signed-off-by: Chris Harris <chris.harris@kitware.com>

### DIFF
--- a/tests/cxx/OperatorPythonTest.cxx
+++ b/tests/cxx/OperatorPythonTest.cxx
@@ -27,7 +27,7 @@ protected:
   void SetUp() override
   {
     dataObject = vtkDataObject::New();
-    pythonOperator = new OperatorPython();
+    pythonOperator = new OperatorPython(nullptr);
   }
 
   void TearDown() override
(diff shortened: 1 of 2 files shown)


tests/cxx/OperatorPythonTest.cxx
```diff
@@ -27,7 +27,7 @@ class OperatorPythonTest : public ::testing::Test
   void SetUp() override
   {
     dataObject = vtkDataObject::New();
-    pythonOperator = new OperatorPython();
+    pythonOperator = new OperatorPython(nullptr);
```
psavery: I think this might crash when ran because `OperatorPython` tries to access its parent `DataSource` [in its constructor](https://github.com/OpenChemistry/tomviz/blob/e3e8e01fe33ee51ad039f8a906ce11f1d9bbcd49/tomviz/operators/OperatorPython.cxx#L160). :slightly_frowning_face:

cjh1: I think your right

psavery: I guess we can either make a `DataSource` here, or maybe modify the python operator constructor to check if its parent exists before calling functions on it?

cjh1: yep, I am looking a creating a `DataSource` but it looking like a rabbit hole ...

cjh1:
@psavery This is fixed up now. After going down the rabbit hole of trying to instantiate a `DataSource` which started pull in large  chunks of ParaView. I opted to just add a check in `OperatorPython`

PR #1941, opened 2019-09-04
Rotate align widget orientation tweaks

psavery:
This has two changes in it:

1. The vertical orientation in the GUI was renamed to "X-ray" instead of "XDT".
2. The orientation is saved between runs so that if a user primarily uses one orientation, they do not need to keep switching.

### DIFF
--- a/tomviz/RotateAlignWidget.cxx
+++ b/tomviz/RotateAlignWidget.cxx
@@ -15,6 +15,9 @@
 
 #include <cmath>
 
+#include <pqApplicationCore.h>
+#include <pqSettings.h>
+
 #include <vtkSMTransferFunctionManager.h>
 #include <vtkSMTransferFunctionProxy.h>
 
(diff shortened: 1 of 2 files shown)


tomviz/RotateAlignWidget.cxx
```diff
@@ -757,6 +777,11 @@ void RotateAlignWidget::updateControls()
 
   this->Internals->axesActor->SetXAxisRange(xAxisRange);
   this->Internals->axesActor->SetYAxisRange(yAxisRange);
+
+  // It would be nice if we could only write the settings when the
+  // widget is accepted, but I don't immediately see an easy way
+  // to do that.
```
cjh1: Yep, would be nice `EditOperatorDialog` has a `editEnded` signal that we could probably connect up to the `EditOperatorWidget`, but I don't think its worth it.

PR #1947, opened 2019-09-06
Add tomopy to tomviz

psavery:
This is a work in progress that will add tomopy into a docker image of
the tomviz pipeline, and it will be runnable there. Currently, only the
Dockerfile has been made.

This Dockerfile is almost identical to the tomviz-pipeline Dockerfile,
except that the base is miniconda3, and it uses conda to install
tomopy.

Since this one uses miniconda3 and tomopy, it is ~3.3x larger in size
than the regular tomviz-pipeline image. `docker image ls` shows it at
4.54GB in size.

### DIFF
--- a/docker/tomviz-tomopy-pipeline/BUILDING.md
+++ b/docker/tomviz-tomopy-pipeline/BUILDING.md
@@ -1,0 +1,10 @@
+Building tomviz-tomopy-pipeline
+===============================
+
+In order to build this container the tomviz repository needs to be in the docker
+build context. Therefore the build must be performed from the root of the
+repository.
+
+    cd <tomviz-repo-root>
+    docker build -f docker/tomviz-tomopy-pipeline/Dockerfile .
+
--- a/docker/tomviz-tomopy-pipeline/BUILDING.md
+++ b/docker/tomviz-tomopy-pipeline/BUILDING.md
@@ -6,5 +6,4 @@ build context. Therefore the build must be performed from the root of the
 repository.
 
     cd <tomviz-repo-root>
-    docker build -f docker/tomviz-tomopy-pipeline/Dockerfile .
-
+    docker build -f docker/tomviz-tomopy-pipeline/Dockerfile -t tomviz/tomopy-pipeline .
--- a/tomviz/DataExchangeFormat.cxx
+++ b/tomviz/DataExchangeFormat.cxx
@@ -22,51 +22,41 @@
 
 namespace tomviz {
 
-bool DataExchangeFormat::read(const std::string& fileName, vtkImageData* image,
-                              const QVariantMap& options)
+static bool readDataSet(const std::string& fileName, const std::string& path,
+                        vtkImageData* image, const QVariantMap& options)
 {
   using h5::H5ReadWrite;
   H5ReadWrite::OpenMode mode = H5ReadWrite::OpenMode::ReadOnly;
   H5ReadWrite reader(fileName.c_str(), mode);
 
-  std::string deDataNode = "/exchange/data";
   // If it isn't a data set, we are done
-  if (!reader.isDataSet(deDataNode))
+  if (!reader.isDataSet(path))
     return false;
 
-  return GenericHDF5Format::readVolume(reader, deDataNode, image, options);
+  return GenericHDF5Format::readVolume(reader, path, image, options);
+}
+
+bool DataExchangeFormat::read(const std::string& fileName, vtkImageData* image,
+                              const QVariantMap& options)
+{
+  std::string path = "/exchange/data";
+  return readDataSet(fileName, path, image, options);
 }
 
 bool DataExchangeFormat::readDark(const std::string& fileName,
                                   vtkImageData* image,
                                   const QVariantMap& options)
 {
-  using h5::H5ReadWrite;
-  H5ReadWrite::OpenMode mode = H5ReadWrite::OpenMode::ReadOnly;
-  H5ReadWrite reader(fileName.c_str(), mode);
-
-  std::string node = "/exchange/data_dark";
-  // If it isn't a data set, we are done
-  if (!reader.isDataSet(node))
-    return false;
-
-  return GenericHDF5Format::readVolume(reader, node, image, options);
+  std::string path = "/exchange/data_dark";
+  return readDataSet(fileName, path, image, options);
 }
 
 bool DataExchangeFormat::readWhite(const std::string& fileName,
                                    vtkImageData* image,
                                    const QVariantMap& options)
 {
-  using h5::H5ReadWrite;
-  H5ReadWrite::OpenMode mode = H5ReadWrite::OpenMode::ReadOnly;
-  H5ReadWrite reader(fileName.c_str(), mode);
-
-  std::string node = "/exchange/data_white";
-  // If it isn't a data set, we are done
-  if (!reader.isDataSet(node))
-    return false;
-
-  return GenericHDF5Format::readVolume(reader, node, image, options);
+  std::string path = "/exchange/data_white";
+  return readDataSet(fileName, path, image, options);
 }
 
 static bool writeData(h5::H5ReadWrite& writer, vtkImageData* image)
--- a/tomviz/DataExchangeFormat.cxx
+++ b/tomviz/DataExchangeFormat.cxx
@@ -61,9 +61,7 @@ bool DataExchangeFormat::readWhite(const std::string& fileName,
 
 static bool writeData(h5::H5ReadWrite& writer, vtkImageData* image)
 {
-  // Create an "/exchange" group
-  writer.createGroup("/exchange");
-
+  // Assume /exchange already exists
   return GenericHDF5Format::writeVolume(writer, "/exchange", "data", image);
 }
 
@@ -86,6 +84,10 @@ bool DataExchangeFormat::write(const std::string& fileName, DataSource* source)
   using h5::H5ReadWrite;
   H5ReadWrite::OpenMode mode = H5ReadWrite::OpenMode::WriteOnly;
   H5ReadWrite writer(fileName, mode);
+
+  // Create a "/exchange" group
+  writer.createGroup("/exchange");
+
   auto t = source->producer();
   auto image = vtkImageData::SafeDownCast(t->GetOutputDataObject(0));
   if (!writeData(writer, image))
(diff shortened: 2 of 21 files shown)


ercius:
@psavery How can this be tested when it's closer to stable? Im not sure what is required to connect tomviz to a docker instance.

psavery:
@ercius I think the main requirement will be that docker needs to be running on the computer. It should work on all of the major operating systems, but I personally haven't tried it on Windows before. When we get an example running, I'll try it out on Windows and write down some instructions for how to do it.

cjh1:
Yes, this should work on Windows, I have tried it in the past without a problem, let us know if you run into any issue. The install of docker on Windows if pretty straight forward.

cryos:
Additionally we are working on generalizing what was done for Docker so that we can point to your Python interpreter. That would need you to set the packages up, but would offer another alternative.

tomviz/DataExchangeFormat.h
```diff
@@ -19,7 +19,16 @@ class DataExchangeFormat
...
+  // Read the dark dataset into the image data
+  bool readDark(const std::string& fileName, vtkImageData* data,
+                const QVariantMap& options = QVariantMap());
+  // Read the white dataset into the image data
+  bool readWhite(const std::string& fileName, vtkImageData* data,
+                 const QVariantMap& options = QVariantMap());
   bool write(const std::string& fileName, DataSource* source);
+
+  // This will not write out dark and white datasets
+  // TODO: figure out how to write dark and white datasets out here
```
psavery: This function appears to be used in `ExportDataReaction`. Unfortunately, if the user saves the data by using this function, the dark/white datasets will not be saved as well, because the `DataSource` object is required for them to be saved.

tomviz/DataExchangeFormat.cxx
```diff
@@ -37,23 +37,86 @@ bool DataExchangeFormat::read(const std::string& fileName, vtkImageData* image,
...
+  using h5::H5ReadWrite;
+  H5ReadWrite::OpenMode mode = H5ReadWrite::OpenMode::ReadOnly;
+  H5ReadWrite reader(fileName.c_str(), mode);
+
+  std::string node = "/exchange/data_white";
+  // If it isn't a data set, we are done
+  if (!reader.isDataSet(node))
+    return false;
+
+  return GenericHDF5Format::readVolume(reader, node, image, options);
```
cjh1: These two methods are pretty much the same, can we refactor them into a single method that take the path to read the volume from, something like this ...

```c++
bool DataExchangeFormat::readData(const std::string& fileName, 
                              const std::string& path, 
                              vtkImageData* image, 
                              const QVariantMap& options
{
// Read the volume at this path
}

bool DataExchangeFormat::readWhite(const std::string& fileName,
                                   vtkImageData* image,
                                   const QVariantMap& options)
{
  return this->readData(fileName, "/exchange/data_white", image, options);
}

```

tomviz/DataExchangeFormat.h
cjh1: I think we can change the signature of the `write(...)` method and then call `dataSource()` [here] (https://github.com/OpenChemistry/tomviz/blob/master/tomviz/ExportDataReaction.cxx#L182) instead of `dataToExport()`, which will give us access to the `DataSource`

tomviz/DataExchangeFormat.cxx
psavery: Done.

tomviz/DataExchangeFormat.h
psavery: Done.

tomviz/PipelineExecutor.cxx
```diff
@@ -176,6 +177,19 @@ DockerPipelineExecutor::DockerPipelineExecutor(Pipeline* pipeline)
 
 DockerPipelineExecutor::~DockerPipelineExecutor() = default;
 
+QString DockerPipelineExecutor::originalFileName()
+{
+  QString ext = ".emd";
+  auto* dataSource = pipeline()->dataSource();
+  if (dataSource->darkData() && dataSource->whiteData())
```
psavery: @cjh1 Here is the function I'm using to determine whether docker writes out to emd or Data Exchange. Because it was simpler, I just checked for the presence of dark and white data to determine whether we should write to Data Exchange. However, if it's preferable, I can instead check the original file extension, and then open up the original file and make sure that "/exchange/data" exists. Let me know.

cjh1: No, that should work fine.

psavery:
I have added a [sample script](https://github.com/OpenChemistry/tomviz/blob/86b77254140c59a779e6037abe6d5d1852120ef6/tomviz/python/Recon_tomopy_gridrec.py) that is loaded when tomviz starts. It is under "Reconstruction:" as "Tomography"->"Tomopy Gridrec Method". The pipeline mode must be set to "Docker", and the tomviz-tomopy-pipeline docker image is required for this to work.

I have an example working that follows the [first tomopy example](https://tomopy.readthedocs.io/en/latest/ipynb/tomopy.html), and it uses [this tooth sinogram](https://github.com/OpenChemistry/tomviz/files/3597197/tooth.h5.gz).

First, the data exchange file is loaded, and the tooth sinogram can be visualized:

tooth_sinogram

Next, select "Tomography"->"Tomopy Gridrec Method", and use 290 as the rotation center, like so:

tomopy_settings

Accept it, and then the tooth reconstruction can be visualized:

tooth_recon


Probably the biggest thing that still needs to be done is that we need to read in and use the "/exchange/theta" dataset from the Data Exchange file. We currently assume the theta range to be from 0 to 360, which is true for many examples, but it would be preferable to use the "/exchange/theta" dataset. We'll need to figure out how we should store it in tomviz - that is the key thing that needs to be determined.

Some other examples need to be attempted as well.

psavery:
I was able to get our tomviz tomopy recon script to reproduce the results of the reconstruction of
[tomo_00001 from tomobank](https://tomobank.readthedocs.io/en/latest/source/data/docs.data.roundrobin.html). They have a [tomopy python script](https://github.com/tomography/tomobank/blob/master/docs/demo/tomopy_rec.py) that uses nearly the same steps as our tomopy recon script, with [one extra step of removing stripes](https://github.com/tomography/tomobank/blob/1c9392690cd646f47cd6607b2de1dc6d349c89e0/docs/demo/tomopy_rec.py#L90). If I add that line to our script, I get pretty much the same results.

However, I should mention that I had to do this with a subsampled version of tomo_00001. The tomobank recon script, when used on the full sized dataset, produces an output that is ~30 GB in size. The output is pictured below.

bnl_full_recon_slice

With the conversions between Fortran and C ordering, along with keeping the original dataset in RAM, I could not find a way to do the full reconstruction in tomviz without exceeding 64 GB of RAM.

Thus, I subsampled the dataset by striding all axes (except for the tilt axis) by three via a python script, and creating a new data exchange file from it. I then ran the tomobank recon script on it, and the following result was produced:

bnl_subsampled_recon_slice
bnl_subsampled_recon_full

When I ran our tomviz tomopy recon script, with the [additional line of
removing the stripes](https://github.com/tomography/tomobank/blob/1c9392690cd646f47cd6607b2de1dc6d349c89e0/docs/demo/tomopy_rec.py#L90) as they do in the tomobank tomopy python script, and a rotation center of exactly half the x axis (341.5 in our case) with no rotation center tuning, I get the following result:

tomviz_subsampled_recon_slice_stripe_nan_removal2
tomviz_subsampled_recon_slice_stripe_nan_removal_volume

Thus, I think we are able to reproduce their results.

psavery:
One thing that might be a concern is that there are many conversions to and from Fortran ordering in the process. Because the tomviz python scripts typically operate on the vtk data objects from tomviz in the threaded pipeline execution environment, many of the scripts are designed to operate on data with Fortran ordering and the tilt axis being the third axis. Because of this, our python scripts in the docker pipeline execution environment automatically convert the data to Fortran ordering and make the tilt axis the third axis as well.

However, tomopy requires the first axis to be the tilt axis, and although both C and Fortran ordering work, C ordering is faster.

There are currently 3 conversions from C to Fortran ordering, and 2 conversions from Fortran to C ordering in this process (and note that the dark and white data also have their ordering converted
each time the main data has its ordering converted):

```
Conversions from C to Fortran Ordering

1. Reading the Data Exchange file into tomviz.
2. Reading the Data Exchange file in the python pipeline in docker.
3. Reading the python pipeline output EMD file back into tomviz.

Conversions from Fortran to C Ordering

1. Writing the Data Exchange file to be read in the python pipeline in
   docker.
2. Writing the output of the python pipeline to EMD to be read by
   tomviz.
```

We could potentially skip conversion number 2 of "C to Fortran" and number 2 of "Fortran to C" if we kept the ordering as C ordering in the python pipeline for tomopy - plus, tomopy would be faster since
the ordering is in C.

Perhaps we can somehow tag operators as requiring C or Fortran ordering and, when reading the state file to be ran in the pipeline, check which orderings the operators need, and don't convert the data to Fortran ordering if no operators require Fortran ordering?

This might also work well with the `tomviz-pipeline` executable, for running a pipeline on many files.

psavery:
The last commit is a big change where we pass a wrapper into the python `transform_scalars()` functions instead of the vtkDataObject. This wrapper contains both the vtkDataObject and the DataSource object.

I have tried to make sure most operators still work, but I may have missed some.

docker/tomviz-tomopy-pipeline/BUILDING.md
```diff
@@ -0,0 +1,10 @@
+Building tomviz-tomopy-pipeline
+===============================
+
+In order to build this container the tomviz repository needs to be in the docker
+build context. Therefore the build must be performed from the root of the
+repository.
+
+    cd <tomviz-repo-root>
+    docker build -f docker/tomviz-tomopy-pipeline/Dockerfile .
```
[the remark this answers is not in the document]
psavery: Done.

cryos:
There is a conflict showing. They are relatively unusual as operators go, I wonder if we could emit a warning when we see that pattern. It is unfortunate that it wasn't standardized, but I think we should probably bite the bullet. As that isn't really our API even it would be hard to add a warning.

psavery:
Merge conflict is resolved, and the examples above were re-tested as well.

psavery:
Should I add a little documentation for this and a "Help" button that links to it?

cjh1:
Absolutely

psavery:
Documentation will be added later.

PR #1949, opened 2019-09-09
If generic HDF5 is a tilt series, skip transpose

psavery:
Ask the user if the generic HDF5 data is a tilt series. If so, skip
the transpose from C ordering to Fortran ordering.

This does not set the image type to be a tilt series, because the
tilt angles are unknown.


psavery:
@cryos Let me know if this does what is needed.

psavery:
This does not appear to be what was needed. Closing, and we can re-open in the future if needed.

Issue #1952, opened 2019-09-10
Child dataset bug in Python Reconstructions


jtschwar:
**Describe the bug**

Hi, whenever I run any of the python tomography reconstructions I get this error:

`No child dataset named  "" defined in dictionary returned from Python script.`

`Dictionary return from Python script is:
 "{'reconstruction': (vtkCommonDataModelPython.vtkImageData)0x1582208e8}"`

I've seen this error on both the SIRT and WBP projection, but I wouldn't be surprised if it will also occur on ART and TV as well. 

**Screenshots**

![image](https://user-images.githubusercontent.com/32110921/64627199-3439e580-d3bd-11e9-888e-f36aa99182ff.png)

**Desktop (please complete the following information):**
 - Operating system: macOS 10.14.6
 - Tomviz version: 1.70-131-g0b9

**Hardware (please complete the following information):**
 - Device: iMac
 - Graphics card: [e.g. NVIDIA Quadro 6000]

cjh1:
@jtschwar I can recreate this, will look into it.

PR #1953, opened 2019-09-11
1952 recon fix

cjh1:


cjh1:
Fixes #1952

psavery:
While it does look like the error no longer appears when the reconstruction is performed, there appear to be a couple of other issues.

I was testing this out using the tilt series EMD sample that we have. I was doing so by inverting the data first, and then performing the reconstruction. With this PR, though, if I invert the tilt series, the data does not appear to invert (see below), and it somehow loses its "tilt series" status (Mark as Tilt Series becomes enabled, for example).

no_invert

Then, if I mark it as a tilt series, and then perform the reconstruction, it does not appear to move the modules down either.

no_move_modules_down

Could you look into these?

PR #1954, opened 2019-09-11
Fix integer overflows for big data

psavery:
Some data sets can be very large (such as 1501x1792x2048). For big
data sets, an integer overflow would occur during the transpose from
C to Fortran or back, and it would cause the data to look very bad.

Cast the index to a size_t before the final multiplication to prevent
the integer overflow.


psavery:
I modified the name since it now fixes integer overflows in two separate places.

PR #1955, opened 2019-09-19
Fix EMD dimension swapping in the executor script

psavery:
Before, the way dimensions were swapped in the EMD reader was different
from the way dimensions were swapped in the EMD writer. This resulted
in "dim3" having the tilt angles instead of "dim1".

In the reader, the order of the dimensions were swapped. Whereas, in
the writer, the names of the dimensions were swapped.

This PR makes the swapping consistent so that the tilt angles are
written to the correct dimension.


psavery:
By the way, I tested both cases for the writer where `dims` is `None` and `dims` is not `None`. They seem to work with these changes.

PR #1958, opened 2019-09-25
New python transform()

psavery:
Work in progress for adding the new python `transform()` method. This method will use a different argument from `transform_scalars()`, but we plan to keep `transform_scalars()` for now for backward compatibility.

Currently, `transform()` is searched for in the script, and it is used if found. If not found, it searches for `transform_scalars()` to use. Currently, `transform()` is being passed the same argument as `transform_scalars()`, but that will change soon...

Fixes: #1974


psavery:
Just a note for myself: when we start using numpy arrays for the data, and when a threaded update to the progress data is called, we can probably convert the numpy array to VTK somewhere around [here](https://github.com/OpenChemistry/tomviz/blob/bb2715aee87128f30ededb131ed6e558ade79e8e/tomviz/python/tomviz/operators.py#L63). It looks like the docker update to the progress data already uses numpy arrays.

psavery:
[Here](https://github.com/OpenChemistry/tomviz/blob/1247a4af98351ed0dd34e5c3a9dd01803e94e72b/tomviz/python/tomviz/threaded_data_object.py#L5) is the current input class for the new `transform()` method for threaded pipeline mode. `active_scalars` is a property for the active scalars. `scalars` is a property for a dict of all the scalars. `tilt_angles` gives the tilt angles, and `dark` and `white` give the dark and white data, respectively.

I am going to wait for feedback on the API before proceeding to change all the operators. Pasted below is an example of the new style in action for a shift:
```
def transform(dataset, shift=[0, 0, 0]):
   import numpy as np

   data_py = dataset.active_scalars

   if data_py is None: #Check if data exists
       raise RuntimeError("No data array found!")

   data_py[:] = np.roll(data_py, shift[0], axis=0)
   data_py[:] = np.roll(data_py, shift[1], axis=1)
   data_py[:] = np.roll(data_py, shift[2], axis=2)

   dataset.active_scalars = data_py
   print('Data has been shifted uniformly.')
```

That last setting of the active scalars is not actually necessary since the view of the data is being modified. But I kept it around for consistency.

psavery:
[Here](https://github.com/OpenChemistry/tomviz/pull/1958/commits/3ef354b5a8a23aeafca154bfb43691102f6f971b#diff-cd7c79c4ecc6336d10d78ecec71c8173) is a working example of the changes for the new `transform()` function. Live updates seem to be working as well.

psavery:
I changed the name of the threaded `DataObject` class to `Dataset` to better match the variable names in the operators.

tomviz/python/Recon_SIRT.py
```diff
@@ -23,14 +22,14 @@ def transform_scalars(self, dataset, Niter=10, stepSize=0.0001,
...
-        tiltAngles = utils.get_tilt_angles(dataset)
+        tiltAngles = dataset.tilt_angles
 
         #remove zero tilt anlges
         if np.count_nonzero(tiltAngles) < tiltAngles.size:
             tiltAngles = tiltAngles + 0.001
 
         # Get Tilt Series
-        tiltSeries = utils.get_array(dataset)
+        tiltSeries = dataset.active_scalars
```
cryos: This looks much more natural

tomviz/python/Recon_SIRT.py
```diff
@@ -92,19 +90,19 @@ def transform_scalars(self, dataset, Niter=10, stepSize=0.0001,
 
                 # Give 4 updates for first iteration.
                 if Nupdates != 0 and i == 0 and (s + 1) % (Nslice//4) == 0:
-                    utils.set_array(child, recon)
+                    child.active_scalars = recon
```
cryos: This too, I think this is a big improvement, and I assume behind the scenes things are working in the same sense as they were before with updates/live updates.

tomviz/python/tomviz/_internal.py
```diff
@@ -22,6 +22,7 @@ def in_application():
 
 if in_application():
     import tomviz._wrapping
+    import vtk
```
cryos: Do we need to import all of VTK, last I checked we built VTK as modules, and we worked hard to improve start up time by just importing the modules we needed. I.e. `from vtk import vtkImageData`, a quick grep shows a few places where we have started importing more but I would prefer to get back to just the required modules.

psavery: Okay, I think I was actually just importing it to get `vtkDataObject`. I'll change it to only get `vtkDataObject`, and I'll check to make sure I don't import all of vtk in other places in this PR.

psavery: Fixed.

tomviz/python/Recon_SIRT.py
psavery: Yep! That is right. I made sure live updates are working, and the setting of the final recon dataset as well.

tomviz/PythonUtilities.h
```diff
@@ -178,6 +185,9 @@ class Python
   /// vtkPythonPythonInterpreter::PrependPythonPath(...)  to do the work.
   static void prependPythonPath(std::string dir);
 
+  /// Create a threaded Dataset object for operators to use
```
cjh1: I think the term 'threaded' is a little overloaded, may be we should  stick with 'internal pipeline' and 'external pipeline'?

tomviz/operators/OperatorPython.cxx
```diff
@@ -417,14 +416,29 @@ bool OperatorPython::applyTransform(vtkDataObject* data)
...
   Python::Object result;
   {
     Python python;
 
     Python::Tuple args(1);
-    args.set(0, pydata);
+
+    // This will look something like:
+    // <function transform_scalars at 0x7fd90c866158>
+    QString name = d->TransformMethod.toString();
```
cjh1: Lets move this to the else block, its only used there?

tomviz/python/tomviz/threaded_dataset.py
```diff
@@ -0,0 +1,54 @@
...
+    def active_scalars(self):
+        return utils.get_array(self._data_object)
+
+    @active_scalars.setter
+    def active_scalars(self, v):
+        utils.set_array(self._data_object, v)
+
+    @property
+    def scalars(self):
+        return { key: value for key, value in utils.arrays(self._data_object) }
```
cjh1: This is the only part of the API that I am not sure about. We have always used scalars to mean a single component.  `scalars` here is referring to multiple components were a `active_scalars` is referring to a single component.

cjh1: @cryos Do you have any thoughts?

tomviz/operators/OperatorPython.cxx
psavery: Actually, I was using it to check if the function to be called is "transform_scalars" or "transform", and then pass the appropriate arguments.

It is used in the `if()` and `else if()` conditions.

cjh1: Ok, I see. Its probably better to get the value of `__name__` then you can do a direct comparison? In theory the format of the output of `str(...}` could change, although I agree it will probably always include the function name :smile:

psavery: Done, this was fixed in the latest commit.

tomviz/PythonUtilities.h
psavery: Fixed.

tomviz/operators/OperatorPython.cxx
cjh1: Looks great thanks!

tomviz/python/tomviz/threaded_dataset.py
psavery: We could potentially rename `scalars` to be `all_scalars`, and then rename `active_scalars` to just `scalars`, which is similar to VTK (`GetScalars` in VTK returns the active scalars if no name is provided).

There isn't really an equivalent in VTK for `all_scalars`, though.

tomviz/python/BinaryThreshold.py
```diff
@@ -3,8 +3,7 @@
```
psavery: Below is an example of an ITK operator using the new `transform()` API. Let me know what you think...

@cjh1 @cryos

tomviz/python/SegmentPores.py
```diff
@@ -360,10 +357,10 @@ def transform_scalars(self, dataset, minimum_radius=0.5, maximum_radius=6.):
                 .GetArrayFromImage(opened)
 
             # temp
-            label_buffer = label_buffer.copy()
-            label_map_dataset = vtkImageData()
-            label_map_dataset.CopyStructure(dataset)
-            utils.set_array(label_map_dataset, label_buffer, isFortran=False)
+            # Transpose the data to Fortran indexing
+            label_buffer = label_buffer.copy().transpose([2, 1, 0])
+            label_map_dataset = dataset.create_child_dataset()
```
psavery: It looks like the histogram range for this operator gets bigger with these changes. I am wondering if it might be because `dataset.create_child_dataset()` [modifies the spacing](https://github.com/OpenChemistry/tomviz/blob/9a05a7f0dbdc18ff8720f0c58e6063b652b2ecba/tomviz/python/tomviz/utils.py#L186) of the copied dataset. I might be able to add an argument to make the child dataset without modifying the spacing.

I am not really sure if this matters either, though.

Old
===
segment_pores_old

New
===
segment_pores_new

tomviz/python/LabelObjectAttributes.py
```diff
@@ -26,8 +26,7 @@ def transform_scalars(self, dataset):
 
         returnValues = None
 
-        scalarType = dataset.GetScalarType()
-        if scalarType == VTK_FLOAT or scalarType == VTK_DOUBLE:
```
psavery: Here is an example of a change in the way we were checking to see if the type was a float. Instead of checking the vtk type, we check the numpy type with `np.issubdtype(dtype, np.floating)`.

tomviz/python/BinaryThreshold.py
```diff
@@ -68,11 +66,12 @@ def transform_scalars(self, dataset, lower_threshold=40.0,
             self.progress.message = "Creating child data set"
 
             # Set the output as a new child data object of the current data set
-            label_map_dataset = vtkImageData()
-            label_map_dataset.CopyStructure(dataset)
+            label_map_dataset = dataset.create_child_dataset()
+
+            result = itk.GetArrayFromImage(threshold_filter.GetOutput())
+            # Transpose the data to Fortran indexing
+            label_map_dataset.active_scalars = result.transpose([2, 1, 0])
```
psavery: Before, `itkutils.set_array_from_itk_image()` would call `utils.set_array()` with `isFortran=False` as an argument. This argument indicated that the indexing needed to be swapped from (i, j, k) to (k, j, i). This is done for ITK outputs because ITK uses C indexing, whereas we use Fortran indexing for tomviz.

Now, we are just calling `result.transpose([2, 1, 0])` to swap the indexing within the operator so that tomviz receives the data with the correct indexing.

Note that Fortran indexing vs. C indexing is not the same as Fortran vs C memory ordering.

psavery:
One thing I noticed in my testing is that, for the SIRT reconstruction, the output of the external pipeline is oriented the wrong way compared to the output of the internal pipeline.

Additionally, the external pipeline live reconstruction progress data is smaller in visual size than the internal pipeline live reconstruction progress data.

I will be looking into these.

psavery:
Actually, I just tried the above out on the master branch with the latest `tomviz/pipeline` docker image, and I got the same issues. So it looks like the above issues are not new to this branch.

psavery:
I think this might be mostly ready for review. I haven't experienced any major issues in my testing of the new operator API. I did encounter a few bugs in the testing (a couple of which I haven't posted to the issues yet), but it seems that they occur in the master branch as well, and are thus not new to this branch.

I posted an issue about the new behavior of Pad_Data. I also mentioned above the change in numbers in some of the ITK operators, which perhaps should be investigated more. But I think overall, this branch is ready for review.

tomviz/python/SegmentPores.py
psavery: I found out that the difference between these images is caused by the fact that the spacing does not get transferred from VTK to ITK using the [new method for getting an ITK image](https://github.com/OpenChemistry/tomviz/pull/1958/files#diff-76468b10025d5704a77ba9aa0448ea81R301). If I set the spacing immediately after that line, I get the old behavior.

We might need to figure out the best way to transfer the spacing to the ITK image, especially in the external pipeline case where the data is coming directly from a numpy array rather than a vtkDataObject.

psavery: This was fixed by adding properties that set and access the spacing in both the internal and external pipelines. The spacing is set on the ITK images, and the results are now the same.

tomviz/python/BinaryThreshold.py
psavery: We moved this logic to a new `itkutils` function.

tomviz/python/tomviz/threaded_dataset.py
psavery: We resolved this through conversation.

cjh1:
@psavery There are some formatting issue we should resolve before merging.

psavery:
@cjh1 Formatting issues are resolved.

psavery:
Glad to get it merged :slightly_smiling_face:

PR #1960, opened 2019-10-01
Thick slice

bnmajor:
Adds the ability to view a composite of  N slices as well as select the compositing mode to use (summation, mean, maximum, or minimum).


tomviz/vtkNonOrthoImagePlaneWidget.cxx
```diff
@@ -112,6 +112,7 @@ vtkNonOrthoImagePlaneWidget::vtkNonOrthoImagePlaneWidget()
   this->Reslice->TransformInputSamplingOff();
   this->Reslice->AutoCropOutputOff();
   this->Reslice->MirrorOff();
+  this->Reslice->SetSlabModeToSum();
```
cryos: Is there a reason to do this here, rather than if the user sets the slab mode?

tomviz/vtkNonOrthoImagePlaneWidget.cxx
```diff
@@ -1807,3 +1808,19 @@ void vtkNonOrthoImagePlaneWidget::SetOpacity(double opacity)
...
+  this->Reslice->SetSlabNumberOfSlices(slices-1);
+}
+
+void vtkNonOrthoImagePlaneWidget::SetThickSliceMode(int mode)
+{
+  if (!this->Reslice) {
+    return;
+  }
+  this->Reslice->SetSlabMode(mode);
+}
```
cryos: Missing newline at end of file.

tomviz/vtkNonOrthoImagePlaneWidget.cxx
bnmajor: I set the mode to sum because you said the typical use case would be summation, but the slab mode defaults to mean.

tomviz/vtkNonOrthoImagePlaneWidget.cxx
bnmajor: Fixed.

tomviz/vtkNonOrthoImagePlaneWidget.cxx
cryos: OK, and I guess that is what is happening with the widgets, makes sense. Thanks.

Issue #1961, opened 2019-10-08
Add back in pink color map


cryos:
We should add back pink_Matlab back in, this is used by some of our neutron users

cryos:
An example figure created using Tomviz and the pink_Matlab color map - https://www.nature.com/articles/s41598-019-49528-9/figures/1

PR #1965, opened 2019-10-11
Add pink_Matlab color map back into presets

bnmajor:


bnmajor:
Should fix #1961

PR #1971, opened 2019-10-15
Clip

bnmajor:
The current clip module is a work in progress.
- Currently using the `vtkNonOrthoImagePlaneWidget` to create the clipping plane since the class already contains all of the plane functionality needed. However it does do more than is needed as it also creates and manages a slice, which is not needed here.
- Despite having used the slice module as a template and starting point, this current implementation is very laggy and occasionally unresponsive to mouse interaction (the slider works consistently, albeit slowly)
- Presently only one, orthogonal clipping plane can be handled at a time
- The actual clipping plane could probably stand to be more obvious, it currently kind of blends into the outline
- The icon is just the slice icon at the moment

### DIFF
--- a/tomviz/CMakeLists.txt
+++ b/tomviz/CMakeLists.txt
@@ -212,6 +212,8 @@ set(SOURCES
 list(APPEND SOURCES
   modules/Module.cxx
   modules/Module.h
+  modules/ModuleClip.cxx
+  modules/ModuleClip.h
   modules/ModuleContour.cxx
   modules/ModuleContour.h
   modules/ModuleContourWidget.cxx
--- a/tomviz/modules/Module.h
+++ b/tomviz/modules/Module.h
@@ -15,6 +15,7 @@
 
 class QWidget;
 class vtkImageData;
+class vtkPlane;
 class vtkSMProxy;
 class vtkSMViewProxy;
 class vtkPiecewiseFunction;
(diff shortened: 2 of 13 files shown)


cryos:
Can you take a look at this @alesgenova and offer some input on getting the clipping to work for the plane.

bnmajor:
The clipping plane now works for one or more orthogonal or non-orthogonal planes. The properties panel currently re-uses the same slider and direction functionality that the slice module uses.
clippingPlanes
The border of the non-orthogonal slices does currently extend outside of the outline and should probably be clipped.
nonOrthoClippingPlane

cryos:
This doesn't seem to work for me on my Linux laptop using an embedded Intel GPU. Maybe we can take a look on Thursday, it can be tough to interact with it in the 3D view too.

bnmajor:
The clipping plane can now be toggled on or off and its direction can be inverted.

invertedPlane

hiddenPlanes

The most recent changes also remove the border from the clipping plane and instead increase the opacity of the plane. This keeps the plane clipped to the dimensions of the outline and keeps the plane visible when it is being moved. If it is preferred the border of the plane could be added back in.

clippedClippingPlane

cryos:
You have a commit with the subject "WIP" that should be edited out, or the description fixed. Just taking a look at this locally now.

bnmajor:
Yes, that commit should not have been there at all actually. I'm going to remove it altogether.

PR #1977, opened 2019-10-16
For HDF5 files, allow strides to be chosen for each axis

### LINKED ISSUE #1956: Don't stride tilt axis for HDF5 files
If the user chooses a stride, the stride is currently applied in all directions.

Don't stride the tilt axis if there is one.

psavery:
This allows strides to be chosen for each axis when subsampling
an HDF5 file.

This additionally allows tilt series files to be subsampled, since the user
can now choose which axes to stride, and they are not required to stride the
tilt axis.

Here is what the subsample widget now looks like:
image

Mostly fixes #1956
Fixes: #1978

### DIFF
--- a/tomviz/DataExchangeFormat.cxx
+++ b/tomviz/DataExchangeFormat.cxx
@@ -68,12 +68,27 @@ bool DataExchangeFormat::read(const std::string& fileName,
 
   dataSource->setData(image);
 
+  // Use the same strides and volume bounds for the dark and white data,
+  // except for the tilt axis.
+  QVariantMap darkWhiteOptions = options;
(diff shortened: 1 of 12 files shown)


psavery:
One thing I am unsure about is: when subsampling a tilt series file, such as the tilt series sample, should we swap the X and Z axes in the subsample widget?

The files are usually in C ordering with C indexing. So technically, the X axis is the tilt axis until we read it in and then swap the X and Z axes.

The example below shows what subsampling the tilt sample currently looks like, where the tilt axis is on X since it hasn't been read into the program yet.

image

PR #1983, opened 2019-10-30
Fix the dark and white field for internal

cryos:
These are vtkImageData objects, and need the same treatment as the main
data. When calling I just saw errors using the vtk_to_numpy call as it
hasn't gotten into the actual data array at that point. This needs more
testing but I have verified that I get numpy arrays of the expected
shape.

### DIFF
--- a/tomviz/python/tomviz/internal_dataset.py
+++ b/tomviz/python/tomviz/internal_dataset.py
@@ -50,11 +50,11 @@ class Dataset:
 
     @property
     def dark(self):
-        return np_s.vtk_to_numpy(self._data_source.dark_data)
+        return utils.get_array(self._data_source.dark_data)
 
     @property
     def white(self):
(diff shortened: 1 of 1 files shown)


cryos:
For reference I saw these errors:
```
  File "my_new_op", line 11, in transform                                        
  File "/home/marcus/build/tomvizsuper/tomviz/lib/site-packages/tomviz/internal_dataset.py", line 53, in dark                                                     
    return np_s.vtk_to_numpy(self._data_source.dark_data)                        
  File "/home/marcus/build/tomvizsuper/paraview/lib/python3.6/site-packages/vtkmodules/util/numpy_support.py", line 216, in vtk_to_numpy                          
    typ = vtk_array.GetDataType()                                                
AttributeError: 'vtkCommonDataModelPython.vtkImageData' object has no attribute 'GetDataType'
Failed to execute the script.
```

Issue #1989, opened 2019-11-05
Selecting multiple data sets does not work


ercius:
**Describe the bug**
File open dialog accessed from File - Open data (also Ctrl - o) allows the user to select multiple files (i.e. by holding shift). However, only one file is actually opened.

**To Reproduce**
1. File - Open data
2. Select more than one data set in the file dialog
3. Press Open
4. Only one data set (the first one selected) is opened

**Expected behavior**
Two possibilities exist. Either is acceptable.
1. Allow selecting multiple data sets in the file dialog and then open all those as separate data sets or
2. Only allow selecting one data set in the file dialog.

**Screenshots**
![image](https://user-images.githubusercontent.com/15657108/68233514-303cd500-ffb4-11e9-99dc-00ef6225ba60.png)

![image](https://user-images.githubusercontent.com/15657108/68233573-48acef80-ffb4-11e9-9291-3f0eae8dde3a.png)

**Desktop (please complete the following information):**
 - Windows 10 
 - tomviz 1.7.0

**Hardware (please complete the following information):**
 - Razer Blade 13 inch
 - GEFORCE MX150

Other notes:
I am not sure how this will affect opening a set of multiple TIF files saved as a single data set. Some programs save each slice of the data as 1 TIF file. Opening each TIF file separately is then not the desired behavior.

bnmajor:
This should be fixed in #1994. When multiple files are selected via the 'Open Data' menu they will all be read in as separate data sets. 

Previously, selecting multiple `.tiff` files from the 'Open Data' menu would prompt the user to create an image stack. Selecting multiple `.tiff` files now will read each file in separately. Image stacks can still be created with the 'Open Stack' menu option.

psavery:
Fixed in #1994.

PR #1994, opened 2019-11-07
Allow users to load multiple files at once

bnmajor:
All data sets selected from the 'Open Data' option in the File menu will be
opened as seperate data sets. Image stacks can still be read in through the 'Open
Stack' option.


cjh1:
Fixes #1989

PR #2000, opened 2019-11-11
Update python on windows for azure-pipelines

psavery:
This updates python to 3.7.5 on Windows, and it requires a re-build of ParaView, since the cached ParaView on Windows used python 3.7.4.


psavery:
Closing this one in favor of #2010

Issue #1997, opened 2019-11-08
Weird issues with slice operator


ercius:
**Describe the bug**
Ive run into a few strange issues with the slice operator in 1.7.0. Some of them seem hard to reproduce so I have not reported them. However, now I can reproduce one. This one seems trivial but maybe it points to a deeper problem.

This has to do with moving the slice (in orthogonal mode) using the keyboard. The slice will change in one direction (right or positive) for a few clicks and then stop. In the other direction (left or negative) you can always change the slice number. Its weird, but it might point to a lower issue which will fix some other bugs I have seen often, but would be difficult to reproduce.

Some issues seem to be connected to large datasets with real space pixel coordinates with large values. This issue is also connected to the same concept.

**To Reproduce**
1. Load the example tilt series
2. Click on the name of the data set in the pipeline to bring up the properties menu
3. Change the Z size of the data set to 1000 nm. (This is wrong but shows the effect)
![image](https://user-images.githubusercontent.com/15657108/68516442-eb75a000-0238-11ea-968a-170b1958281b.png)
4. Click on the slice operator in the pipeline.
5. Click on the blue bar which allows you to change the slice number with the mouse and keyboard. This now has keyboard focus
![image](https://user-images.githubusercontent.com/15657108/68516513-16f88a80-0239-11ea-8263-97e2422f5fd9.png)
6. If you press the left keyboard button I can always change the slice number. If I press the right keyboard button sometimes the bar moves and updates the viewer. Sometimes it does not.

**Expected behavior**
The keyboard should move the slice in both directions. This trivial issue might be indicative of a larger issue with the slice operator. Sometimes when I load a data set the slice gets stuck on one side of the data set (at the max or min slice value) and will not move. I think its connected with the real space coordiantes, but Im not sure.

**Screenshots**

**Desktop (please complete the following information):**
 - Operating system:  Windows 10
 - Tomviz version: 1.7.0 and 1.7.0 nightlies

**Hardware (please complete the following information):**
 - Device: Razer Stealth 13
 - Graphics card: NVIDIA MX150

**Additional context**
I can try to reproduce the other issues with the slice operator. It requires sharing large data sets though and might be better done over a video chat.

cryos:
This would be good to look at @alesgenova

cryos:
I think this is now resolved too

alesgenova:
Issue is still partially broken, I'm working on it

PR #2014, opened 2019-12-09
Offset plane bounds

bnmajor:
Offset the plane bounds so that the plane does not get clipped and become
unusable when it reaches the bounds of the data source. This fixes a bug that
would make the plane impossible to select by mouse if it was moved to the maximum
position in one of the orthogonal directions.


bnmajor:
This addresses one of the issues mentioned in #1997

Issue #1968, opened 2019-10-14
Pause the pipeline


cryos:
We should allow for pausing the pipeline, and starting it. Also, when loading a state file we should ask if the user wants to run the pipeline. We would need to offer the ability to run the pipeline.

PR #2019, opened 2019-12-11
Optionally prevent pipelines from executing on state load

alesgenova:
Fixes: #1968 


cjh1:
@cryos I only had one comment with @alesgenova is going to push a fix for and then this should be good to go.

PR #2021, opened 2019-12-13
Rebuild paraview on azure-pipelines

psavery:
Mac is failing currently. It looks like it may be caused by some
dependencies being updated on the VM, which invalidates the paraview
build cache.

Re-build paraview to fix this.


psavery:
Looks like this fixes the pipeline...

PR #2022, opened 2019-12-13
Add options to view dark and white data

psavery:
This adds two menu items under "View": "Show Dark Data" and
"Show White Data". If the currently activated data source
contains dark and white data, these two menu items will be
enabled. Selecting the menu items results in a blocking
pop-up window that shows an image slice of the dark/white data.
The image slice uses the same color map as the data source.
The interactor is currently vtkInteractorStyleRubberBand2D, just
because that is what was used for the image slices in the
RotateAlignWidget.

This currently views the dark/white data as a slice along the
z axis, which might not always be the case.

### DIFF
--- a/tomviz/ViewMenuManager.cxx
+++ b/tomviz/ViewMenuManager.cxx
@@ -11,7 +11,14 @@
 
 #include <vtkCommand.h>
 #include <vtkGridAxes3DActor.h>
+#include <vtkImageProperty.h>
+#include <vtkImageSlice.h>
+#include <vtkImageSliceMapper.h>
+#include <vtkInteractorStyleRubberBand2D.h>
 #include <vtkProperty.h>
+#include <vtkRenderWindow.h>
+#include <vtkRenderer.h>
+#include <vtkScalarsToColors.h>
 #include <vtkTextProperty.h>
 
 #include <QAction>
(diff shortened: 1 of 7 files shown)


tomviz/ViewMenuManager.cxx
```diff
@@ -11,7 +11,14 @@
 
 #include <vtkCommand.h>
 #include <vtkGridAxes3DActor.h>
+#include <vtkImageProperty.h>
```
cryos: I think that the view dark/white dialog is likely to need to grow in the future, and I don't think we want to introduce all of these classes to the ViewMenuManager. I think creating a SliceViewDialog would be preferable where we can change which slice is passed to it would offer better encapsulation, and let us grow this dialog as necessary.

tomviz/ViewMenuManager.cxx
```diff
@@ -199,4 +226,126 @@ void ViewMenuManager::setShowOrientationAxes(bool show)
...
+  updateDataSourceEnableStates();
+}
+
+void ViewMenuManager::updateDataSourceEnableStates()
+{
+  m_showDarkDataAction->setEnabled(m_dataSource && m_dataSource->darkData());
+  m_showWhiteDataAction->setEnabled(m_dataSource && m_dataSource->whiteData());
+}
+
+void ViewMenuManager::showDarkData()
```
cryos: The showDarkData and showWhiteData code should be combined, and just take the appropriate image as input rather than duplicating so much code. It might be better to simply offer a combobox where the user can switch between the two rather than making them click around.

PR #2034, opened 2020-02-14
Allow subsampling in the interactive python console

psavery:
The primary purpose of this PR is to allow subsampling options to be passed and used in the interactive python console. These subsampling options will only be used for HDF5 type files. Currently, this option is only available for creating a data source from a file.

However, this PR also has a few other changes in it:
1. [Fixed a key error](https://github.com/OpenChemistry/tomviz/commit/f96198eaaa7e0396526b2556912f8081109e18ee) for when the interactive python console is used with HDF5 files.
2. [Never pop up the subsampling dialog on state file load](https://github.com/OpenChemistry/tomviz/commit/1b3b3fcbdb42835e4055c0e677dc45ce151aa0e8), which prevents the subsample dialog from popping up when the python console is being used.
3. [Only correct invalid volume bounds](https://github.com/OpenChemistry/tomviz/commit/a4a72270aca4595d8966ec51a81c8866a8606a3e) during subsampling, which allows the user to pass in `-1` in the python console in order to use default values.

An example can be seen below.

```python
from tomviz import state
subsampleSettings = { 
    'strides': [1, 2, 2], 
    'volumeBounds': [0, 1400, 20, -1, 100, 800] 
}
ds = state.DataSource(
    fileNames=['/home/patrick/data/tomviz/bnl/tomo_00001/tomo_00001_subsampled/subsampled.h5'],
    subsampleSettings=subsampleSettings
)
```

If a negative number is passed for any volume bounds, the default is used (for lower bounds, the default is 0, and for upper bounds, the default is the max).

For the six numbers in the volume bounds, index `i * 2` is the lower bound for dimension `i`, and index `i * 2 + 1` is the upper bound for dimension `i`.

### DIFF
--- a/tomviz/python/tomviz/state/_schemata.py
+++ b/tomviz/python/tomviz/state/_schemata.py
@@ -186,10 +186,17 @@ class GradientOpacityMap(Schema):
         unknown = INCLUDE
 
 
+class SubsampleSettingsSchema(Schema):
+    # TODO: maybe we can force 'strides' to be of length 3,
+    # and volumeBounds to be of length 6?
+    strides = fields.List(fields.Int())
+    volumeBounds = fields.List(fields.Int())
+
+
 class ReaderSchema(Schema):
     fileNames = fields.List(fields.String)
     name = fields.String()
-    subsampleSettings = fields.Dict(values=fields.List(fields.Int()))
+    subsampleSettings = fields.Nested(SubsampleSettingsSchema())
 
     @post_load
     def make_reader(self, data, **kwargs):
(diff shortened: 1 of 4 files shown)


tomviz/python/tomviz/state/_schemata.py
```diff
@@ -189,6 +189,7 @@ class Meta:
 class ReaderSchema(Schema):
     fileNames = fields.List(fields.String)
     name = fields.String()
+    subsampleSettings = fields.Dict(values=fields.List(fields.Int()))
```
cjh1: I think here we can add a field definition for SubSampleSettings, that explicitly defines the fields expected?

psavery: Done.

PR #2039, opened 2020-02-17
Display voxel index and value when mouse is over slice

alesgenova:
voxel_value

Signed-off-by: Alessandro Genova <alessandro.genova@kitware.com>

### DIFF
--- a/tomviz/MainWindow.h
+++ b/tomviz/MainWindow.h
@@ -11,6 +11,7 @@
 #include <QScopedPointer>
 
 class QMenu;
+class vtkVector3i;
 
 namespace Ui {
 class MainWindow;
@@ -81,7 +82,7 @@ private slots:
 
   void setEnabledPythonConsole(bool enabled);
 
-  void onMouseOverVoxel(int i, int j, int k, double v);
+  void onMouseOverVoxel(const vtkVector3i& ijk, double v);
 
 private:
   Q_DISABLE_COPY(MainWindow)
--- a/tomviz/MainWindow.h
+++ b/tomviz/MainWindow.h
@@ -81,6 +81,8 @@ private slots:
 
   void setEnabledPythonConsole(bool enabled);
 
+  void onMouseOverVoxel(int i, int j, int k, double v);
+
 private:
   Q_DISABLE_COPY(MainWindow)
 
--- a/tomviz/Utilities.cxx
+++ b/tomviz/Utilities.cxx
@@ -1132,29 +1132,31 @@ bool vtkRescaleControlPoints(std::vector<vtkTuple<double, 4>>& cntrlPoints,
   return true;
 }
 
-bool getVoxelValue(vtkImageData* data, double point[3], int indices[3],
-                   double& scalar)
+double getVoxelValue(vtkImageData* data, const vtkVector3d& point,
+                     vtkVector3i& indices, bool& ok)
 {
-  auto pointId = data->FindPoint(point);
+  vtkVector3d p(point.GetData());
+  auto pointId = data->FindPoint(p.GetData());
   if (pointId < 0) {
-    return false;
+    ok = false;
+    return 0;
   }
 
-  double position[3];
-  data->GetPoint(pointId, position);
+  vtkVector3d position;
+  data->GetPoint(pointId, position.GetData());
 
-  double bounds[6];
-  data->GetBounds(bounds);
+  vtkVector3d origin;
+  data->GetOrigin(origin.GetData());
 
-  double spacing[3];
-  data->GetSpacing(spacing);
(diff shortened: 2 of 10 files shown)


tomviz/vtkNonOrthoImagePlaneWidget.cxx
```diff
@@ -749,8 +756,21 @@ void vtkNonOrthoImagePlaneWidget::OnMouseMove()
...
+    if (this->VoxelTimerId == -1) {
+      this->Interactor->AddObserver(vtkCommand::TimerEvent,
+                                    this->VoxelTimerCommand);
+      this->VoxelTimerId = this->Interactor->CreateOneShotTimer(100);
+    }
+    // Uncomment else block if event should only emitted after the mouse is done
+    // moving
+    // else {
+    //   this->Interactor->ResetTimer(this->VoxelTimerId);
+    // }
```
cjh1: Do we need this block?

alesgenova: Since the logic is not necessarily easy to follow, I figure I might leave those instructions there.
Essentially, at the moment we are throttling the firing of the event, if we uncomment those lines we would be debouncing it.

tomviz/modules/Module.h
```diff
@@ -170,6 +170,9 @@ public slots:
   void clipFilterUpdated(vtkPlane* plane, bool newFilter);
   void updateClipFilter(vtkPlane* plane, bool newFilter);
 
+  /// Emitted when the mouse is over a voxel
+  void mouseOverVoxel(int, int, int, double);
```
cryos: I think we should use vtkVector3i here instead of three ints.

tomviz/MainWindow.h
```diff
@@ -81,6 +81,8 @@ private slots:
 
   void setEnabledPythonConsole(bool enabled);
 
+  void onMouseOverVoxel(int i, int j, int k, double v);
```
cryos: Let's use vtkVector3i here

tomviz/Utilities.cxx
```diff
@@ -1130,4 +1131,40 @@ bool vtkRescaleControlPoints(std::vector<vtkTuple<double, 4>>& cntrlPoints,
   }
   return true;
 }
+
+bool getVoxelValue(vtkImageData* data, double point[3], int indices[3],
```
cryos: Let's define API with vtkVector3d and vtkVector3i, why not return the value, and take an optional bool by reference for ok

cryos: You can use .GetData() to get a pointer to the double* or int* for the vtkVector types

tomviz/Utilities.cxx
```diff
@@ -1130,4 +1131,40 @@ bool vtkRescaleControlPoints(std::vector<vtkTuple<double, 4>>& cntrlPoints,
...
+                   double& scalar)
+{
+  auto pointId = data->FindPoint(point);
+  if (pointId < 0) {
+    return false;
+  }
+
+  double position[3];
```
cryos: vtkVector3d

tomviz/Utilities.cxx
```diff
@@ -1130,4 +1131,40 @@ bool vtkRescaleControlPoints(std::vector<vtkTuple<double, 4>>& cntrlPoints,
...
+  data->GetPoint(pointId, position);
+
+  double bounds[6];
+  data->GetBounds(bounds);
+
+  double spacing[3];
```
cryos: vtkVector3d

tomviz/Utilities.h
```diff
@@ -229,6 +230,10 @@ void openHelpUrl(const QString& path = "");
 bool vtkRescaleControlPoints(std::vector<vtkTuple<double, 4>>& cntrlPoints,
                              double rangeMin, double rangeMax);
 
+/// Get the value of a voxel at the given world coordinates
+bool getVoxelValue(vtkImageData* data, double point[3], int ijk[3],
```
cryos: I think we should define using vtkVector3[d|i], and just return the value.

tomviz/vtkNonOrthoImagePlaneWidget.h
```diff
@@ -498,6 +516,10 @@ class vtkNonOrthoImagePlaneWidget : public vtkPolyDataSourceWidget
   void GenerateArrow(); // generate the default arrow
   void UpdateArrowSize(); // update the arrow to be visible based on camera pos
 
+  bool PickPointOnSlice(
+    int displayPos[2],
```
cryos: Indentation looks odd, I think vtkVector2i and vtkVector3d would be better to use here, and in the API.

PR #2042, opened 2020-02-20
Print messages from external python as they occur

psavery:
Previously, when messages were printed in external python, they would not appear in the terminal or the tomviz "Messages" box until the operator completed.

This update gets tomviz to print out messages from the external python in both the terminal and in the tomviz "Messages" box as they occur, rather than waiting for the operator to complete.

Here's my test transform I was working with:
```python
def transform(dataset):
    import time
    import sys

    print('Message to stdout')
    print('Sleeping 10 seconds...')
    time.sleep(10)
    print('Message to stderr', file=sys.stderr)
    print('Sleeping 10 seconds...')
    time.sleep(10)
    print('Done!')
```

And the output:
```
[2020-02-20 15:15:55,950] INFO: Executing pipeline on /tmp/tomviz-jdpLTO/original.h5
Operator doesn't support progress updates.
Message to stdout
Sleeping 10 seconds...
[2020-02-20 15:15:55,974] INFO: Executing 'test_script' operator
Pipeline started in external python!
Sleeping 10 seconds...
Message to stderr
Done!
[2020-02-20 15:16:15,994] INFO: Execution complete.
[2020-02-20 15:16:15,995] INFO: Writing transformed data.
[2020-02-20 15:16:16,040] INFO: Write complete.
```

This seems good for the most part, except the two lines:
```
[2020-02-20 15:15:55,974] INFO: Executing 'test_script' operator
Pipeline started in external python!
```

appear to be printed later than expected. I'm not sure why. It looks like they should be occurring [before the transform](https://github.com/OpenChemistry/tomviz/blob/52954a0d680f21ff6b70e020df8f32dbf46f5352/tomviz/python/tomviz/executor.py#L617-L619).

Also, the `Message to stderr` occurs after the second `Sleeping 10 seconds...`, which is kind of weird. Maybe it has something to do with the signals being different: one is `QProcess::readyReadStandardError` and one is `QProcess::readyReadStandardOutput`.

### DIFF
--- a/tomviz/ExternalPythonExecutor.cxx
+++ b/tomviz/ExternalPythonExecutor.cxx
@@ -1,6 +1,8 @@
 /* This source file is part of the Tomviz project, https://tomviz.org/.
    It is released under the 3-Clause BSD License, see "LICENSE". */
 
+#include <QRegularExpression>
+
 #include "ExternalPythonExecutor.h"
 #include "DataSource.h"
 #include "EmdFormat.h"
@@ -185,11 +187,11 @@ void ExternalPythonExecutor::onStdOutReceived()
 {
   QString s = m_process->readAllStandardOutput();
   m_receivedStdOut += s;
-  // Since qDebug() will already print a newline, chop off the newline
-  // from the string if it is there.
-  if (!s.isEmpty() && s[s.size() - 1].isSpace()) {
-    s.chop(1);
-  }
+
+  // Since qDebug() will already print a newline, get rid of the
+  // newline from the string if it is there.
+  auto regexp = QRegularExpression("(?:\r\n|\n)$");
+  s = s.remove(regexp);
   qDebug().noquote() << s;
 }
 
@@ -197,11 +199,11 @@ void ExternalPythonExecutor::onStdErrReceived()
 {
   QString s = m_process->readAllStandardError();
   m_receivedStdErr += s;
-  // Since qDebug() will already print a newline, chop off the newline
-  // from the string if it is there.
-  if (!s.isEmpty() && s[s.size() - 1].isSpace()) {
-    s.chop(1);
-  }
+
+  // Since qDebug() will already print a newline, get rid of the
+  // newline from the string if it is there.
+  auto regexp = QRegularExpression("(?:\r\n|\n)$");
+  s = s.remove(regexp);
   qDebug().noquote() << s;
 }
 
--- a/tomviz/ExternalPythonExecutor.cxx
+++ b/tomviz/ExternalPythonExecutor.cxx
@@ -22,6 +22,8 @@ Pipeline::Future* ExternalPythonExecutor::execute(vtkDataObject* data,
                                                   QList<Operator*> operators,
                                                   int start, int end)
 {
+  m_receivedStdOut.clear();
+  m_receivedStdErr.clear();
 
   auto future = ExternalPipelineExecutor::execute(data, operators, start, end);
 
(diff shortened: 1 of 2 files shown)


psavery:
It looks like the mis-ordered
```
[2020-02-20 15:15:55,974] INFO: Executing 'test_script' operator
Pipeline started in external python!
```

lines only occurs if you print things really early in the transform (like immediately before/after the imports). It might be okay to just ignore these mis-ordered prints. I'm not sure.

cjh1:
"Pipeline started in external python!" is the interesting on as it come from a qDebug() statement. Is it because we are doing `qDebug("Pipeline started in external python!")` rather than `qDebug() << "Pipeline started in external python!";`

cjh1:
"[2020-02-20 15:15:55,974] INFO: Executing 'test_script' operator" is coming from a python logger so there might be some buffering going on their ...

psavery:
@cjh1 It looks like there are a few signals that get fired for `Pipeline started in external python!` to get printed, including `ProgressReader::pipelineStarted()`. Maybe that's why it's delayed.

tomviz/ExternalPythonExecutor.cxx
```diff
@@ -174,4 +181,28 @@ QString ExternalPythonExecutor::commandLine(QProcess* process)
     .arg(process->arguments().join(" "));
 }
 
+void ExternalPythonExecutor::onStdOutReceived()
+{
+  QString s = m_process->readAllStandardOutput();
+  m_receivedStdOut += s;
+  // Since qDebug() will already print a newline, chop off the newline
+  // from the string if it is there.
```
cjh1: Won't this add a newline for those that don't have one ( for example write with `write` )?

psavery: Yeah, it will. But getting `qDebug` to not print a newline seems to be tough - I haven't found a way to not do it. And we need `qDebug` or the output won't appear in the tomviz "Messages" box.

cjh1: Worse is it will remove a character that is not necessarily a `\n'?

psavery: Yeah, but I'm not sure if Windows returns `\n`. It might return something else instead. I'll take a look.

psavery: @cjh1 In the latest commit, I started using a [regular expression](https://github.com/OpenChemistry/tomviz/pull/2042/commits/2a095cc5af5828adeed93e3ba8b2d711cc6c2f75#diff-4976d2c27844f18f6ff4d5c847ee0d0fR193), which is probably a much better way to remove the newline at the end. The regular expression only matches with either a single `\n` (unix) or a single `\r\n` (windows) at the very end of the line. If there is one of these carriage returns at the end of the line, that carriage return gets removed.

A newline will still get added if a user prints via `sys.stdout.write()`, but that is kind of difficult to avoid here.

Just to record this here: if we ever decide in the future that we really don't want the newlines to be added, we can look into using the `vtk` printing functions. I think our tomviz message box is also hooked up to them.

cjh1:
@psavery Do we know why the Mac build is failing?

psavery:
@cjh1 I think it might just be an out-of-date paraview build cache. I can submit a PR to change the cache number to see if it builds correctly.

psavery:
To update here, it is not being caused *only* by an out-of-date paraview build cache. Looks like gtest is failing to compile on Mac. I will look into why. But I'll also merge this now since this PR is most likely not causing any issues on the Mac.

PR #2045, opened 2020-02-24
Remove proxies from slice module

psavery:
Before, the properties panel would communicate with the slice widget
like so: 
```
Qt Properties Panel <---> Properties Proxy <---> Slice Widget
```

With this PR, the middle step (Properties Proxy) is removed, and 
the properties panel communicates with the slice widget directly.

Additionally, the pass through filter is removed, and the slice widget
is connected directly to the data source.

### DIFF
--- a/tomviz/modules/ModuleSlice.cxx
+++ b/tomviz/modules/ModuleSlice.cxx
@@ -21,21 +21,11 @@
 #include <vtkTrivialProducer.h>
 
 #include <pqCoreUtilities.h>
-#include <pqDoubleVectorPropertyWidget.h>
 #include <pqLineEdit.h>
-#include <pqProxiesWidget.h>
-#include <vtkPVArrayInformation.h>
-#include <vtkPVDataInformation.h>
-#include <vtkPVDataSetAttributesInformation.h>
 #include <vtkPVDiscretizableColorTransferFunction.h>
-#include <vtkSMPVRepresentationProxy.h>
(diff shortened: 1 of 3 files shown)


psavery:
There are several slice module settings where there are 3 copies of the setting. For instance, for the `sliceIndex` setting, the three copies are:

1. The value of the `IntSliderWidget` [m_sliceSlider](https://github.com/OpenChemistry/tomviz/blob/bed686cfdc66d7283bb986049aaba6fe181e3305/tomviz/modules/ModuleSlice.h#L114).
2. [m_slice](https://github.com/OpenChemistry/tomviz/blob/bed686cfdc66d7283bb986049aaba6fe181e3305/tomviz/modules/ModuleSlice.h#L117) in the `ModuleSlice` class.
3. The `SliceIndex` in the slice widget (accessible via [GetSliceIndex](https://github.com/OpenChemistry/tomviz/blob/bed686cfdc66d7283bb986049aaba6fe181e3305/tomviz/vtkNonOrthoImagePlaneWidget.h#L181) and [SetSliceIndex](https://github.com/OpenChemistry/tomviz/blob/bed686cfdc66d7283bb986049aaba6fe181e3305/tomviz/vtkNonOrthoImagePlaneWidget.h#L185)).

All three are supposed to be the same value at all times, I think. We could potentially remove number 2 for these cases, but I'm not aware of this causing a problem anywhere currently, so I'm not sure if it should be changed.

PR #2054, opened 2020-02-27
Swap axes of all scalars for multi-scalars data from HDF5

psavery:
When loading an HDF5 file containing angles, the X and Z axes are swapped so that the tilt axis is the fast axis. `vtkImagePermute`, however, only swaps the X and Z axes of the active scalars. This causes some big problems later if there are multiple scalar arrays in the image data.
    
This PR adds a unified function for swapping the axes that will swap axes of all of the scalars. The parts of the code that need to swap axes of the image data should call this function.
    
This also adds a common function for reading in the angles in GenericHDF5Format. The various HDF5 formats that need to read in angles now call this function.

Finally, this also adds in the ability of the GenericHDF5 format to read in angles. A list of possible places where the angles may be located in a generic HDF5 file was added. Currently, only "/angle" is on the list. When a generic HDF5 file is read, if a dataset is found at one of these locations, that dataset is read in as angles.

### DIFF
--- a/tomviz/FxiFormat.cxx
+++ b/tomviz/FxiFormat.cxx
@@ -11,7 +11,6 @@
 
 #include <vtkDataArray.h>
 #include <vtkImageData.h>
-#include <vtkImagePermute.h>
 #include <vtkNew.h>
 #include <vtkPointData.h>
 #include <vtkTrivialProducer.h>
@@ -42,18 +41,6 @@ bool FxiFormat::read(const std::string& fileName, vtkImageData* image,
 {
   std::string path = "/img_tomo";
   return readDataSet(fileName, path, image, options);
-}
-
-static void swapXAndZ(vtkImageData* image)
-{
-  if (!image)
-    return;
-
-  vtkNew<vtkImagePermute> permute;
-  permute->SetFilteredAxes(2, 1, 0);
-  permute->SetInputData(image);
-  permute->Update();
-  image->ShallowCopy(permute->GetOutput());
 }
 
 bool FxiFormat::read(const std::string& fileName,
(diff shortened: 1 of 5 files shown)


psavery:
This seems to be working in my tests.

tomviz/FxiFormat.cxx
```diff
@@ -125,62 +112,30 @@ bool FxiFormat::readWhite(const std::string& fileName,
...
-  array->SetNumberOfTuples(dims[0]);
-  if (!reader.readData(thetaNode, type, array->GetVoidPointer(0))) {
-    std::cerr << "Failed to read the angles\n";
-    return angles;
-  }
-
-  switch (array->GetDataType()) {
-    // This is done so we can ensure we read the type correctly...
-    vtkTemplateMacro(readAngleArray<VTK_TT>(array, angles));
+  std::string path = "/angle";
```
cryos: Is this what is used by HXN?

psavery: Yes it is.

tomviz/GenericHDF5Format.cxx
```diff
@@ -450,6 +551,23 @@ bool GenericHDF5Format::read(const std::string& fileName, vtkImageData* image,
     }
   }
 
+  // Look for some common places where there are angles, and
+  // load in the angles if we find them.
+  QVector<double> angles;
+  std::vector<std::string> placesToSearch = { "angle" };
```
psavery: @cryos Actually, this is the HXN part I think.

cryos: We should probably add "angles" to the search, I think that will be a common variant.

psavery: Fixed

PR #2055, opened 2020-02-28
Fix a few memory leaks in DockerExecutor

psavery:
There were a few cases where `deleteLater()` would not get called
on the QObject if an error occurred in the docker invocation. Fix
the memory leaks.


tomviz/DockerExecutor.cxx
```diff
@@ -228,6 +227,7 @@ void DockerPipelineExecutor::containerError(int containerExitCode)
                            QString("Docker logs failed with: %1\n\n%2")
                              .arg(exitCode)
                              .arg(logsInvocation->stdErr()));
+              logsInvocation->deleteLater();
               return;
```
psavery: I added a `deleteLater()` here rather than removing the return statement because I was not sure, if the docker logs invocation failed, whether we wanted to remove the container or not, which is done below.

PR #2064, opened 2020-03-05
Rebuild paraview on azure-pipelines

psavery:
It looks like the dependencies might be out-of-date for Linux. Force
a rebuild of ParaView.

It might be nice if we were to add some kind of automatic dependency
checking in the future, and get it to automatically rebuild ParaView
if the dependencies are out-of-date.

This will take a few hours. Let's see if it builds successfully.


psavery:
Closing this PR in favor of #2065

PR #2071, opened 2020-03-10
Pipeline templates

bnmajor:
Allow users to apply pre-made templates to their data source, or create, save, and apply their own pipeline templates.

pipeline-template


bnmajor:
This currently only includes one default template, based off of the docs for [alignment](https://tomviz.readthedocs.io/en/latest/alignment/) and [reconstruction](https://tomviz.readthedocs.io/en/latest/reconstruction/), but more could easily be created and added. Additionally, both operators and modules that have been applied to the data source are saved and applied from templates. If useful, users could be allowed to select whether or not they would like to remember modules when they save a template.

psavery:
I took a look at the azure pipeline failures, and it appears they might be legitimate. I get the same linking error locally on Linux. It looks like cmake is failing to find a file on Mac and Windows.

tomviz/CMakeLists.txt
```diff
@@ -512,6 +514,22 @@ foreach(file ${resource_files})
     COMPONENT runtime)
 endforeach()
 
+set(template_files
+  Back_Projection_Reconstruction.tvsm
+)
+
+# Install templates
+file(MAKE_DIRECTORY "${tomviz_BINARY_DIR}/share/tomviz/Templates")
```
cryos: Let's keep `templates` lower case for all paths please.

tomviz/CMakeLists.txt
```diff
@@ -512,6 +514,22 @@ foreach(file ${resource_files})
...
+foreach(file ${template_files})
+  list(APPEND SOURCES "templates/${file}")
+  execute_process(COMMAND ${CMAKE_COMMAND} -E ${script_cmd}
+    "${tomviz_SOURCE_DIR}/tomviz/templates/${file}"
+    "${tomviz_BINARY_DIR}/share/tomviz/Templates/${file}")
```
cryos: Same here

tomviz/templates/Back_Projection_Reconstruction.tvsm
```diff
@@ -0,0 +1,247 @@
...
+            "description": "",
+            "id": "0x56251acc7eb0",
+            "label": "Remove Bad Pixels",
+            "script": "def transform(dataset, threshold=None):\n    \"\"\"Remove bad pixels in tilt series.\"\"\"\n\n    import scipy.ndimage\n    import numpy as np\n\n    tiltSeries = dataset.active_scalars.astype(np.float32)\n\n    for i in range(tiltSeries.shape[2]):\n        I = tiltSeries[:, :, i]\n        I_pad = np.lib.pad(I, (1, 1), 'edge')\n\n        # calculate standard deviation in a 3 x 3 window\n        averageI2 = scipy.ndimage.filters.uniform_filter(I_pad ** 2)\n        averageI = scipy.ndimage.filters.uniform_filter(I_pad)\n        std = np.sqrt(abs(averageI2 - averageI**2))[1:-1, 1:-1]\n\n        medianI = scipy.ndimage.filters.median_filter(I_pad, 2)[1:-1, 1:-1]\n\n        #identiy bad pixels\n        badPixelsMask = abs(I - medianI) > std * threshold\n\n        I[badPixelsMask] = medianI[badPixelsMask]\n        tiltSeries[:, :, i] = I\n\n    # Set the result as the new scalars.\n    dataset.active_scalars = tiltSeries\n",
+            "type": "Python"
+        },
+        {
+            "id": "0x7f3e04014550",
+            "offsets": [
```
cryos: Will this be OK with image stacks that have different numbers of slices? I wonder what we need to do to make this more of a template, i.e. can we leave settings like this empty to then use the default values?

tomviz/SaveLoadTemplateReaction.cxx
```diff
@@ -0,0 +1,121 @@
...
+
+bool SaveLoadTemplateReaction::saveTemplate()
+{
+  bool ok;
+  QString text = QInputDialog::getText(tomviz::mainWidget(), tr("Save Pipeline Template"),
+                                         tr("Template Name:"), QLineEdit::Normal, QString(), &ok);
+  QString fileName = text.replace(" ", "_");
+  if (ok && !text.isEmpty()) {
+    QString path = QApplication::applicationDirPath() +
+                    "/../share/tomviz/templates/" + fileName + ".tvsm";
```
cjh1: In an installed version of Tomviz the application is unlikely to be able to write to this path. I think the template generated by the user whould be stored in the "Tomviz directory". This location is currently used for custom operators, see `MainWindow`. This is the code that defines the path:

```CPP
QStringList locations =
     QStandardPaths::standardLocations(QStandardPaths::HomeLocation);
    QString home = locations[0];
    QString path = QString("%1%2tomviz").arg(home).arg(QDir::separator());
```

We should move this code to a utility function and then use it in both places. We can have a sub directory called `templates`.

tomviz/MainWindow.cxx
```diff
@@ -990,4 +1001,25 @@ void MainWindow::syncPythonToApp()
   sync.call(args, kwargs);
 }
 
+void MainWindow::findPipelineTemplates() {
+  m_pipelineTemplates->clear();
+
+  QDir dir = QApplication::applicationDirPath() + "/../share/tomviz/templates/";
+  foreach (QString file, dir.entryList()) {
+    QString fileName = file.split(".")[0];
+    QString menuName = file.split(".")[0].replace("_", " ");
```
cjh1: If you use `entryInfoList` here this is all provided in `FileInfo` without the need to `split` anything.

tomviz/SaveLoadTemplateReaction.cxx
```diff
@@ -0,0 +1,121 @@
...
+  }
+  return false;
+}
+
+bool SaveLoadTemplateReaction::loadTemplate(const QString& fileName)
+{
+  QString path = QApplication::applicationDirPath() +
+                 "/../share/tomviz/templates/" + fileName + ".tvsm";
+  QFile openFile(path);
+  openFile.open(QIODevice::ReadOnly);
```
cjh1: We should check that we where successful in opening the file

tomviz/SaveLoadTemplateReaction.cxx
```diff
@@ -0,0 +1,121 @@
...
+  QJsonParseError error;
+  auto contents = openFile.readAll();
+  auto doc = QJsonDocument::fromJson(contents, &error);
+
+  // Get the parent data source, as well as the active (i.e. data and output)
+  auto activeParent = ActiveObjects::instance().activeParentDataSource();
+  auto activeData = ActiveObjects::instance().activeDataSource();
+  ModuleManager::instance().removeAllModules(activeData);
```
cjh1: Why do we remove all modules? It seems that if we don't touch any existing operators we shouldn't touch existing modules?

cjh1:
@bnmajor This works nicely in my testing.

tomviz/CMakeLists.txt
bnmajor: Fixed.

tomviz/CMakeLists.txt
bnmajor: Fixed.

tomviz/SaveLoadTemplateReaction.cxx
bnmajor: This has been moved into a utility function and is being used in `importCustomTransform` as well as for saving and loading template files.

tomviz/MainWindow.cxx
bnmajor: I seem to have overlooked that function completely at first, thank you. Fixed.

tomviz/SaveLoadTemplateReaction.cxx
bnmajor: Thank you, fixed.

tomviz/SaveLoadTemplateReaction.cxx
bnmajor: I think my initial thought was that if someone is loading a template that includes modules they might not want the default modules as well, but you're right - if it's being assumed that any existing operators should be left in place, the same should be done with modules to keep the behavior consistent.

cjh1: @cryos Do you have any thoughts/feels on what this behavior should be?

cryos: I think templates should be expected to be able to combine, i.e. I would add one template to another possibly. I wonder if we might actually add a clear all to the context menu, but I would rather default to appending templates to the end of a pipeline, allowing users to freely mix their starting point with canned templates.

tomviz/Utilities.h
```diff
@@ -178,6 +178,9 @@ QString readInPythonScript(const QString& scriptName);
 // of the built-in tomviz python operator scripts.
 QString readInJSONDescription(const QString& scriptName);
 
+// Get the path for the Tomviz directory
+QString getApplicationPath();
```
cryos: I think this name is a little confusing, I would prefer something along the lines of `userDataPath` to signify the data path. We usually avoid the use of get too, it is a Qt convention we follow. I am not sure we should combine getting a path with creating a directory, but I think we can live with it for simplicity. The Python scripts are looking in `.tomviz` and `tomviz` as far as I remember.

cryos:
I see you already changed the behavior, I think it is what most of us would expect. One minor change inline of a utility function name. I am just testing things out locally, but don't expect to turn up anything new.

PR #2072, opened 2020-03-11
Add full state files (.tvh5)

psavery:
The tvh5 file is a full state file, in that the tomviz state is saved,
and all of the data sources are saved as well.

The full state file is kind of like an EMD file. It has an EMD node at
/data/tomography, which is where the active data source is saved. But 
it is extended so that at the root of the EMD file, there is a
"tomviz_state" dataset, which contains the tomviz state, and there is
a "tomviz_datasources" group, which contains all of the data sources.
Each data source is saved as an EMD node, and the group each data
source is in is the id of the data source (which matches the id of the 
data source in the state file). The data source that is active is
created via a soft link to /data/tomography, so it is not saved
twice.

The user can save the state files via "File"->"Save State", and 
then selecting "Tomviz full state files" (which is now the default) 
from the format selector. The file will be saved as a ".tvh5" file, and 
it can be loaded back into tomviz via "Load State", or as a command
line argument after the `tomviz` executable.

### DIFF
--- a/tomviz/modules/ModuleManager.cxx
+++ b/tomviz/modules/ModuleManager.cxx
@@ -881,7 +881,21 @@ bool ModuleManager::deserialize(const QJsonObject& doc, const QDir& stateDir,
   d->dir = QDir();
   m_stateObject = QJsonObject();
 
-  // Now to restore all of our cameras...
+  // Restore the views to their state before the modules were added
+  setViews(views);
+
+  d->LastStateLoadSuccess = true;
+
+  if (d->RemaningPipelinesToWaitFor == 0) {
+    emit stateDoneLoading();
+  }
+  return true;
+}
+
+void ModuleManager::setViews(const QJsonArray& views)
+{
+  // This sets all views according to their settings in the view proxy.
+  // It should be called after the views have been deserialized.
   for (int i = 0; i < views.size(); ++i) {
     auto view = views[i].toObject();
     auto viewProxy =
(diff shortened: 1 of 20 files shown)


psavery:
I think this is mostly done. I just need to add a menu item to save the full state in the default location, and save it in a transactional way.

psavery:
We could also re-design the organization of this if that is desirable. For instance, we could make a `Tvh5Format` class that calls the EMD functions, if that might be preferred.

cjh1:
@psavery I think `Tvh5Format` is a good plan if its not too much work. `EmdFormat` is getting a little large.

tomviz/modules/ModuleManager.cxx
```diff
@@ -190,6 +190,27 @@ bool ModuleManager::hasRunningOperators()
   return false;
 }
 
+QList<DataSource*> ModuleManager::dataSources()
+{
+  QList<DataSource*> ret;
+  std::copy(d->DataSources.begin(), d->DataSources.end(),
+            std::back_inserter(ret));
```
cjh1: Why the copy?

psavery:
> @psavery I think `Tvh5Format` is a good plan if its not too much work. `EmdFormat` is getting a little large.

I think it shouldn't be too much work. I can move it.

cjh1:
@psavery This is working nicely. We need to disable or at least warn the user if they trying to write out a skinny state file after they have loaded a full state file, currently it seems just to work, but the state file has no file paths in it, which will cause some surprises when it is loaded.

tomviz/modules/ModuleManager.cxx
psavery: The `d->DataSources` variable is a `QList<QPointer<DataSource>>`. I was copying them to convert them to a `QList<DataSource*>`. We could potentially also return a `QList<QPointer<DataSource>>`, but I thought most places would prefer the shorter variable name (plus, other functions in `ModuleManager` are returning QLists to raw pointers as well).

psavery:
Maybe we should also potentially pop up the same kind of loading dialog that the `.tvsm` state file pops up when loading, just so that if it takes a long time to load, the user will know that it is loading, and not frozen for another reason?

psavery:
This should hopefully now also work in conjunction with the normal tomviz state file. I think it is ready for review.

PR #2073, opened 2020-03-12
Use PEP 508 URL dependencies to specify patched jsonpatch

cjh1:
This ensure that anyone using the setup.py to install tomviz's python packages  will get the right version of jsonpatch.


psavery:
This linux build failed because of a pytest, by the way. Not sure why.

cjh1:
@psavery Seems to have resolved itself

Issue #2058, opened 2020-03-03
Occasional crashing after most operations


psavery:
Unfortunately, there appears to be a crash that happens occasionally after most operations (see backtrace pasted below). It appears to be caused by the voxel features in `vtkNonOrthoImagePlaneWidget`. I believe it is a race condition because it does not happen consistently, and the backtrace shows that a timer function is causing it (`VoxelTimerFired()`). It happens for me maybe 10-20% of the time an operator is performed.

My best guess is that the crash occurs when the timer fires as the slice module is being moved down to the operator results, and the slice is in some kind of invalid state (or possibly deleted) that results in a crash.

```
Loguru caught a signal: SIGSEGV
Stack trace:
25      0x55824cafd88a _start + 42
24      0x7f970ff43b97 __libc_start_main + 231
23      0x55824cafa801 main + 993
22      0x7f9710ccb9b4 QCoreApplication::exec() + 148
21      0x7f9710cc290a QEventLoop::exec(QFlags<QEventLoop::ProcessEventsFlag>) + 314
20      0x7f9710d1d88f QEventDispatcherGlib::processEvents(QFlags<QEventLoop::ProcessEventsFlag>) + 95
19      0x7f97013866dc g_main_context_iteration + 44
18      0x7f9701386650 /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0(+0x4c650) [0x7f9701386650]
17      0x7f9701386417 g_main_context_dispatch + 743
16      0x7f9710d1d4f1 /usr/lib/x86_64-linux-gnu/libQt5Core.so.5(+0x2e34f1) [0x7f9710d1d4f1]
15      0x7f9710d1cd2e QTimerInfoList::activateTimers() + 1038
14      0x7f9710cc48d8 QCoreApplication::notifyInternal2(QObject*, QEvent*) + 280
13      0x7f971904d104 QApplication::notify(QObject*, QEvent*) + 676
12      0x7f971904583c QApplicationPrivate::notify_helper(QObject*, QEvent*) + 156
11      0x7f9710cf407b QObject::event(QEvent*) + 155
10      0x7f9710d00328 QTimer::timerEvent(QTimerEvent*) + 40
9       0x7f9710cfffc7 QTimer::timeout(QTimer::QPrivateSignal) + 55
8       0x7f9710cf3555 QMetaObject::activate(QObject*, int, int, void**) + 2149
7       0x7f9710cff160 QSignalMapper::map(QObject*) + 544
[...]

PR #2076, opened 2020-03-12
Fix crash cause by voxel value one-shot timer

alesgenova:
Fixes: #2058 

Signed-off-by: Alessandro Genova <alessandro.genova@kitware.com>


alesgenova:
@psavery After this fix I cannot get Tomviz to crash anymore. If you get the chance can you confirm it is fixed for you too?

PR #2078, opened 2020-03-12
Add SHARED to add_library(...)

cjh1:


cjh1:
@cryos This is what I needed to do to ensure a shared library gets built.

cryos:
We can do this too, I suspect that CMake changed something, this forces it to shared all the time. We probably have to do this to be more correct as I think we established static is entirely broken now.

cjh1:
@cryos So should this go in or did you have another approach?

cryos:
This can go in, then we need to assess the impact on platforms. I am not convinced this will work on Windows due to the lack of export macros, and we may need to come up with a better solution/rethink if that is the case. I can take that on if necessary.

As far as I can tell the Python module must be shared, and it must be loaded dynamically. So we need to maybe reduce the number of classes if we need to decorate them with export macros to things that are wrapped, and then enhance them. We can do this first and take a look at build results. I don't understand why this PR didn't trigger any builds.

psavery:
Looks like this did cause a linking error on Windows

cryos:
Yes, Windows defaults to not exporting symbols, and Tomviz was originally just an application. The static was added to aid in testing, and later used for Python. I will take a look.

cryos:
This won't work, closing, #2095 will hopefully get us where we need to be (it may need some further tweaking).

PR #2079, opened 2020-03-12
Python updates

cryos:


tomviz/CMakeLists.txt
```diff
@@ -594,6 +594,7 @@ configure_file(tomvizConfig.h.in tomvizConfig.h @ONLY)
 configure_file(tomvizPythonConfig.h.in tomvizPythonConfig.h @ONLY)
 
 add_library(tomvizlib ${SOURCES} ${accel_srcs})
+target_link_libraries(tomvizlib PRIVATE ${PYTHON_LIBRARIES})
```
cjh1: I think we can use the target here `Python::Python`?

cryos: Missed this, I don't think it is all that important. It should resolve to the same thing. I am just going to leave it, the repetition looks weird in the new namespace stuff...

PR #2081, opened 2020-03-23
Skip re-ordering data for tilt series

psavery:
Previously, tilt series required two deep copies of the data:
one to re-order from C to Fortran, and one to swap the X and
Z axes.

However, it was discovered that we can perform both of these actions
simultaneously by simply re-labeling the dimensions - no deep
copies required, and no modification of data.

Now, for tilt series, reading and writing do not require any deep
copies. Volumes, however, still require one deep copy to convert
from C to Fortran ordering and back.


psavery:
There is also some potential for us to do this on the python side as well. [Here](https://github.com/OpenChemistry/tomviz/blob/63b087ee781c8b5879d7c139741a2361aa19e469/tomviz/python/tomviz/executor.py#L441), a deep copy is always made, since `np.asfortranarray()` returns a deep copy if the array's memory order isn't Fortran.

We could potentially skip both deep copies (converting to Fortran memory and then swapping axes) for tilt series in the python.

PR #2082, opened 2020-03-24
Add ability to contour by specific arrays

psavery:
Rather than only contouring by the active scalars, allow the user
to choose which array to contour by.

Screenshot from 2020-03-24 10-34-07

### DIFF
--- a/tomviz/modules/ModuleContour.cxx
+++ b/tomviz/modules/ModuleContour.cxx
@@ -31,7 +31,9 @@ public:
   bool ColorByArray = false;
   bool UseSolidColor = false;
   QString ColorArrayName;
+  QString ContourArrayName;
   vtkNew<vtkActiveScalarsProducer> ColorArrayProducer;
+  vtkNew<vtkActiveScalarsProducer> ContourArrayProducer;
 };
 
 ModuleContour::ModuleContour(QObject* parentObject) : Module(parentObject)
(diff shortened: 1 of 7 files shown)


alesgenova:
I tested it out and it works well for me. The code looks good too (btw, glad the `vtkActiveScalarsProducer` ended up being useful here too).

From the user facing controls, I wonder if it would be better having something similar to the active scalars that we have in the other modules, where we also have a `Default` option that keeps the active scalars in sync with whichever array is selected on the datasource itself.

So for example, in the contour you really have two active scalars (the contour by, and color by).
By default they would both be set to `Default` (i.e. in sync with the datasource), and the user can lock them to a specific array using the dropdown.

cryos:
Yes, we should retain the default scalar across all visualization modules, i.e. the active scalar should be default/what is active on the data source, and color by should be nothing. Color by doesn't need a default in my opinion - you should actively choose one to use but it defaults to off/not used.

psavery:
@alesgenova is going to get the active scalars settings in the `Module` class to use the name rather than an index. I am waiting for that change before proceeding.

psavery:
This is ready for review. Maybe someone can review it?

psavery:
It looks like the clang-format is [asking for extra indentation](https://dev.azure.com/OpenChemistry/tomviz/_build/results?buildId=261&view=logs&jobId=ea90928a-f2cd-57ff-21a6-274eb7bf65e5&j=ea90928a-f2cd-57ff-21a6-274eb7bf65e5&t=eec4a24d-200b-5b92-e72b-2a5477e40ec4) on lines of code after vtk macros are used, which looks kind of strange.

@cryos, what do you think?

cryos:
That can be ignored, I think this looks OK to merge.

PR #2088, opened 2020-04-03
TEST: Test git clang format

psavery:
It should fail


psavery:
This causes a git-clang-format failure, as intended. Test passed. Closing.

PR #2090, opened 2020-04-07
Save the module active scalars by name in the state file

alesgenova:

### DIFF
--- a/tomviz/DataSource.h
+++ b/tomviz/DataSource.h
@@ -234,6 +234,7 @@ public:
   int activeScalarsIdx() const;
   /// Get the scalars name for a given index.
   QString scalarsName(int arrayIdx) const;
+
   /// Get the scalars idx for a given name.
   int scalarsIdx(const QString& arrayName) const;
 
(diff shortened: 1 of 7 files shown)


psavery:
I haven't tried this out, but I think it looks good to me.

tomviz/modules/Module.h
```diff
@@ -114,7 +114,9 @@ class Module : public QObject
 
   /// Returns the active scalars of the module
   int activeScalars() const { return m_activeScalars; }
-  static const int DEFAULT_SCALARS;
+  QString activeScalarsName() const;
+  static const QString DEFAULT_SCALARS_NAME;
```
cryos, line 118: Can we not use macro style names, I think using s_defaultScalarsName and s_defaultScalarsIdx would be preferable.

tomviz/DataSource.h
```diff
@@ -234,6 +234,8 @@ class DataSource : public QObject
   int activeScalarsIdx() const;
   /// Get the scalars name for a given index.
   QString scalarsName(int arrayIdx) const;
+  /// Get the scalars idx for a given name.
```
cryos, line 237: Leave a blank line before new function/docs

cryos:
Can you just fix the DCO error, then I think we can merge.

PR #2095, opened 2020-04-08
Python wrapping

cryos:


psavery:
In azure-pipelines, it looks like there is a linking issue on Windows.

For Linux and Mac, it looks like there is a segfault in one of the ctests.

tomviz/PipelineProxy.cxx
```diff
@@ -0,0 +1,791 @@
...
+{
+  return new PipelineProxy();
+}
+
+void PipelineProxyFactory::registerWithFactory()
+{
+  PythonFactory::instance().setPipelineProxyFactory(new PipelineProxyFactory);
+}
+
+} // namespace tomviz
```
cjh1, line 791: So this is essentially what `PipelineStateManager` was?

cryos: Yes, but this can be added to the factory at runtime, and avoid needing almost all Tomviz classes. It really is for runtime initialization, and avoiding sucking all of Tomviz in for what is a very simple interface.

cryos:
Merging, got OK from Chris, want to look at packaging issues ASAP.

PR #2106, opened 2020-04-09
Ensure the slice and contour modules updates after first operator

### LINKED ISSUE #2086: Slice module not updating its data after first operator
On the master branch, if you perform two operators in a row, the slice module only appears to update its data after the first operator, but not after the second operator.

For instance, load the tilt series sample, and perform "Bin Tilt Images x2" twice on it. The first time, it updates properly. The second time, only the outline updates:
slice_not_updated

If you delete the slice module and then make it again, it appears to have its data set properly:
slice_updated

alesgenova:
Fixes #2086

### DIFF
--- a/tomviz/modules/ModuleSlice.cxx
+++ b/tomviz/modules/ModuleSlice.cxx
@@ -68,7 +68,7 @@ bool ModuleSlice::initialize(DataSource* data, vtkSMViewProxy* vtkView)
     onTextureInterpolateChanged(m_interpolate);
     pqCoreUtilities::connect(m_widget, vtkCommand::InteractionEvent, this,
                              SLOT(onPlaneChanged()));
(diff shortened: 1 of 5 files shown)


tomviz/modules/ModuleSlice.cxx
```diff
@@ -365,6 +368,12 @@ void ModuleSlice::dataUpdated()
   emit renderNeeded();
 }
 
+void ModuleSlice::pipelineFinished()
+{
+  m_producer->SetOutput(dataSource()->producer()->GetOutputDataObject(0));
```
cryos, line 373: Let's add a FIXME comment here to state that the pipeline connection is broken, and we should investigate a more elegant fix after the release.

cryos: How does this account for multiple channel too - it is always selecting channel 0 isn't it?

PR #2122, opened 2020-08-20
Ensure reordering of the volume in GenericHDF5Format

### LINKED ISSUE #2121: Tomviz (v1.9) can't read non-square volumes stored in h5 datasets
**Describe the bug**

If a tomographic volume stored in an h5 file is not square (ie 225 x 135 x 135) the volume comes out distorted and incorrect

Opening the File in Tomviz v1.9

vs

Opening the same File in Tomviz v1.7
image

**To Reproduce**

I can provide a sample dataset if necessary. The input volume just has to be stored in a h5 file and not square. 

**Expected behavior**

The way how tomviz v1.7 opens h5 files is correct

 - Operating system: macOS 10.15.6 and Windows 10
 - Tomviz version: [e.g. 1.4.0-58]

cryos:
In the case of simple HDF5 files the reorder was not being called, this
can be missed with square volumes. Fixes #2121, thanks to Patrick Avery
for pointing out the lines where this was happenings.


cryos:
Yes, I was just looking at that block. I think this is better and very simple, I will see if I make a non-square tilt series and maybe a follow on PR, or if we want to hold off I can rework this. Probably next week as I am about to head out of the door.

PR #2126, opened 2020-11-10
Customized operator for passive volume monitoring.

jtschwar:
Following up on our meeting yesterday. Here is a custom python operator that monitors a volume from a remote server and downloads it every time the file is modified.

### DIFF
--- a/tomviz/python/Recon_live_tomo.py
+++ b/tomviz/python/Recon_live_tomo.py
@@ -1,0 +1,66 @@
+import tomviz.operators
+import numpy as np
+import paramiko
+import os, time
+import h5py
+
+class liveTomoOperator(tomviz.operators.CancelableOperator):
+	
+	def transform(self,dataset,hostName=None,userName=None,password=None,
+                  port=None,filename=None,remoteDirectory=None,localDirectory=None):
+
+		# localDirectory = '/Users/jonathanschwartz/Desktop/'
+
+		transport = paramiko.Transport((hostName, port))
+		transport.connect(username=userName,password=password)
+		sftpConnection = paramiko.SFTPClient.from_transport(transport)
+		sftpConnection.chdir(remoteDirectory)
+
+		# Create child for recon
+		child = dataset.create_child_dataset()
+		lastMtime = 0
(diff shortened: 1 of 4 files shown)


tomviz/python/Recon_live_tomo.py
```diff
@@ -0,0 +1,69 @@
+import tomviz.operators
+import numpy as np
+import paramiko
+import h5py
+
```
cjh1, line 5: We currently don't package either paramiko or h5py, I thought that these dependencies had been removed? Packaging h5py may not be straight forward given that we already link to the hdf5 c library and would have to make sure we package a compatibly version of h5py.

tomviz/python/Recon_live_tomo.py
```diff
@@ -0,0 +1,69 @@
...
+        # One last update of the child data
+        child.active_scalars = recon
+
+        returnValues = {}
+        returnValues["reconstruction"] = child
+        return returnValues
+
+
+def check_for_new_volume(sftp, filename, localDirectory, lastModifiedTime):
+    import time
```
cjh1, line 54: Any reason for importing inside the function?

tomviz/python/Recon_live_tomo.py
```diff
@@ -0,0 +1,69 @@
...
+                    file = h5py.File(localDirectory + filename, 'r')
+                    if firstLoad:
+                        recon = np.zeros(file['recon'].shape, dtype=np.float32)
+                        firstLoad = False
+                    recon[:, :, :] = file['recon'][:, :, :]
+                    child.active_scalars = recon
+                    self.progress.data = child
+                    file.close()
+                except: 
+                    continue
```
cjh1, line 41: Are you sure you want to ignore the potential exceptions here? If there is a specific exception class that you want to ignore I would add in the type in the except block. So called blind except blocks are generally a bad idea.

cjh1 (commented):
A few comments in line. It also looks like you have merge conflicts so need to merge it master.

cjh1:
You also need to add the sign of by for the DCO

Issue #2125, opened 2020-09-25
Segfault when merging image array components


cryos:
I am using `master` and get a segfault when trying to merge image array components, if I keep them as separate arrays it works as expected. The relevant part of the backtrace looks to be:
```
12      0x559bcf226b00 tomviz::ModuleFactory::createModule(QString const&, tomviz::DataSource*, vtkSMViewProxy*) + 48
11      0x559bcf2436db tomviz::ModuleSlice::initialize(tomviz::DataSource*, vtkSMViewProxy*) + 107
10      0x559bcf240236 tomviz::ModuleSlice::setupWidget(vtkSMViewProxy*) + 550
9       0x559bcf2a3f26 vtkNonOrthoImagePlaneWidget::SetInputConnection(vtkAlgorithmOutput*) + 246
8       0x559bcf2a3741 vtkNonOrthoImagePlaneWidget::SetPlaneOrientation(int) + 97
7       0x7fdfc0f2ea46 vtkDemandDrivenPipeline::UpdateInformation() + 38
6       0x7fdfc0f2e956 vtkDemandDrivenPipeline::UpdateDataObject() + 38
5       0x7fdfc0f2e0e1 vtkDemandDrivenPipeline::UpdatePipelineMTime() + 97
4       0x7fdfc0f2f0c7 vtkDemandDrivenPipeline::ComputePipelineMTime(vtkInformation*, vtkInformationVector**, vtkInformationVector*, int, unsigned long*) + 71
3       0x7fdfc0f2374d vtkAlgorithm::ComputePipelineMTime(vtkInformation*, vtkInformationVector**, vtkInformationVector*, int, unsigned long*) + 13
2       0x559bcf29bbb0 vtkActiveScalarsProducer::GetMTime() + 96
1       0x559bcf29b8de vtkActiveScalarsProducer::ReSync() + 142
0       0x7fdfc07290f0 /usr/lib/libpthread.so.0(+0x140f0) [0x7fdfc07290f0]
(  36.048s) [paraview        ]                       :0     FATL| Signal: SIGSEGV
```
The images I am testing with are pretty small, 61x61x31. Keeping this as a note, I will see if removing the creation of the slice module helps later.

PR #2128, opened 2020-12-01
Fix merge array components crashing

psavery:
This consists of two fixes. The commit messages for each are pasted below:

1. Ensure array is active in DataSource constructor
    
Some of the filters from VTK/ParaView don't always set an active
array. This was the case for using the PythonCalculator to merge
arrays together.
    
To make sure an array is active, check to see if there are arrays
and that no array is active, and if that is the case, make one
active.

2. Get actual range of all components for histogram
    
`arrayPtr->GetFiniteRange(minmax, -1)` was being used before, and this
actually returns the range of the L2 norm over all of the components.
This was resulting in some negative indices being computed in the
histogram computation code, and a subsequent seg fault.
    
Instead, loop over all the components and get the actual range
of all the components. This fixes the segmentation fault.

Fixes: #2125

### DIFF
--- a/tomviz/DataSource.cxx
+++ b/tomviz/DataSource.cxx
@@ -1267,6 +1267,7 @@ void DataSource::init(vtkImageData* data, DataSourceType dataType,
   if (data) {
     auto tp = vtkTrivialProducer::SafeDownCast(source->GetClientSideObject());
     tp->SetOutput(data);
+    ensureActiveArray();
   }
 
   // Initialize maps to track array renames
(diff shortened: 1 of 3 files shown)


cjh1:
@psavery Looks good, I am fixing up the flake8 CI in #2129

cryos:
This is better as it doesn't crash, but I can't seem to visualize the data in any useful way when loading the three channels. Most modules are disabled
image
Thanks for fixing, we should think about what we can do to visualize RGB volumes using the volume renderer and slice modules.

cryos (approved):

PR #2136, opened 2020-12-02
Specify python executable to use in azure pipelines

psavery:
This forces cmake to use the one that was specified. 

Previously, which python it used was hard to predict, and it was choosing the 
wrong python on Mac and Windows. This was causing some failed tests.

Sometimes, it was even mismatching different executables and libraries for the same build.

To avoid these, force cmake to use the python that was specified.


cjh1 (approved):

cjh1:
:fireworks:

PR #2137, opened 2020-12-03
Unpin numpy version for tomviz-pipeline package

cjh1:
We shouldn't need to pin the version and it was causing a build error in one of our docker images.


psavery (approved):
LGTM

PR #2139, opened 2020-12-09
For azure pipelines, skip running docker on mac

psavery:
Installing docker on Mac in azure-pipelines really only worked 80-90%
of the time, and would cause occasional job failures, both during the
docker installation, and while running docker tests.

Skip installing docker and running docker tests on Mac to reduce the
number of job failures. Hopefully, docker succeeding on Linux is good enough.


cjh1 (approved):
LGTM, thank you

PR #2140, opened 2020-12-10
Map 3-component data to RGBA values for volumes

psavery:
There are some things here that can be improved, but this is a basic
working example.

This takes the 3 components and rescales their range to be 0 - 1 (so
the full range of RGB is used). Then, it places the norm of the original
3-components into the fourth component (which gets passed through the
opacity functions to determine opacity). This allows utilizing the
opacity editor to modify the opacity of the image.

Here is the example we are currently working with:

real_example

There is a little yellow on the corners, but not a lot of color in the middle.
I am guessing that this is because the data in the middle have similar
values in their components, which would result in gray/white colors.

Here's a fake example where the components at different parts of the image are
very different from each other, resulting in lots of color. It's a linear scale of
0 - 1 for the red component, 0 for the green component, and linear scale of 
1 to 0 for the blue component.

fake_example

So it seems that it is working.

The histogram appears to be displaying the norm of the components of
the data. And the color editor won't work for this mode, since the colors of the
image are using the data directly for RGB.

Internally, we are using vtkVolumeProperty's IndependentComponents, which
requires 2 or 4-component data. For 4-component data, the first three
components are RGB, and the fourth component is passed through the opacity
functions for opacity.

An option hasn't been added yet to turn this mapping off, so currently, all
3-component data maps to RGBA values for volumes.

Also, this only works for float/double data. Needs to be generalized more
for integers.

One possible improvement is that a filter might be better instead of an
internal object for mapping 3-component data to 4-component RGBA data.

It might also be good to make sure the opacity editor is working properly...
it seems to me the opacities of this are off.

Do we also want the ability to visualize individual components (using our
regular rendering mode), and the magnitude (norm) of all the components?

@cryos

### DIFF
--- a/tomviz/modules/ModuleFactory.cxx
+++ b/tomviz/modules/ModuleFactory.cxx
@@ -54,8 +54,7 @@ bool ModuleFactory::moduleApplicable(const QString& moduleName,
 
   if (dataSource && view) {
     if (dataSource->getNumberOfComponents() > 1) {
-      if (moduleName == "Contour" || moduleName == "Volume" ||
-          moduleName == "Threshold") {
+      if (moduleName == "Contour" || moduleName == "Threshold") {
         return false;
       }
     }
--- a/tomviz/modules/ModuleVolume.cxx
+++ b/tomviz/modules/ModuleVolume.cxx
@@ -36,6 +36,8 @@
 
 namespace tomviz {
 
+static void computeRange(vtkDataArray* array, double range[2]);
+
 ModuleVolume::ModuleVolume(QObject* parentObject) : Module(parentObject)
 {
   connect(&HistogramManager::instance(), &HistogramManager::histogram2DReady,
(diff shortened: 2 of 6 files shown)


psavery:
Looks like if we add an ability to edit the range that gets colored, that you can get nicer colorings:

image

cryos:
Thanks for looking at this @psavery, I think the ability to edit the range is probably important. I think what you have shown is a huge improvement in general, and this looks like a good direction to push. The RGB trick is a vast improvement over no visualization, it is still limited to three components but I think that can help with quite a bit of data. There is also the question of how we make loading data like this easier, but in general I think pushing down the path you have shown is desirable. If it is RGB data I think we have chosen to render in this way, using an internal filter to the module seems like the right way to do. I will try and build this locally, we could do with more examples of data to test this out - usually we are looking at a technique with elemental selectivity - e.g. red is nickel, blue is cobalt, etc.

psavery:
@cryos Thank you for the feedback! I just pushed a WIP for the range editor. That commit involves some hard-coding (and the ranges editor might not be that pretty), so it will probably have some changes soon.

psavery:
The ranges editor now uses sliders (see gif below), and is only visible when we are doing RGBA component mapping.

Note that these changes have only been tested on float data. I'm not sure what will happen if we try to use another kind of data...

Ranges_edito

psavery:
The latest updates make it so that volumes for multi-component data display the norm of the data by default. And, for 3-component data, there is now a checkbox in the volume panel to indicate whether the data should be displayed with RGBA mapping.

cjh1 (approved):
LGTM

PR #2141, opened 2020-12-17
Include vtk_hdf5.h rather than hdf5.h

cjh1:
This is to ensure we pick up the right header, otherwise we may pick up a system version instead.


psavery (approved):
LGTM

PR #2143, opened 2021-01-18
Use upstream jsonpatch

cjh1:
We have added code to the interactive console support to allow us to use the upstream verson of jsonpatch.


psavery (approved):
LGTM

PR #2144, opened 2021-01-18
Add "solidity" volume module property

cjh1:
https://user-images.githubusercontent.com/1031053/104940004-24ea8500-597f-11eb-9f4a-9c903344c2cc.mp4

This property is 1/ScalarOpacityUnitDistance controlling how "solid" the volume appears.

We may need to change the slider so its non-linear, you get the most change at the lower end.


cjh1:
@psavery Do you thing you could review this? I just fixed the conflicts.

psavery (approved):
LGTM

PR #2145, opened 2021-01-20
Add image viewer mode to the view menu

psavery:
https://user-images.githubusercontent.com/9558430/105114271-25247680-5a8c-11eb-8aa6-4a86e295e323.mp4

This sets up a slice module, moves the camera, and adds a slider widget
below the central widget so that the user can slide through the 
images/slices.

When checked, these things are done:

1. Sets the projection to orthographic
2. Sets the render interactor mode to 2D
3. Finds the first slice module or creates one. Hides all other modules.
4. Align the camera so the single slice is filling the screen
5. Show the slider widget below the central widget

This assumes that the images are in the Z-direction as well.

When the user unchecks the image viewer mode, the projection type,
render interactor, modules, and most of the view are restored to
the settings they had before.

It currently doesn't hide the histogram widget automatically. You can shrink it to make the image bigger.
But that's possibly something we can add too if it is desired.

### DIFF
--- a/tomviz/CentralWidget.cxx
+++ b/tomviz/CentralWidget.cxx
@@ -18,6 +18,7 @@
 #include <pqSettings.h>
 #include <vtkPVDiscretizableColorTransferFunction.h>
 #include <vtkSMPropertyHelper.h>
+#include <vtkSMRenderViewProxy.h>
 #include <vtkSMViewProxy.h>
 
 #include <QModelIndex>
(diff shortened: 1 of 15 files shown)


cjh1:
@psavery This is looking great. I think there is one improvement we can make, which can be in a follow up PR. It would nice if there was an option in the load image stack dialog to load the data in this mode.

cjh1 (approved):
LGTM

psavery:
@cjh1 Sounds good, thank you!

cjh1:
@psavery We should also document this feature in tomvizdocs

PR #2146, opened 2021-01-22
Add auto contrast button similar to that of ImageJ

psavery:
This button performs the same kind of logic that ImageJ does when
an auto contrast is requested. That is:

1. Modify the LUT so that the end points are within a certain threshold
   of the data.
2. Preserve state so if the user clicks the button multiple times,
   the threshold keeps increasing until it resets.

However, unlike ImageJ, we are doing an auto contrast for the whole
3D dataset. ImageJ does the auto contrast for a single slice, or a
selected region. We do not support these yet.

This also adds placeholder nodes for both the colorbar editor and the
opacity editor mostly to make it look nicer. When the range of the
LUT/opacity editor is scaled down, the placeholder nodes are added
at the ends so that the user can still modify the color/opacity of
those regions. Unless the placeholder nodes (or their nearest neighbor)
are modified, they will be removed before performing operations such
as rescaling or inversion, and then re-added afterwards.

This also fixes a bug with the color map editor where the LUTProxy
would not stay up to date with changes to the LUT, which would cause
the LUTProxy to over-write changes to the LUT when it pushed an
update (so, for instance, if a user modified points on the colorbar,
and then performed an inverse/reset range, their modifications would
get over-written). Now, similar to the opacity editor, the proxy gets
updated when changes are made to the vtkObject directly.

slice_auto_contrast

3d_auto_contrast


cjh1 (approved):
This looks really good, thank you!

PR #2147, opened 2021-01-25
Add color map live updates and a settings dialog

psavery:
Live updates are now performed for deleting and moving control
points in the color map editor. This also adds a color map settings
editor where the color space can be changed.

https://user-images.githubusercontent.com/9558430/105761840-3997c180-5f19-11eb-9bcd-a6b32615e79e.mp4

Commit messages are pasted below.

Re-build LUT every time it is modified
    
This gives us live updates for edits to the color bar. So now, things
like deleting or moving a control point will give live updates, where
they did not give live updates before (and you had to do things like
creating a new control point or inverting the color map to get it to
update).
    
Strangely, re-building the LUT is also considered a modify event, so
we have to block the signal to avoid infinite recursion.

Add color map settings widget
    
This is currently only for editing the color space of the color map.
But it may be used for other things in the future as well, such as
setting the NaN color.

### DIFF
--- a/tomviz/CMakeLists.txt
+++ b/tomviz/CMakeLists.txt
@@ -38,6 +38,8 @@ set(SOURCES
   CloneDataReaction.h
   ColorMap.cxx
   ColorMap.h
+  ColorMapSettingsWidget.cxx
+  ColorMapSettingsWidget.h
   ConvertToFloatReaction.cxx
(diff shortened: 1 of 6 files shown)


cjh1 (approved):
LGTM, nice improvement.

PR #2150, opened 2021-01-28
Add image viewer mode option to image stack dialog

psavery:
This adds a checkbox to the image stack dialog, which indicates
whether or not to put the view in image viewer mode after the
data finishes loading.

image


cjh1 (approved):
LGTM

PR #2164, opened 2021-04-19
Change path of homebrew Qt from qt to qt@5

psavery:
Apparently, homebrew changed the path of qt from

```
/usr/local/opt/qt
```

to

```
/usr/local/opt/qt@5
```

Update our Qt paths to reflect this.


alesgenova (approved):

psavery:
Looks like the Mac tests are not fully fixed, but this fixes at least one issue. I'll go ahead and merge.

PR #2165, opened 2021-04-19
Azure pipelines: build tomviz from paraview install tree

psavery:
This is done because we cannot cache the paraview build tree on Linux because it is getting to be too big, and the VM runs out of disk space when it tries to tar it.

The install tree uses much less disk space.


alesgenova (approved):

psavery:
@alesgenova Looks like there is an error now when we build tomviz on Windows from the cached paraview install tree... :slightly_frowning_face:

PR #2166, opened 2021-04-20
Link to gtest imported targets

psavery:
Newer versions of CMake have renamed the GTest result variables,
and this is causing them to not be linked on the Mac.

Link to the gtest imported targets instead...


psavery:
@alesgenova this fixes the mac build.

alesgenova (approved):

PR #2170, opened 2021-05-06
Add basic FXI workflow widget

psavery:
This allows users to perform and visualize test rotations in the widget, in order to choose a rotation for reconstruction.

An internal infrastructure change that made this possible is that custom python widgets now have access to the up-to-date
python script in the operator dialog, so the custom python widgets are now able to run code in the script. In this case,
the "test_rotations()" function is found in the script and ran in order to generate the test rotations. The user is able to
modify the script if they choose for generating the test rotations.

fxi_workflow2

### DIFF
--- a/tomviz/Behaviors.cxx
+++ b/tomviz/Behaviors.cxx
@@ -5,6 +5,7 @@
 
 #include "ActiveObjects.h"
 #include "AddRenderViewContextMenuBehavior.h"
+#include "FxiWorkflowWidget.h"
 #include "MoveActiveObject.h"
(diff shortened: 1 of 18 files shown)


psavery:
I might say we can let the `flake8` errors go - because they are mostly coming from code I copy and pasted from those who provided the fxi workflow code.

alesgenova (approved):

alesgenova:
https://user-images.githubusercontent.com/15913822/121752137-843b6400-cadd-11eb-8ecd-a41d6bf73851.mp4

PR #2181, opened 2021-07-19
Allow custom parameters in FXI workflow operator

psavery:
This commit adds the ability for the user to add custom parameters
to both the operator UI and the main `transform()` function via the
`description.json` file for the FXI workflow operator.

https://user-images.githubusercontent.com/9558430/126194631-ba33b36a-eff8-4a25-813e-f9d4dd0c8a6a.mp4

There is now a general `setupUI()` function on the
`CustomPythonOperatorWidget` that matches the signature of `setupUI()`
in the `OperatorWidget` class. The base class version does not do
anything, but the FxiWorkflowWidget will use it to check for extra
parameters and use the `InterfaceBuilder` class to add them to a
dedicated QLayout. These additional parameters are utilized in
`getValues()` and `setValues()`.

An additional `updateWidgetValues()` function was added to the
`InterfaceBuilder` that takes a `QObject` similar to `parameterValues()`,
which updates the widget values to match the values in m_parameterValues
(which can be set via `InterfaceBuilder::setParameterValues()`).

Some of the functionality in `FxiWorkflowWidget::setupUI()` may be better
off transfered to `CustomPythonOperatorWidget::setupUI()` to make it
easier to do this kind of thing in other operators in the future. Currently,
though, it is all in `FxiWorfklowWidget`.

### DIFF
--- a/tomviz/FxiWorkflowWidget.cxx
+++ b/tomviz/FxiWorkflowWidget.cxx
@@ -7,7 +7,9 @@
 #include "ActiveObjects.h"
 #include "ColorMap.h"
 #include "DataSource.h"
+#include "InterfaceBuilder.h"
 #include "InternalPythonHelper.h"
+#include "OperatorPython.h"
 #include "PresetDialog.h"
 #include "Utilities.h"
 
@@ -125,11 +127,13 @@ public:
   InternalPythonHelper pythonHelper;
   QPointer<FxiWorkflowWidget> parent;
   QPointer<DataSource> dataSource;
+  QPointer<InterfaceBuilder> interfaceBuilder;
   int sliceNumber = 0;
   QScopedPointer<InternalProgressDialog> progressDialog;
   QFutureWatcher<void> futureWatcher;
   bool testRotationsSuccess = false;
   QString testRotationsErrorMessage;
+  QStringList additionalParameterNames;
 
   Internal(Operator* o, vtkSmartPointer<vtkImageData> img, FxiWorkflowWidget* p)
     : op(o), image(img)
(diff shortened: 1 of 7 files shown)


tomviz/FxiWorkflowWidget.cxx
```diff
@@ -209,6 +216,98 @@ vtkStandardNewMacro(InteractorStyle)
...
+    QJsonObject root = json.object();
+
+    // Get the parameters for the operator
+    QJsonValueRef parametersNode = root["parameters"];
+    if (parametersNode.isUndefined() || !parametersNode.isArray()) {
+      return;
+    }
+    auto parameters = parametersNode.toArray();
+
+    // Here is the list of parameters for which we already have widgets
```
alesgenova, line 251: Why not let the method automatically generate the UI for all the possible parameters?
What would happen if the user customizes the operator and removes one of the expected parameters?

alesgenova (approved):

tomviz/FxiWorkflowWidget.cxx
psavery: Well, having expected parameters allows us to do more things to help out the user. For instance, the "Rotation Center" parameter is automatically set to match whatever test rotation center the user selects. And the "Slice Stop" parameter is automatically set to the max slice value, and it is not allowed to go above that. These widgets are currently defined in the UI file.

For the Rotate Align Widget, the user also gets visual feedback from the parameters they choose. And there is opportunity for more complex parameter selection designs if there are expected parameters.

So I think there is a place for expected parameters. But we could potentially redesign how they are done. Maybe make them invisible and not used if the user removes them from the description file?

For now, though, the user shouldn't remove the expected parameters from the description.json file. I don't know what will happen if they do. I guess we'd have to design for that use case, maybe in a separate PR.

PR #2182, opened 2021-07-20
Add "Conform Volume" to pipeline view context menu

psavery:
This adds a new action called "Conform Volume" to the pipeline view
right-click context menu that is only visible if exactly two volumes
are selected, and they do not have the same name.

If the user triggers the "Conform Volume" action, a dialog will appear
explaining what will be done, and asking the user to select which volume
should be conforming (and, consequently based upon their choice, which
volume will be conformed to).

If the user presses "OK", a new DataSource will be created based upon the
conforming volume that will have dimensions, extent, spacing, origin, and
display position matching that of the "conform to" volume. This is done
by first resizing the conforming volume via `vtkImageResize` so that the
dimensions of the first image will match the second, and then
`vtkImageChangeInformation` to change the extents, spacing, and origin
to match the "conform to" volume. The display position is also updated.

https://user-images.githubusercontent.com/9558430/126393509-6889c476-8af7-4601-9fce-ce608ebfaafa.mp4

### DIFF
--- a/tomviz/CMakeLists.txt
+++ b/tomviz/CMakeLists.txt
@@ -44,6 +44,10 @@ set(SOURCES
   ColorMapSettingsWidget.h
   ComboTextEditor.cxx
   ComboTextEditor.h
+  ConformVolumeDialog.cxx
+  ConformVolumeDialog.h
+  ConformVolumeReaction.cxx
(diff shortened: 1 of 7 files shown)


alesgenova (approved):
Looks great!

PR #2186, opened 2021-08-18
 Real-Time-Tomography. 

jtschwar:
Current draft for real-time tomography features. 

Unfortunately, I'm still having trouble accessing eigen from VTK, I've tried a few ways such as including its location in the Paraview-VTK directory and trying to include a few module definitions listed in vtk_eigen.h from that directory.

Signed-off-by: Jonathan Schwartz <jtschw@umich.edu>

### DIFF
--- a/tomviz/CMakeLists.txt
+++ b/tomviz/CMakeLists.txt
@@ -3,7 +3,8 @@ find_package(OpenGL REQUIRED)
 include_directories(SYSTEM
   ${Qt5Widgets_INCLUDE_DIRS}
   ${Qt5Concurrent_INCLUDE_DIRS}
-  ${PROJECT_SOURCE_DIR}/thirdparty/pybind11/include)
+  ${PROJECT_SOURCE_DIR}/thirdparty/pybind11/include
+  ${ParaView_DIR}/VTK/ThirdParty/eigen)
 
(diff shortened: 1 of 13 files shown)


cjh1:
@jtschwar  Why was `thirdparty/pybind11` changed? We should stick with 2.5.0 unless there is a good reason to push things forward, in which case we need to resolve the build issues that would cause.

jtschwar:
@cjh1 that was accidental. I'll update this PR to exclude that change.
 
Although, I'm having one final issue where I can't import a final dependency (<pybind11/eigen.h>) which lets me transfer data between python to c++ because the vtkEigen wrapper redefines the module as <vtkeigen/Eigen/x> but pybind11 is wanted to import the native library (<Eigen/x>). 

My three solutions are as following: 1) update pybind11 so that it imports the vtk Version of Eigen (I'm assuming that's too complicated) 2) set the Cmake Variable - VTK_MODULE_USE_EXTERNAL_vtkeigen = True when compiling Paraview so that I can use the native library call (<Eigen/x>) (might also be an unlikely solution?) 3) Use vtk to pass data between the C++ and Python, which I'm not too familiar with so I'd appreciate some example scripts to look at. 

Currently, my code implementation is using the second solution

PR #2212, opened 2021-10-28
Add time series steps to DataSource and loading

psavery:
With these changes, if the user opens multiple files, a dialog will
ask the user if they wish to open up the files as a time series. If
they select yes, all of the files will be loaded, but only the data
source from the first file will be kept, and the image data from the 
other files will be added as time series steps to the data source.
The files are currently sorted when they are loaded using a regular
string sort, so the first one in this order will be kept, and the 
others will be added in order as time series steps.

This also sets up the animation to have a number of time steps equal
to the number of time steps the user loaded, and it adds a connection
in the DataSource to detect when the time changes. This allows the user to
use the VCR toolbar as well as the animation widget to control the time
step of the current data source.

The time series steps are currently not supported in state files, and 
any operators will operate on the current image data, and the output
data source will not have any time series steps.

An example can be seen below:

https://user-images.githubusercontent.com/9558430/139334179-8bd4de67-1206-4e2b-80f4-c5238b6a6f55.mp4

### DIFF
--- a/tomviz/ActiveObjects.cxx
+++ b/tomviz/ActiveObjects.cxx
@@ -297,6 +297,12 @@ DataSource* ActiveObjects::activeParentDataSource()
   return m_activeParentDataSource;
 }
 
+pqTimeKeeper* ActiveObjects::activeTimeKeeper() const
+{
+  pqServer* server = pqActiveObjects::instance().activeServer();
+  return server ? server->getTimeKeeper() : nullptr;
+}
+
 Pipeline* ActiveObjects::activePipeline() const
 {
 
--- a/tomviz/ActiveObjects.h
+++ b/tomviz/ActiveObjects.h
@@ -125,9 +125,13 @@ public slots:
   void enableRotation(bool b);
   void enableScaling(bool b);
 
+  void setShowTimeSeriesLabel(bool b);
+
   bool translationEnabled() const { return m_translationEnabled; }
   bool rotationEnabled() const { return m_rotationEnabled; }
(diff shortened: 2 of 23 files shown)


psavery:
Left to do:
- [ ] Add a vtkTextWidget that displays the current time series step label in the render window
- [ ] Add a section in the Data Properties panel to edit the labels, label visibility (in the render view), and order of time series steps

cjh1 (approved):
LGTM

RupertPaoZ:
> 

hello， I am trying to use the 'time series' feature. However I can't find the button of 'time series'. Exactly, there is no pop-ups while opening data. The version I am using is 1.10 (stable).

psavery:
There appears to be some issue with Windows updating the nightly. But the binary is being built.

@RupertPaoZ You can find the latest Windows binaries [here](https://github.com/OpenChemistry/tomviz-superbuild/actions/runs/3851609227) (scroll to the bottom of the page).

psavery:
Looks like [this](https://github.com/softprops/action-gh-release/issues/280) is the issue. I can try reverting the action-gh-release version to fix it.

cjh1:
@psavery Lets try that

RupertPaoZ:
> There appears to be some issue with Windows updating the nightly. But the binary is being built.
> 
> @RupertPaoZ You can find the latest Windows binaries [here](https://github.com/OpenChemistry/tomviz-superbuild/actions/runs/3851609227) (scroll to the bottom of the page).

Thanks a lot!

PR #2216, opened 2021-12-07
Add manual manipulation operator

psavery:
This PR first adds orientation as a property of the data source, and allows the user to manually manipulate the orientation.

Next, this PR adds a manual manipulation operator, which allows the user to manually manipulate the position, orientation, and scale of the data source. The operator then applies the visual transformations to the voxels. This can be used for manual volume alignment (which can sometimes be helpful to do before performing automatic volume alignment, for instance).

https://user-images.githubusercontent.com/9558430/145095679-fb526065-e0c5-410f-9071-e7e3d82561e2.mp4

This PR also re-works some of the data properties panel to allow multiple different types of visual interactions and to make the transformation widgets more concise.


tomviz/python/ManualManipulation.py
```diff
@@ -0,0 +1,84 @@
...
+
+    if rotation is None or all(np.isclose(x, 0) for x in rotation):
+        # No rotation. Nothing to do.
+        return
+
+    if not array.dtype == np.float32:
+        raise Exception('Input array must be float32')
+
+    # Wait to import ITK until we know we have to
+    import itk
```
cjh1, line 23: Will our packaged version of ITK do here?

psavery: Yes it should! In fact, I just tested it in tomviz 1.9.0 and it worked.

psavery:
This is getting to be in a pretty good state. You can see a combination of all manual manipulation transforms being used in [this video](https://drive.google.com/file/d/1Jij4VkFKfveWeyApQuF_aolcVFW8BDs6/view?usp=sharing). That includes rotation, translation, scaling, padding/cropping, and resampling (where padding/cropping and resampling are done to make the voxels align with the reference dataset if one is selected).

cjh1 (approved):
LGTM, thank you, this is looking really good!

PR #2217, opened 2021-12-07
Allow enumeration data types other than int

psavery:
Previously, the enumeration data type was required to be an int.
However, there are some cases where it is better for it to be
something else, such as a string. It is very helpful for integrating
the ITK Elastix library for it to be a string, since that is what
we will actually be setting in the ITK Elastix parameters.

Thus, try to convert the data to an int if possible (otherwise, it
will always be interpreted as a float, which will cause issues with
some operators). If it's not an int, just let it be whatever type
it is (float, string, bool, None).


psavery:
This commit is also in #2218. I'm gonna close this in favor of that PR.

PR #2218, opened 2021-12-07
Add ITK Elastix registration operator

psavery:
This adds a volume registration operator that uses ITK Elastix.

To use it, ITK Elastix must be installed via `pip install itk-elastix`.
This was designed to work in both pip and conda environments. We will
look into adding this to the tomviz conda build.

https://user-images.githubusercontent.com/9558430/145103030-2461bf8a-2b5b-48ee-b721-67271e628c50.mp4

### DIFF
--- a/tomviz/CMakeLists.txt
+++ b/tomviz/CMakeLists.txt
@@ -404,6 +404,7 @@ set(python_files
   BinaryOpen.py
   BinaryClose.py
   DummyMolecule.py
+  ElastixRegistration.py
   LabelObjectAttributes.py
   LabelObjectPrincipalAxes.py
   LabelObjectDistanceFromPrincipalAxis.py
@@ -462,7 +463,7 @@ set(python_files
   TV_Filter.py
   PoreSizeDistribution.py
   Tortuosity.py
-  Recon_real_time_tomography.py 
+  Recon_real_time_tomography.py
   )
 
 set(json_files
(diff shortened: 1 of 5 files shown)


psavery:
> @psavery Is itk-elastix enable in the version of ITK in our super build?

No, I think it is separate from ITK, and that you can't enable it in ITK. We might be able to clone [the repository](https://github.com/InsightSoftwareConsortium/ITKElastix) and build it, but I'm not sure. I know that we can, however, `pip install itk-elastix`, which works in conda environments as well.

cjh1:
We may want todo what we did for [databroker](https://github.com/OpenChemistry/tomviz/blob/master/tomviz/MainWindow.cxx#L577-L589) and only enable the operator if we can import `itk-elastix`. Not sure how hard this would be todo for an operator.

psavery:
So currently, the action lives inside the `AddPythonTransformReaction` [here](https://github.com/OpenChemistry/tomviz/pull/2218/files#diff-89654cfe8fac22265fb8d00bbe9ebdf09e9efebcc9105ea6c4d37a4508a2626dR117) similar to what the other actions/reactions are doing.

Should we implement something so that the `AddPythonTransformReaction` can figure out whether to enable/disable the action? I have thought of two options:

1. Try to import the script in python and if any errors occur, disable the action. This means at the file level, some logic could be present to determine whether the right dependencies (and right versions) are available, and if not, raise an exception.
2. Have a list of required dependencies in the `description.json` file. Maybe try to import those modules in python, and if it fails, disable the module. Although things like version requirements might would be nice too.

cjh1 (approved):
LGTM, I think for now we can punt on disabling if elastix is not available. We can look at one of the cleaner solutions in the future.

PR #2223, opened 2022-02-03
Use cmake_path to compute path to python libraries

psavery:
Rather than assuming that the path from the exe dir to the python libraries is
"exe_dir + "/../" TOMVIZ_PYTHON_INSTALL_DIR", compute the relative path
in cmake and use that instead.

When the exe dir was "Library/bin", and the library dir was
"Library/lib", the previous path that was computed was
"../Library/lib/...", which should have been "../lib/..." instead. This
computation fixes the issue.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -63,6 +63,13 @@ if(APPLE)
   endif()
 endif()
 
+# Get the relative path from the tomviz executable install location
+# (INSTALL_RUNTIME_DIR) to the python install dir (tomviz_python_install_dir).
+# Tomviz will use this when it starts to find the python packages.
+cmake_path(RELATIVE_PATH tomviz_python_install_dir
+           BASE_DIRECTORY "${INSTALL_RUNTIME_DIR}"
+           OUTPUT_VARIABLE tomviz_path_from_exe_to_python_install_dir)
+
 # Add our CMake modules to the search path.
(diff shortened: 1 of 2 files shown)


cjh1 (approved):

mathstuf:
Note that this code uses `cmake_path` which requires CMake 3.20. Should the minimum be bumped or should the logic be changed? If the minimum needs bumped, the buildbots will need updated to use a newer CMake.

CMakeLists.txt
```diff
@@ -63,6 +63,13 @@ if(APPLE)
   endif()
 endif()
 
+# Get the relative path from the tomviz executable install location
+# (INSTALL_RUNTIME_DIR) to the python install dir (tomviz_python_install_dir).
+# Tomviz will use this when it starts to find the python packages.
+cmake_path(RELATIVE_PATH tomviz_python_install_dir
+           BASE_DIRECTORY "${INSTALL_RUNTIME_DIR}"
+           OUTPUT_VARIABLE tomviz_path_from_exe_to_python_install_dir)
```
mathstuf, line 71: `file(RELATIVE_PATH)` is another way to do this that doesn't require CMake 3.20.

cjh1: @psavery Can you fix this up so we don't need 2.30?

PR #2224, opened 2022-02-14
Add animation helper dialog with camera orbit

psavery:
The animation helper dialog adds some convenient tools for creating
animations. The first tool that is present is the ability to create
a camera orbit animation. This uses the current camera settings: the
animation starts at the current camera position, and rotates around
the focal point to the right.

Other animation helper tools will be added soon as well...

### DIFF
--- a/tomviz/ActiveObjects.h
+++ b/tomviz/ActiveObjects.h
@@ -125,12 +125,18 @@ public slots:
   void enableRotation(bool b);
   void enableScaling(bool b);
 
+  /// Set whether to enable time series animations.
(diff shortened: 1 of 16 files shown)


ercius:
animations are getting some love and attention!

cjh1:
@psavery Looks like we have a build issue?

cjh1 (approved):

psavery:
I'm adding this video here as an example

https://user-images.githubusercontent.com/9558430/214188389-61411d4c-8a53-4a3f-b4bf-f8c79d29a258.mp4

PR #2228, opened 2022-03-07
First pass at migrating the code to use tiled

cryos:
This uses tiled, pretty much replicates all core with a hack to show the
last twenty results. Still no angles....

### DIFF
--- a/tomviz/python/tomviz/io/_databroker.py
+++ b/tomviz/python/tomviz/io/_databroker.py
@@ -91,8 +91,7 @@ def variables(catalog_name, run_uid, table):
 
 
 def load_variable(catalog_name, run_uid, table, variable):
-    dataset = c[catalog_name][run_uid][table][variable]
-    data = dataset[variable].data
+    data = c[catalog_name][run_uid][table]['data'][variable].data
     shape = data.shape
     data = data.reshape((shape[0]*shape[1], shape[2], shape[3]))
 
--- a/tomviz/python/tomviz/io/_databroker.py
+++ b/tomviz/python/tomviz/io/_databroker.py
@@ -88,12 +88,19 @@ def variables(catalog_name, run_uid, table):
 
 
 def load_variable(catalog_name, run_uid, table, variable):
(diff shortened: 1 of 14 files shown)


cjh1:
Thanks @cryos I will give this a try tomorrow.

cryos:
Working with Dan on getting the angles now, this works but is has a ways to go.

tomviz/python/tomviz/io/_databroker.py
```diff
@@ -1,12 +1,16 @@
...
 from vtk import vtkImageData
 
 _installed = False
+DEFAULT_URL = "https://tiled.nsls2.bnl.gov"
+TILED_URL = os.getenv("TILED_URL", DEFAULT_URL)
 try:
-    from databroker import catalog
+    from tiled.client import from_uri
+    c = from_uri(TILED_URL, "dask")
     from databroker.queries import TimeRange
```
cjh1, line 14: Can we get rid of this databroker import?

cjh1:
@cryos This is ready for you to take a look at.

tomviz/python/tomviz/io/_databroker.py
```diff
@@ -1,12 +1,15 @@
+import os
 import numpy as np
-
 import tomviz.utils
 
 from vtk import vtkImageData
 
 _installed = False
+DEFAULT_URL = "https://tiled.nsls2.bnl.gov"
+TILED_URL = os.getenv("TILED_URL", DEFAULT_URL)
```
cryos, line 9: I wonder if we might make beamline settable with env variables like this, but in the C++ is certainly more than enough to work on this tomorrow as we also tie to the andor image

cryos (commented):
I think this looks good, I haven't had a chance to build and test myself. I can do that this afternoon, but given the time merging now, building, and letting me test in the full Conda environment later is likely best. Thanks for jumping on this Chris!

tomviz/python/tomviz/io/_databroker.py
cjh1: Sure, I can do that

PR #2229, opened 2022-03-08
First pass at PyXRF workflow

psavery:
This first pass includes most of the essential steps, including making
the HDF5 files from scan numbers, creating the csv log file,
processing the projections, creating a single HDF5 file at the end,
and extracting the output from that HDF5 file for Tomviz to use.

The various dialogs allow users to choose options, including selecting
which scans to use.

A video example is provided below.

https://user-images.githubusercontent.com/9558430/157328127-51e7e137-2ab6-4e51-b708-12189742b8cd.mp4


cjh1 (approved):
LGTM

PR #2230, opened 2022-03-09
FXI Workflow Improvements

psavery:
This sets the default test rotations start and stop about the predicted
center, and it also sets the default slice number to be the middle of
this range.

More improvements will be added to this PR as well...


psavery:
A circle mask operator using tomopy was added to this PR as well.

https://user-images.githubusercontent.com/9558430/157544988-136f2867-f416-4884-a387-94b7e5ada21e.mp4

cjh1 (approved):
LGTM

PR #2232, opened 2022-03-11
Fxi workflow improvements

psavery:
1. Change default delta from 20 to 40 for start/stop around the center
2. Change the default color map to grayscale (and separate out the color map from the data source)
3. Add general parameters to the FXI workflow widget, including denoise_flag and dark_scale
4. Made `file_name` and `metadata` accessible in python
5. Read scan id and store it as metadata in the `FxiFormat` reader


cjh1 (approved):
LGTM, thank you!

cjh1:
Do we have a build error?

PR #2235, opened 2022-03-17
PyXRF Workflow Fixes

psavery:
1. Remove a tty check from the progress library
    
This doesn't work in tomviz due to the fact that stdout and stderr
have been replaced by vtk helpers. Skip the check to avoid an error
while `process_proj()` is running.

2. Fix multiprocessing in process_projections

Because we are using an embedded tomviz, the multiprocessing module
is unable to find the python executable, nor is it able to set up the
python paths correctly. Do these things so that multiprocess can
function correctly.

3. Add some error output for starting pyxrf

This way, if PyXRF fails to start/run, we can hopefully see what went
wrong.
    
Additionally, this allows the user to specify the full path to the pyxrf
executable via the environment variable TOMVIZ_PYXRF_EXECUTABLE.

4. Fix logic for final copy of csv file

We started using absolute paths in tomviz, so we need to add in some
logic to handle the copy paths correctly in case they are absolute.

### DIFF
--- a/tomviz/PyXRFProcessDialog.cxx
+++ b/tomviz/PyXRFProcessDialog.cxx
@@ -18,6 +18,7 @@
 #include <QMessageBox>
 #include <QPointer>
 #include <QProcess>
+#include <QProcessEnvironment>
 #include <QTextStream>
 
 namespace tomviz {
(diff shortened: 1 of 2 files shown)


cjh1 (approved):
LGTM

PR #2237, opened 2022-03-24
Databroker/Tiled optimization: use a small in-memory cache.

danielballan:
attn @cjh1 

As written, code like this

https://github.com/OpenChemistry/tomviz/blob/e175aba637572d711caf15949a2040a28dbe6a05/tomviz/python/tomviz/io/_databroker.py#L90-L97

pays a high latency cost to round-trip from the server on every `[]` lookup. By configuring the client to use a small (max 1 MB) in-memory cache, we can turn those network requests into fast dict lookups. The cache operates like a standard web browser cache, respecting standard Cache-Control headers and expiry rules.

Because this is set to a small total capacity, it will not attempt to cache large array data, but it will cache the small metadata messages and reduce the latency cost.

The operation of the cache can be observed by turning on verbose logging:

```py
from tiled.client import show_logs
show_logs()
```

---


cjh1 (approved):
LGTM

PR #2240, opened 2022-04-14
Ctf correction

jtschwar:
Adding a ctf correction function for BF/Cryo TEM (in response for reviewer in our Nature Comm manuscript)

### DIFF
--- a/tomviz/python/ctf_correct.py
+++ b/tomviz/python/ctf_correct.py
@@ -1,72 +1,75 @@
 import numpy as np
 
-# Given an dataset containing one or more 2D images, apply CTF operations on them. 
+
+# Given an dataset containing one or more 2D images, 
+# apply CTF operations on them.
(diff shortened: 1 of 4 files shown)


tomviz/python/ctf_correct.py
```diff
@@ -0,0 +1,76 @@
...
+                       (DF - 0.5 * WL * WL * radialMeshSq * Cs))
+
+    CTFim = -w1 * np.sin(Xr) - w2 * np.cos(Xr)
+
+    return CTFim
+
+
+# Calculate the Electron Wavelength is Angstroms
+def ElectronWavelength(kV):
+    return 12.2639 / np.sqrt(kV * 1e3 + 0.97856 * (kV**2))
```
cjh1, line 76: Can we use variable to describe these "magic" numbers?

jtschwar: @cjh1 yep, I've updated this function so it contains all the labeled variables now

cjh1 (approved):
LGTM

Issue #2246, opened 2022-05-25
Report missing dependencies


cjh1:
Report missing dependencies that fail to import [here](https://github.com/OpenChemistry/tomviz/blob/master/tomviz/python/tomviz/pyxrf/__init__.py#L6)

PR #2247, opened 2022-05-25
Show PyXRF import error in tooltip

psavery:
If some required modules for PyXRF failed to import, then display the
import error in the tooltip. This provides needed information on why the
import failed.

An example is shown below.

image

Fixes: #2246


cjh1 (approved):
LGTM, thanks for the quick turn around!

cjh1 (approved):

PR #2255, opened 2022-06-22
Force gradient opacity for multi-volumes only

psavery:
Previously, as part of the multi-volume work, a gradient opacity was
set on every volume to circumvent a bug with multi-volumes, where they
wouldn't work without a gradient opacity.

However, setting a gradient opacity appears to have slowed rendering of
volumes by roughly 4x.

As such, do not set the gradient opacity on volumes unless we are
actively using multi-volumes, to avoid the penalty in rendering
performance.

I have tested these changes with the multi-volume features, including
toggling on/off multi-volumes, and loading a state file where multi-volumes
are active, and the changes appear to work properly.


cjh1 (approved):
LGTM

PR #2257, opened 2022-07-22
Update for tiled v0.1.0a66 API changes.

danielballan:
Signed-off-by: Dan Allan <dallan@bnl.gov>

Tiled (core library of Databroker nowadays) made some background-incompatible changes in v0.1.0a66. This PR accounts for the (fairly minor) changes to the Tiled Python client API. The changes will also work with older version of the tiled client. That is: they are _necessary_ for v0.1.0a66+ and they are effectively a no-op for older versions.


danielballan:
attn @cryos @cjh1

cryos (approved):
Adding some details for me at least - in the updated client API it returns an `ArrayClient` which we must act on, either fetch the array, or slice it in various ways. Just one more abstraction ;-)

Issue #2214, opened 2021-11-02
Crashes when clicking "About Tomviz" (OSX)


Hovden:
**Describe the bug**
Tomviz crashes when "About Tomviz" is selected from drop down.

**To Reproduce**
1. Go to 'tomviz > "About tomviz' on OSX build
2. See crash

[Macbook Pro]

cjh1:
@Hovden Is this on M1?

Hovden:
No. This is on my Macbook Pro.

jtschwar:
@cjh1 I was able to reproduce this issue on my mac (2015 Mabook Pro running BigSur)

harripj:
I also have this issue on my Intel MacBook Pro 16".

PR #2260, opened 2022-08-16
Add Nature Communications citation to welcome window

jtschwar:

### DIFF
--- a/tomviz/WelcomeDialog.ui
+++ b/tomviz/WelcomeDialog.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>517</width>
-    <height>246</height>
+    <width>562</width>
+    <height>261</height>
    </rect>
   </property>
   <property name="windowTitle">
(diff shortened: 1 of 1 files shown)


cryos (commented):
I don't think this belongs in the welcome dialog, and personally I think we should look as a group at how we want to ask for citations, probably offer a hyperlink to any DOIs. I think that the About dialog is more appropriate, I would still like to develop a full Tomviz paper as the preferred citation as we discussed during the SBIR I was the PI on when at Kitware but in its absence I see the motivation to promote this more.

jtschwar:
I'm not sure if the About dialog is the best location, there's still an issue that it causes tomviz to crash on Mac operating systems ( #2214 )

Hovden:
@cryos I agree that we need a generic tomviz paper. Currently, it is important we accumulate citations for our existing manuscript (the 2-year citation count will influence future papers). We can swap them in the dialogue when the other manuscript is published. I am available to talk over the phone about writing the next paper.

Hovden:
Following up on this. Are we merging?

Hovden:
@cryos Is this an easy bug you could resolve? Then add the citation to the about. Also, we purchased a CC license for the manuscript so we can distribute with tomviz.

PR #2265, opened 2022-09-21
Fix python syntax highlighter

psavery:
ParaView no longer automatically connects the highlighter, so we must connect it manually.

Additionally, we have to put it before the text is set so that the text will be highlighted.

Finally, it appears that the convention now in ParaView is to make the parent the script, so do that.


cjh1 (approved):
LGTM

PR #2266, opened 2022-09-24
Wait to load looking glass plugin until used

psavery:
Since there is now an EULA that pops up when the looking glass plugin is loaded, we don't want to load the looking glass plugin automatically when tomviz starts, because we don't want all users to be prompted to accept an EULA for a plugin that they may not use.

Now, there is a placeholder action added to the view menu that, when triggered, causes the looking glass plugin to be loaded, and then the placeholder action removes itself (it is effectively replaced by the actual looking glass dock widget toggle view action).

This should result in seemingly identical behavior for most users.


cjh1 (approved):
LGTM

PR #2273, opened 2022-11-30
Add slice module to animation helper

psavery:
This currently only works for orthogonal slices, however.

If we want this to work with non-orthogonal slices, we can perhaps allow a decimal number range between 0 - 1, for instance, and add some logic somewhere to iterate over computed positions.


ercius:
@psavery This would be great to get into the 2.0 release

tomviz/AnimationHelperDialog.ui
```diff
@@ -7,7 +7,7 @@
     <x>0</x>
     <y>0</y>
     <width>699</width>
-    <height>448</height>
+    <height>449</height>
```
cjh1, line 10: Any reason for the height increase or was is just designer doing its thing ?

cjh1 (approved):
LGTM, feel free to merge once CI completes.

tomviz/AnimationHelperDialog.ui
psavery: I think it was just the designer doing its thing.

Issue #1848, opened 2019-04-23
Set custom slice normal to view direction


ercius:
When looking at data I often use a volume render to find an interesting axis and then I want to slice down that axis using a custom (non-orthogonal) slice.   It would be great to have a button to set the slice normal to the current viewing direction.

**Describe the solution you would like**
Add a button to set a custom (non-orthogonal) slice to the current viewing direction of the camera.

**Describe alternatives you have considered**
The only way to do this now is to rotate the 3D viewer camera to the desired view direction and then set the slice "arrow" so it is parallel to the current view direction. See screenshots below.

**Additional context**
1) Desired viewing direction (111 of NP)
![1](https://user-images.githubusercontent.com/15657108/56606057-4b1c9e80-65ba-11e9-9a9a-48aca3449a35.png)
2) Add a slice and set direction to custom
![2](https://user-images.githubusercontent.com/15657108/56606110-62f42280-65ba-11e9-8b2d-92197da2554c.png)
3) Rotate arrow to be parallel to camera viewing direction
![3](https://user-images.githubusercontent.com/15657108/56606144-6e474e00-65ba-11e9-98b1-db46ad01e2ec.png)
4) Remove volume render. I can now see down the (111) NP axis.
![4](https://user-images.githubusercontent.com/15657108/56606172-7b643d00-65ba-11e9-81ec-14b45bffcc0d.png)

This is not just for atomic resolution data. It is also useful for many other types of data like nanowires or other structures where the desired viewing direction is not orthogonal to the X, y or Z directions.

PR #2276, opened 2023-01-25
Add button to set slice normal to view

psavery:
This sets the slice normal to be the same as the view direction.

https://user-images.githubusercontent.com/9558430/214708921-ee7fce09-40ab-4bba-b8ee-7498be46e489.mp4

Fixes: #1848


cjh1 (approved):
LGTM, just waiting on CI

Issue #2275, opened 2023-01-25
Animation helper "Export movie" button


ercius:
Add a button to the Animation Helper to open the Export Movie workflow. This would be the same as the "File -- Export movie" command, but more obvious to the user. It would be good to add below "Clear all Animations" so its available on all tabs.

psavery:
I ended up putting it right above "Clear all Animations" instead, since I thought it less likely for users to accidentally push "Clear all Animations" when they intended to export a movie :-). But it is still available on all tabs.

PR #2277, opened 2023-01-25
Add "Export Movie" button to the animation helper

psavery:
This proceeds with the same function call as if the user had clicked "File" -> "Export Movie...".

image

Fixes: #2275


cjh1 (approved):
Looks good, I do wonder though if we should start have some of these button laid out horizontal, rather than the massive buttons taking up the whole width.

PR #2278, opened 2023-01-27
Remove old CI code and scripts

psavery:
CircleCI had a large security breach in January. Any old secrets need to be rotated out.

However, we haven't used CircleCI in a long time. Instead of checking if old secrets are still in place, just eliminate all of the unused CI in general so we don't need to worry about it.


cjh1 (approved):
LGTM

Issue #2154, opened 2021-03-08
Summation saturates slice image


ercius:
**Describe the bug**
The slice module allows for a slice thickness and defaults to Aggregation = 'summation'. If you sum any slices the slice image becomes saturated. This is because the max/min display value is set for the normal data range, but a sum increase the intensities by 3x, 5x, etc. 

It is also impossible to change the displayed data range higher than the maximum value in the data set. So, the summation option can not really be used at all.

I would suggest removing the summation option and making the default Aggregation to mean.
OR simply:
Just change the default to mean

**To Reproduce**
1. Load a dataset and click on the slice operator.
2. Leave the Aggregation as "summation"
3. Increase the slice thickness to 3, 5, 7. The slice image saturates.

**Expected behavior**
The slice should not saturate. Using the mean instead of summation will fix this problem.

**Screenshots**
![image](https://user-images.githubusercontent.com/15657108/110382017-2e06e280-800f-11eb-8382-d5c0a2c7e645.png)

![image](https://user-images.githubusercontent.com/15657108/110382142-57277300-800f-11eb-93ba-58e1a43c4b6b.png)

**Desktop (please complete the following information):**
 - Operating system: Windows 10
 - Tomviz version: 1.9.0

PR #2280, opened 2023-02-04
Change the default thick slice type to mean

psavery:
This does not saturate, which is better for visualization.

Fixes: #2154


cjh1 (approved):
LGTM

Issue #2279, opened 2023-01-31
Reconstruction slices saturated in tilt axis alignment wizard


ercius:
**Describe the bug**
The tilt axis alignment wizard has three reconstructed slices at the bottom. These can be saturated if the values in the reconstructed slices are beyond the range of the original tilt axis histogram. These images do not respond to the histogram range values after the wizard is open.

**To Reproduce**
1. Load example tilt series
2. Convert to float (to avoid overflow)
3. Add a constant of 1000 simulating a constant background value.
4. Open the tilt axis alignment (manual) wizard.

**Expected behavior**
All images in this wizard should be normalized 0-1 before display. We are not worried about quantitive ranges.

**Screenshots**
Result from steps above:
![image](https://user-images.githubusercontent.com/15657108/215633837-3ab2156e-1313-43dd-9c78-f7bced26c4de.png)

Result without added constant.
![image](https://user-images.githubusercontent.com/15657108/215633635-cd8348b3-2046-451f-bf92-0290b86c1005.png)

**Desktop (please complete the following information):**
 - Windows 11
 - Tomviz version: Nightly downloaded 2023/01/30

**Hardware (please complete the following information):**
 - Device: Laptop
 - Graphics card: NVIDIA

psavery:
Just to leave a note here: #2281 does fix this issue as long as there is almost no rotation. It did so by making the color range via the data in the inscribed circle. But when there is rotation, it isn't an inscribed circle as much anymore, so we may need to revisit this if that becomes important.

PR #2281, opened 2023-02-06
Improve alignment preview image color ranges

psavery:
For datasets that had a minimum greater than zero, the preview images in the alignment widget would often appear saturated. That is because the actual values in the inscribed circle would become much higher than the values outside of it (which we should ignore anyways).

Thus, when we compute the range to include for the color range, only check pixel values inside the inscribed circle (and a certain distance from the edge as well, since edge pixels appear to drop fast in value).

This makes the preview images look much better.

Fixes: #2279

### DIFF
--- a/tomviz/RotateAlignWidget.cxx
+++ b/tomviz/RotateAlignWidget.cxx
@@ -241,8 +241,35 @@ public:
       this->reconSliceMapper[i]->SetSliceNumber(0);
       this->reconSliceMapper[i]->Update();
 
-      double range[2];
-      reconArray->GetRange(range);
+      double range[2] = {DBL_MAX, -DBL_MAX};
+
+      // Get the range of the inscribed circle only
+      auto radius = static_cast<double>(Nray) / 2;
+
+      // The max distance for what we will keep is the radius multiplied by
+      // some reduction factor in order to exclude edge pixels. The images come
+      // out better if we exclude edge pixels since the pixel values tend to
+      // drop fast near the edges. So this factor is a magic number.
+      auto maxDistance = radius * 0.97;
+      for (int j = 0; j < Nray; ++j) {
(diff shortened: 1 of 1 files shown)


psavery:
@ercius Unfortunately, I did just notice one other problem too.

If the tilt rotation axis is non-zero, then the image isn't really an inscribed circle anymore, and thus we encounter saturation again:

image

ercius:
Hm. That makes this harder. The wizard is not meant for huge rotation corrections and you probably cant take care of every edge case. I recommend subtracting the minimum before doing this analysis in such a case. So, lets go with your current solution which will be helpful in most experiments.

psavery:
It looks like the images even start appearing more saturated with about a 5 degree rotation.
image

Should I also subtract the minimum from the data before performing the slice reconstruction? Or are we going to leave that for the user to do?

ercius:
I'm wary to subtract the minimum. I sometimes use this to see what the
reconstruction will look like. So, I think we need to encourage users to
set the minimum to zero.

psavery:
Okay! @cjh1 This is ready for your review, then.

tomviz/RotateAlignWidget.cxx
```diff
@@ -241,8 +241,35 @@ class RotateAlignWidget::RAWInternal
       this->reconSliceMapper[i]->SetSliceNumber(0);
       this->reconSliceMapper[i]->Update();
 
-      double range[2];
-      reconArray->GetRange(range);
+      double range[2] = {DBL_MAX, -DBL_MAX};
+
+      // Get the range of the inscribed circle only
+      auto radius = static_cast<double>(Nray) / 2;
```
cjh1, line 247: For future reference, if you divide by `2.0` you don't need the `static_cast`

cjh1 (approved):
LGTM

Issue #2241, opened 2022-04-28
Manual tilt axis incorrect when "vertical" selected


ercius:
**Describe the bug**
Note: I realize that the screenshots below show 1.9.0. I tested this on 1.10.0 and the I see the same behavior.

The manual tilt axis alignment wizard does not work correctly when the "vertical" tilt axis direction is selected. I generated projections of a test object from +/- 90 degrees to remove any issues with in plane alignments in the screen shots below. The test object is 9 round spheres arranged in a body-centered cubic arrangement. The tilt axis should be horizontal, but I can change that by using the Data Transforms - Rotate command to rotate around the Z-axis.

1) Here is the object with no rotation applied
![Screenshot 2022-04-28 104506](https://user-images.githubusercontent.com/15657108/165816961-ef505243-ed5e-4567-8fa2-2a408b406d1d.png)

2) Here is the object with a 5 degree rotation applied in the pipeline.  The tilt axis is Horizontal. The yellow tilt axis makes sense and the reconstructed slices are correct.
![Screenshot 2022-04-28 104622](https://user-images.githubusercontent.com/15657108/165817075-d01a327a-bb91-4760-ae60-ba769877e8da.png)

3) Here is the object with a 95 degree rotation applied and the tilt axis set as "vertical" and -5 degree rotated tilt axis. The yellow tilt axis is correct but the reconstructed slices are incorrect
![Screenshot 2022-04-28 104707](https://user-images.githubusercontent.com/15657108/165817286-29b4c2bf-0bc1-43ed-9e4c-fe6cb9afc282.png)

4) Here is the object rotated with 95 degree applied and the tilt axis set as vertical. I use a 5 degree rotated tilt axis. The yellow tilt axis is incorrect but the reconstructed slices are correct.
![Screenshot 2022-04-28 104730](https://user-images.githubusercontent.com/15657108/165817471-d4abe16d-03e7-47f7-9ede-a614404503d5.png)

5) Here is the tilt axis wizard and the main window together. I set the tilt axis to -5 degrees and the yellow tilt axis matches the expected direction in the figure. However, the reconstructed slices at the bottom are incorrect. I pressed apply so that the main window on the left updated with the -5 degree rotation applied and this figure *looks correct*.  
![Screenshot 2022-04-28 110111](https://user-images.githubusercontent.com/15657108/165817760-1c9ce567-009e-459c-9012-3bec2e991bba.png)

[...]

PR #2283, opened 2023-02-28
Rotate axis oppositely for vertical alignment

psavery:
This is a quick fix so that the drawn axis matches the drawn images correctly.

Fixes: #2241


psavery:
@ercius Thank you for providing me with that script. I was able to reproduce the issue in #2241.

Here is what the data looks like now for the vertical cases. Can you verify that they look correct?

# -5 rotation
negative_five

# +5 rotation
five

cjh1 (approved):
LGTM

PR #2286, opened 2023-04-19
The int used by nTuples overflows for big volumes

cryos:
This hit me today where GetNumberOfTuples() returns vtkIdType but this templated function was using int which then overflows for big volumes.


cryos:
I am looking at a subset of a 217GB reconstruction, a highly cropped version still caused this issue. I have my work cut out for me ;-)

psavery (approved):
LGTM!

psavery:
By the way, Mac is failing because tbb@2020 is no longer available (see #2282 where I tried to fix it but could not). So don't worry about that one.

psavery:
> I am looking at a subset of a 217GB reconstruction, a highly cropped version still caused this issue. I have my work cut out for me ;-)

That's a lot of data!

PR #2287, opened 2023-04-20
Fix a few more int overflows from vtkIdType

cryos:
I spotted a few more obvious cases where we can overflow an int counter and fixed them. This was mainly from grep, so there may be more lurking.


psavery (approved):
LGTM

PR #2302, opened 2025-12-06
Remove external dependency on 'mock'

padraic-shafer:
This PR removes an unnecessary external dependency on 'mock'. It now uses `unittest.mock` instead.


padraic-shafer:
Submitted to wrong upstream repo.

